PR #5, opened 2020-06-10
Set KeepAliveTimeout to 620000 (620secs)

pengin7384:
- Fix 502 error

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -285,8 +285,8 @@ const server = app.listen(PORT, () => {
   logger.info(`[${CLIENT_PREFIX}] Press Ctrl+C to quit.`);
 });
 
-server.keepAliveTimeout = 620000;
-server.headersTimeout = 630000;
+server.keepAliveTimeout = 620 * 1000; // 620 seconds
+server.headersTimeout = 630 * 1000; // 630 seconds
 
 // Lets start this p2p server up so we listen for changes in either DATABASE
 // or NUMBER OF SERVERS
--- a/client/index.js
+++ b/client/index.js
@@ -280,10 +280,12 @@ app.get('/get_address', (req, res, next) => {
 
 // We will want changes in ports and the database to be broadcast across
 // all instances so lets pass this info into the p2p server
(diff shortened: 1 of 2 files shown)


client/index.js
```diff
@@ -280,11 +280,13 @@ app.get('/get_address', (req, res, next) => {
 
 // We will want changes in ports and the database to be broadcast across
 // all instances so lets pass this info into the p2p server
-app.listen(PORT, () => {
+const server = app.listen(PORT, () => {
   logger.info(`[${CLIENT_PREFIX}] App listening on port ${PORT}`);
   logger.info(`[${CLIENT_PREFIX}] Press Ctrl+C to quit.`);
 });
 
+server.keepAliveTimeout = 620000;
```
liayoo, line 288: Could you also set headersTimeout to be greater than keepAliveTimeout? (related issue: https://github.com/nodejs/node/issues/27363)

pengin7384: Thanks! Done.

pengin7384:
PTAL! @liayoo

liayoo:
LGTM. Thanks for fixing the issue!

pengin7384:
PTAL! @platfowner

client/index.js
```diff
@@ -280,11 +280,14 @@ app.get('/get_address', (req, res, next) => {
...
```
platfowner, line 288: Does this value have any special meaning? It seems that some articles (like https://shuheikagawa.com/blog/2019/04/25/keep-alive-timeout/) suggest some values 61 * 1000.

And what about adding some comments about the unit? e.g.
server.keepAliveTimeout = 62 * 1000;  // 62 seconds

pengin7384: Google load balancer has a tcp timeout of 600 seconds.([link](https://cloud.google.com/load-balancing/docs/https#timeouts_and_retries)) keepAliveTimeout should be set to a value greater than this. In fact, for the tracker server, load balancing has not been applied yet, so it is not necessary to set it up yet.

Comments are added!

PR #13, opened 2020-07-22
Run when all stuck and restart together

minsulee2:
Changes:

- When a new coming node has got the same number of the rest of nodes in network, it is now able to run with init().

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -288,6 +288,10 @@ class P2pServer {
             // Check if chain subsection is valid and can be
             // merged ontop of your local blockchain
             if (data.number <= this.node.bc.lastBlockNumber()) {
+              if (data.number === this.node.bc.lastBlockNumber() &&
+                  this.consensus.status === ConsensusStatus.STARTING) {
+                    this.consensus.init(this.node.bc.lastBlock());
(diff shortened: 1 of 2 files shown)


server/index.js
```diff
@@ -318,16 +322,18 @@ class P2pServer {
                 this.requestChainSubsection(this.node.bc.lastBlock());
               }
             } else {
+              logger.info(`[${P2P_PREFIX}] Failed to merge incoming chain subsection.`);
               // FIXME: Could be that I'm on a wrong chain.
               if (data.number <= this.node.bc.lastBlockNumber()) {
-                logger.info(`[${P2P_PREFIX}] Failed to merge incoming chain subsection.`);
+                logger.info(`[${P2P_PREFIX}] I am a leader(${data.number} > ${this.node.bc.lastBlockNumber()}).`);
```
liayoo, line 328: I think this log could be a little misleading - maybe change to "I am ahead"?

minsulee2: Oh! fair enough! I will change and merge it! thanks!

minsulee2:
Thanks for the quick review!

PR #14, opened 2020-07-22
Bugfix/lia/blockpool cleanup

liayoo:
- Fixed block pool clean-up logic to delete records of blocks with numbers and epochs smaller than the finalized block
- Improved logs


liayoo:

PR #15, opened 2020-07-22
Bugfix/consensus/duplicate messages

liayoo:
- Changed to prevent processing same consensus messages more than once
- Adjust and sync epoch every 100 epoch

### DIFF
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -181,6 +181,12 @@ class Consensus {
         logger.error(`[${LOG_PREFIX}:${LOG_SUFFIX}] Proposal is missing required fields: ${msg.value}`);
         return;
       }
+      if (this.node.tp.transactionTracker[proposalTx.hash]) {
+        if (DEBUG) {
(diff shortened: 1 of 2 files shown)


consensus/index.js
```diff
@@ -92,9 +94,18 @@ class Consensus {
     if (this.epochInterval) {
       clearInterval(this.epochInterval);
     }
-    this.epochInterval = setInterval(() => {
+    this.epochInterval = setInterval(async () => {
       this.tryFinalize();
       let currentTime = Date.now();
+      if (this.state.epoch % 100 === 0) {
+        // adjust time
+        const iNTPData = await ntpsync.ntpLocalClockDeltaPromise();
```
minsulee2, line 102: What about try {} catch { return 0 } block for this?
If ntp fails, it may cause epoch corruption.

liayoo: Thanks, added a try-catch block.

liayoo:
Addressed all comments, PTAL @minsulee2

consensus/index.js
minsulee2: Coooooool. Thanks for the update!

PR #16, opened 2020-07-23
Feature/consensus/stop

minsulee2:
Changes:

- Safe exit process with SIGINT
- Now support socket.on('close')

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -20,6 +20,14 @@ process.on('uncaughtException', function (err) {
   logger.error(`[${CLIENT_PREFIX}]` + err);
 });
 
+process.on('SIGINT', _ => {
+  logger.info("############################################################");
+  logger.info("##          Stopping the blockchain client....            ##");
+  logger.info("############################################################");
+  p2pServer.stop();
+  process.exit();
+});
+
 if (!fs.existsSync(PROTOCOL_VERSIONS)) {
   throw Error('Missing protocol versions file: ' + PROTOCOL_VERSIONS);
 }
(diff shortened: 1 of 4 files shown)


tracker-server/index.js
```diff
@@ -124,7 +124,7 @@ server.on('connection', (ws) => {
 
   ws.on('close', (code) => {
     logger.info(`\nDisconnected from node [${node ? abbrAddr(node.address) : 'unknown'}] ` +
-        `with code: ${code}`);
+                `with code: ${code}`);
```
liayoo, line 127: remove spaces?

client/index.js
```diff
@@ -20,6 +20,14 @@ process.on('uncaughtException', function (err) {
   logger.error(`[${CLIENT_PREFIX}]` + err);
 });
 
+process.on('SIGINT', _ => {
+  logger.info("############################################################");
+  logger.info("##          Stopping the blockchain client....            ##");
+  logger.info("############################################################");
```
liayoo, line 26: (suggestion) can we make this into one log?

server/index.js
```diff
@@ -372,7 +393,7 @@ class P2pServer {
     });
 
     socket.on('close', () => {
```
liayoo, line 395: Could you leave a todo for handling/recording disconnections and errors here?

minsulee2: I left TODO here!

tracker-server/index.js
minsulee2: I got this

client/index.js
minsulee2: made it smaller :)

minsulee2:
Thanks for the review!

PR #17, opened 2020-07-23
Tweak get subtree functions / rules logic.

platfowner:
It's simply pulling ChainUtil.isDict() out of the loops, but github diff shows like a big change.

### DIFF
--- a/db/index.js
+++ b/db/index.js
@@ -520,26 +520,28 @@ class DB {
         config: DB.getFunctionConfig(curFuncNode),
       })
     }
-    const varNodeName = DB.getVariableNodeName(curFuncNode);
-    // 1) Traverse non-variable child nodes.
-    for (const key in curFuncNode) {
-      const nextFuncNode = curFuncNode[key];
-      if (key !== varNodeName && ChainUtil.isDict(nextFuncNode)) {
+    if (ChainUtil.isDict(curFuncNode)) {
+      const varNodeName = DB.getVariableNodeName(curFuncNode);
+      // 1) Traverse non-variable child nodes.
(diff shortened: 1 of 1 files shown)


db/index.js
```diff
@@ -667,27 +669,29 @@ class DB {
...
     }
-    const varNodeName = DB.getVariableNodeName(curRuleNode);
-    // 1) Traverse non-variable child nodes.
-    for (const key in curRuleNode) {
-      const nextRuleNode = curRuleNode[key];
-      if (key !== varNodeName && ChainUtil.isDict(nextRuleNode)) {
+    if (ChainUtil.isDict(curRuleNode)) {
+      const varNodeName = DB.getVariableNodeName(curRuleNode);
+      // 1) Traverse non-variable child nodes.
+      for (const key in curRuleNode) {
```
minsulee2, line 675: Sorry for the late review on merged PR though, the 'key' in the for loop with 'in' is not the actual key value but number. Is this intended?

minsulee2: If it is not, it should be 'of' instead of 'in' :)

liayoo: isDict() condition makes sure that curRuleNode here an object, not an array.. so I think the current code is correct.

minsulee2: Oh, my apology! I thought it is an array somehow. Thanks for the comment!

platfowner: @minsulee2  Thanks for asking. It'll become much clearer by this CL soon => https://github.com/ainblockchain/ain-blockchain/tree/feature/platfowner/node_abstraction 
And I'll add some sanity checking logic to avoid array input soon!

PR #18, opened 2020-07-23
Feature/platfowner/node abstraction

platfowner:
Summary:
- Add StateNode class for node abstraction
- Add state-util for util functions
- Make DB functions use StateNode instead of Javascript object for DB states
- Make ChainUtil reuse RuleUtil functions
- Add tests for ChainUtil and state-util

### DIFF
--- a/db/state-node.js
+++ b/db/state-node.js
@@ -1,24 +1,40 @@
 class StateNode {
   constructor() {
+    this.isLeaf = false;
     // Used for internal nodes only.
     this.childMap = new Map();
     // Used for leaf nodes only.
     this.value = null;
   }
 
-  static create(childMap, value) {
+  static create(isLeaf, childMap, value) {
     const node = new StateNode();
+    node.isLeaf = isLeaf;
     node.childMap = new Map(childMap);
     node.value = value;
     return node;
   }
 
   makeCopy() {
-    return StateNode.create(this.childMap, this.value);
+    return StateNode.create(this.isLeaf, this.childMap, this.value);
+  }
+
+  getIsLeaf() {
+    return this.isLeaf;
+  }
+
+  setIsLeaf(isLeaf) {
+    this.isLeaf = isLeaf;
+  }
+
+  resetValue() {
+    this.setValue(null);
+    this.setIsLeaf(false);
   }
 
   setValue(value) {
     this.value = value;
+    this.setIsLeaf(true);
   }
 
   getValue() {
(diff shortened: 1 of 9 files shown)


db/state-node.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  deleteChild(label) {
+    this.childMap.delete(label);
+    if (this.getNumChild() === 0) {
+      this.setIsLeaf(true);
+    }
+  }
+
+  getChildLabels() {
```
liayoo, line 70: Suggestion - there are one-liners for this operation, such as
Array.from(this.childMap.keys()) or [ ...this.childMap.keys() ]
(ref: https://stackoverflow.com/questions/35341696/how-to-convert-map-keys-to-array)

liayoo: And the same goes for getChildNodes()

db/state-util.js
```diff
@@ -0,0 +1,76 @@
...
+      }
+    }
+    return true;
+  } else {
+    return ChainUtil.isBool(obj) || ChainUtil.isNumber(obj) || ChainUtil.isString(obj) ||
+        obj === null;
+  }
+}
+
+function convertToStateTree(obj) {
```
liayoo, line 20: Suggestion - how about something like jsonToStateTree() and stateTreeToJson() ?

db/state-util.js
```diff
@@ -0,0 +1,76 @@
...
+    for (const key in obj) {
+      const childObj = obj[key];
+      const isValidChild = isValidStateObject(childObj);
+      if (!isValidChild) {
+        return false;
+      }
+    }
+    return true;
+  } else {
+    return ChainUtil.isBool(obj) || ChainUtil.isNumber(obj) || ChainUtil.isString(obj) ||
```
liayoo, line 15: Are we not allowing arrays to be in state objects?

db/state-node.js
platfowner: Done
Thanks!

db/state-util.js
platfowner: It looks like 'JSON' and 'javascript object' are not always interchangeable. Tried to rename jsObjectToStateTree() and stateTreeToJsObject(). PTAL

db/state-util.js
platfowner: Right. Like firebase realtime database, it would make things much simpler if we do not allow array for states. What do you think?

liayoo: I see, I think that's reasonable. Will update docs.ainetwork.ai to reflect this then!

db/state-util.js
liayoo: Right, the new names look good to me!

db/state-util.js
minsulee2: See. I will keep this in mind!

platfowner:
Thanks for the review!
Now merging..

PR #19, opened 2020-07-24
Now enable ahead node runs in STARTING stage

minsulee2:
Changes:

- ahead node is able to reach at init() when starting node again.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -306,15 +306,20 @@ class P2pServer {
             if (DEBUG) {
               logger.debug(`[${P2P_PREFIX}] Receiving a chain subsection: ${JSON.stringify(data.chainSubsection, null, 2)}`);
             }
-            // Check if chain subsection is valid and can be
-            // merged ontop of your local blockchain
             if (data.number <= this.node.bc.lastBlockNumber()) {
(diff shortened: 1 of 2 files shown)


server/index.js
```diff
@@ -306,15 +306,20 @@ class P2pServer {
...
               logger.debug(`[${P2P_PREFIX}] Receiving a chain subsection: ${JSON.stringify(data.chainSubsection, null, 2)}`);
             }
-            // Check if chain subsection is valid and can be
-            // merged ontop of your local blockchain
             if (data.number <= this.node.bc.lastBlockNumber()) {
-              if (data.number === this.node.bc.lastBlockNumber() &&
-                  this.consensus.status === ConsensusStatus.STARTING) {
-                    this.consensus.init(this.node.bc.lastBlock());
+              if (this.consensus.status === ConsensusStatus.STARTING) {
+                if (!data.chainSubsection && !data.catchUpInfo) {
```
liayoo, line 311: Looks like this and the if statement below are doing the same thing? Maybe combine the two with an OR?

minsulee2:
Oh, Thanks for the review! I will merge it as soon as leaving this comment

PR #24, opened 2020-07-26
Feature/consensus/whitelist validators

liayoo:
Setting a whitelist of validators who can participate in consensus.

### DIFF
--- a/blockchain/block.js
+++ b/blockchain/block.js
@@ -315,7 +315,7 @@ class Block {
     const proposer = ownerAccount.address;
     const validators = {};
     for (let i = 0; i < ConsensusConsts.INITIAL_NUM_VALIDATORS; i++) {
-      validators[GenesisAccounts.others[i].address] = 250;
+      validators[GenesisAccounts.others[i].address] = ConsensusConsts.INITIAL_STAKE;
     }
     return new this(lastHash, lastVotes, transactions, number, epoch, timestamp,
         proposer, validators);
--- a/blockchain/block.js
+++ b/blockchain/block.js
@@ -6,7 +6,7 @@ const ChainUtil = require('../chain-util');
 const Transaction = require('../tx-pool/transaction');
 const {
   GENESIS_OWNERS, ADDITIONAL_OWNERS, GENESIS_RULES, ADDITIONAL_RULES, GENESIS_FUNCTIONS,
-  ADDITIONAL_FUNCTIONS, PredefinedDbPaths, GenesisToken, GenesisAccounts
+  ADDITIONAL_FUNCTIONS, PredefinedDbPaths, GenesisToken, GenesisAccounts, GenesisWhitelist
 } = require('../constants');
 const { ConsensusDbPaths, ConsensusConsts } = require('../consensus/constants');
 const BlockFilePatterns = require('./block-file-patterns');
@@ -165,13 +165,7 @@ class Block {
     const whitelistValOp = {
       type: 'SET_VALUE',
       ref: `/${ConsensusDbPaths.CONSENSUS}/${ConsensusDbPaths.WHITELIST}`,
-      value: {
-        '0x00ADEc28B6a845a085e03591bE7550dd68673C1C': 250,
-        '0x01A0980d2D4e418c7F27e1ef539d01A5b5E93204': 250,
-        '0x02A2A1DF4f630d760c82BE07F18e5065d103Fa00': 250,
-        '0x03AAb7b6f16A92A1dfe018Fe34ee420eb098B98A': 250,
-        '0x04A456C92A880cd59D7145C457475515a6f6E0f2': 250
-      }
+      value: GenesisWhitelist
     };
 
     // Function configs operation
--- a/blockchain/block.js
+++ b/blockchain/block.js
@@ -162,7 +162,7 @@ class Block {
 
     // Consensus (whitelisting) operation
     // TODO(lia): increase this list to 10
-    const whitelistOp = {
+    const whitelistValOp = {
       type: 'SET_VALUE',
       ref: `/${ConsensusDbPaths.CONSENSUS}/${ConsensusDbPaths.WHITELIST}`,
       value: {
@@ -224,6 +224,29 @@ class Block {
       ref: '/',
       value: ownerConfigs
     };
+    const whitelistRuleOp = {
+      type: 'SET_RULE',
+      ref: '/consensus/whitelist',
+      value: `auth === '${ownerAccount.address}'`
+    }
+    const whitelistOwnerOp = {
+      type: 'SET_OWNER',
+      ref: '/consensus/whitelist',
+      value: {
+        [ownerAccount.address]: {
+          "branch_owner": true,
+          "write_function": true,
+          "write_owner": true,
+          "write_rule": true
+        },
+        "*": {
+          "branch_owner": false,
+          "write_function": false,
+          "write_owner": false,
+          "write_rule": false
+        }
+      }
+    }
 
     // Transaction
     const firstTxData = {
@@ -231,7 +254,7 @@ class Block {
       timestamp,
       operation: {
         type: 'SET',
-        op_list: [ tokenOp, balanceOp, whitelistOp, functionsOp, rulesOp, ownersOp ]
+        op_list: [ tokenOp, balanceOp, whitelistValOp, functionsOp, rulesOp, ownersOp, whitelistRuleOp, whitelistOwnerOp ]
       }
     };
     const firstSig = ainUtil.ecSignTransaction(firstTxData, keyBuffer);
--- a/blockchain/block.js
+++ b/blockchain/block.js
@@ -8,6 +8,7 @@ const {
   GENESIS_OWNERS, ADDITIONAL_OWNERS, GENESIS_RULES, ADDITIONAL_RULES, GENESIS_FUNCTIONS,
   ADDITIONAL_FUNCTIONS, PredefinedDbPaths, GenesisToken, GenesisAccounts
 } = require('../constants');
+const { ConsensusDbPaths, ConsensusConsts } = require('../consensus/constants');
 const BlockFilePatterns = require('./block-file-patterns');
 const zipper = require('zip-local');
 const sizeof = require('object-sizeof');
@@ -159,6 +160,20 @@ class Block {
       throw Error('Missing genesis rules file: ' + GENESIS_RULES);
     }
 
+    // Consensus (whitelisting) operation
+    // TODO(lia): increase this list to 10
+    const whitelistOp = {
+      type: 'SET_VALUE',
+      ref: `/${ConsensusDbPaths.CONSENSUS}/${ConsensusDbPaths.WHITELIST}`,
+      value: {
+        '0x00ADEc28B6a845a085e03591bE7550dd68673C1C': 250,
+        '0x01A0980d2D4e418c7F27e1ef539d01A5b5E93204': 250,
+        '0x02A2A1DF4f630d760c82BE07F18e5065d103Fa00': 250,
+        '0x03AAb7b6f16A92A1dfe018Fe34ee420eb098B98A': 250,
+        '0x04A456C92A880cd59D7145C457475515a6f6E0f2': 250
+      }
+    };
+
     // Function configs operation
     const functionConfigs = JSON.parse(fs.readFileSync(GENESIS_FUNCTIONS));
     if (ADDITIONAL_FUNCTIONS) {
@@ -216,7 +231,7 @@ class Block {
       timestamp,
       operation: {
         type: 'SET',
-        op_list: [ tokenOp, balanceOp, functionsOp, rulesOp, ownersOp ]
+        op_list: [ tokenOp, balanceOp, whitelistOp, functionsOp, rulesOp, ownersOp ]
       }
     };
     const firstSig = ainUtil.ecSignTransaction(firstTxData, keyBuffer);
@@ -282,6 +297,9 @@ class Block {
     const epoch = 0;
     const proposer = ownerAccount.address;
     const validators = {};
+    for (let i = 0; i < ConsensusConsts.INITIAL_NUM_VALIDATORS; i++) {
+      validators[GenesisAccounts.others[i].address] = 250;
+    }
     return new this(lastHash, lastVotes, transactions, number, epoch, timestamp,
         proposer, validators);
   }
--- a/blockchain/block.js
+++ b/blockchain/block.js
@@ -220,12 +220,12 @@ class Block {
     };
     const whitelistRuleOp = {
       type: 'SET_RULE',
-      ref: '/consensus/whitelist',
+      ref: `/${ConsensusDbPaths.CONSENSUS}/${ConsensusDbPaths.WHITELIST}`,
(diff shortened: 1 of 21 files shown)


blockchain/genesis_owners.json
```diff
@@ -22,5 +22,19 @@
         }
       }
     }
+  },
+  "consensus": {
+    "whitelist": {
+      ".owner": {
+        "owners": {
+          "0xAAAf6f50A0304F12119D218b94bea8082642515B": {
```
platfowner, line 30: What about adding initial setting with "*" here and replace it with the explicit account in block.js?
Using explicit account directly here makes multiple edit point when we need to update the account (e.g. in prod release).

blockchain/block.js
```diff
@@ -159,6 +160,20 @@ class Block {
...
     }
 
+    // Consensus (whitelisting) operation
+    // TODO(lia): increase this list to 10
+    const whitelistOp = {
+      type: 'SET_VALUE',
+      ref: `/${ConsensusDbPaths.CONSENSUS}/${ConsensusDbPaths.WHITELIST}`,
+      value: {
+        '0x00ADEc28B6a845a085e03591bE7550dd68673C1C': 250,
+        '0x01A0980d2D4e418c7F27e1ef539d01A5b5E93204': 250,
```
platfowner, line 170: Question: What does the value 250 mean here? Could it be different from the staking value? If not, could non-numeric value like 'true' be used here?

blockchain/index.js
```diff
@@ -39,6 +39,8 @@ class Blockchain {
         logger.info("## Starting NON-FIRST-NODE blockchain with EMPTY blocks... ##");
         logger.info("#############################################################");
         logger.info("\n");
+        this.chain = [Block.genesis()];
+        this.writeChain();
```
platfowner, line 43: Shouldn't this be fetched from the peers? Is this change for setting consensus properly?

blockchain/genesis_rules.json
```diff
@@ -16,14 +16,17 @@
       "$number": {
         ".write": "newData === null && !!getValue('/consensus/number/' + (Number($number) + 1000))",
         "propose": {
-          ".write": "newData !== null && util.isDict(newData) && newData.proposer === auth && Number($number) === newData.number && (newData.number === 1 || getValue('/consensus/number/' + (newData.number - 1) + '/propose/validators/' + auth) > 0)"
+          ".write": "newData !== null && util.isDict(newData) && newData.proposer === auth && Number($number) === newData.number && getValue('/consensus/whitelist/' + auth) > 0 && (Number($number) < 4 || getValue('/deposit_accounts/consensus/' + auth + '/value') > 0)"
```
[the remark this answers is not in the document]
minsulee2: It seems just because the fixed set of nodes are 5 now?

consensus/index.js
```diff
@@ -381,6 +352,28 @@ class Consensus {
       }
       return;
     }
+    const validators = prevBlock.validators;
+    // check that the transactions from block #1 amount to +2/3 deposits of initially whitelisted validators
+    if (number === 1) {
+      const majority = ConsensusConsts.MAJORITY * Object.values(validators).reduce((a, b) => { return a + b; }, 0);
+      const depositTxs = Consensus.filterDepositTxs(proposalBlock.transactions);
+      const depositSum = depositTxs.reduce((a, b) => { return a + b.operation.value; }, 0);
```
minsulee2, line 360: Question: it seems all nodes start together ?!

blockchain/genesis_rules.json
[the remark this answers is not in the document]
liayoo: It's because we need at least 3 more notarized blocks after the genesis block, in order for the deposit transactions in block#1 to be reflected in the finalized state. 
To be more accurate, the number should be the current lastBlockNumber() instead of the $number in the path though. Currently, we don't have the access to the information in our evalFunc() and I actually wanted to discuss this with you guys: Can I change the evalFunc to pass the lastBlockNumber variable? Or add the variable in RuleUtil?

blockchain/block.js
liayoo: It's just a number that the permissioned nodes agreed upon. I think the initial version could be fixed amounts where nodes first decide on the numbers that cannot change. I imagine the whitelisting process to be:
1. Node requests the admin to be whitelisted, specifying how much it will stake
2. If admin and other nodes accept, admin sets /consensus/whitelist/{nodeAddress} = {stakeAmount} in genesis block
3. Node stakes (/deposit/consensus/{nodeAddress}/{key}/value = {stakeAmount})

If we later support increasing or decreasing stakes, nodes could get the agreement from admin/other nodes, admin would update the whitelist, and the node would deposit/withdraw. But right now, if we set the value in the whitelist to be true and obtain the stakes from /deposit_accounts, I'm afraid we'll face the same problem that we wanted to avoid (for now), which is that validator set can change at any moment. What do you think?

consensus/index.js
liayoo: Yes! With a pre-defined set of whitelisted nodes, we need +2/3 nodes (*not all*) to create and notarize the block #1 (genesis block is still an exception) If we don't have enough nodes on board, we'll keep waiting until we do..

blockchain/genesis_rules.json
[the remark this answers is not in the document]
platfowner: Got it. Passing the value as a new argument looks better to me as util is currently used for static util functions.

blockchain/block.js
platfowner: Ok
Then, can we make this staking value replace STAKE env variable?

consensus/index.js
```diff
@@ -6,10 +6,10 @@ const logger = require('../logger');
 const { Block } = require('../blockchain/block');
 const BlockPool = require('./block-pool');
 const DB = require('../db');
+const Transaction = require('../tx-pool/transaction');
 const PushId = require('../db/push-id');
 const ChainUtil = require('../chain-util');
-const { DEBUG, MessageTypes, STAKE, HOSTING_ENV, WriteDbOperations, PredefinedDbPaths }
-  = require('../constants');
+const { DEBUG, STAKE, WriteDbOperations, PredefinedDbPaths, HOSTING_ENV } = require('../constants');
```
platfowner, line 12: STAKE constant is not used any more?

blockchain/index.js
liayoo: Right - reverted the change.

start_servers.sh
```diff
@@ -1,10 +1,10 @@
 node $0/../tracker-server/index.js &
 sleep 5
-STAKE=250 ACCOUNT_INDEX=0 HOSTING_ENV=local node $0/../client/index.js &
+ACCOUNT_INDEX=0 HOSTING_ENV=local node $0/../client/index.js &
```
platfowner, line 3: Could you update README.md as well?

blockchain/genesis_rules.json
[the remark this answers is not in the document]
liayoo: Updated evalFunc to use the blockchain instance's lastBlockNumber() or pass the number itself if we're constructing a DB snapshot for validating received blocks.

blockchain/block.js
```diff
@@ -159,6 +160,20 @@ class Block {
       throw Error('Missing genesis rules file: ' + GENESIS_RULES);
     }
 
+    // Consensus (whitelisting) operation
+    // TODO(lia): increase this list to 10
+    const whitelistOp = {
+      type: 'SET_VALUE',
+      ref: `/${ConsensusDbPaths.CONSENSUS}/${ConsensusDbPaths.WHITELIST}`,
+      value: {
+        '0x00ADEc28B6a845a085e03591bE7550dd68673C1C': 250,
```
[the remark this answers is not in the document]
liayoo: Thanks for the suggestion, updated the code.

blockchain/genesis_owners.json
liayoo: Done!

blockchain/block.js
liayoo: Replaced with GenesisWhitelist, which uses INITIAL_STAKE.

consensus/index.js
liayoo: Removed.

db/index.js
```diff
@@ -9,10 +9,15 @@ const Functions = require('./functions');
 const RuleUtil = require('./rule-util');
 
 class DB {
-  constructor() {
+  constructor(chainOrBlockNumber) {
```
platfowner, line 12: Is it possible to avoid using one variable for different types? e.g. 

constructor(bc, defaultBlockNumber) {
...
..this.bc = bc;
..this.defaultBlockNumber = defaultBlockNumber;
...
}

new DB(bc);
new DB(null, 100);

lastBlockNumber() {
..if (this.bc) {
....return this.bc.lastBlockNumber();
..}
..return this.defaultBlockNumber;
}

evalFunc(...
....this.lastBlockNumber(), ...);


Using one variable for different types sometimes could make confusion, which can allow future bugs?

liayoo: Agreed - split the variable into two!

start_servers.sh
liayoo: Done

db/index.js
```diff
@@ -744,7 +746,15 @@ class DB {
     const evalFunc = this.makeEvalFunction(ruleString, pathVars);
     return evalFunc(address, data, newData, timestamp, this.getValue.bind(this),
                     this.getRule.bind(this), this.getFunction.bind(this), this.getOwner.bind(this),
-                    new RuleUtil(), ...Object.values(pathVars));
+                    new RuleUtil(), this.lastBlockNumber(), ...Object.values(pathVars));
+  }
+
+  lastBlockNumber() {
+    if (!!this.bc && !!this.bc.lastBlockNumber) {
```
platfowner, line 753: Is !!this.bc.lastBlockNumber really needed?

blockchain/index.js
platfowner: Is this already reverted?

blockchain/block.js
```diff
@@ -282,6 +314,9 @@ class Block {
     const epoch = 0;
     const proposer = ownerAccount.address;
     const validators = {};
+    for (let i = 0; i < ConsensusConsts.INITIAL_NUM_VALIDATORS; i++) {
+      validators[GenesisAccounts.others[i].address] = 250;
```
platfowner, line 318: What about defining this value as a constant?

blockchain/block.js
```diff
@@ -209,14 +218,37 @@ class Block {
       ref: '/',
       value: ownerConfigs
     };
+    const whitelistRuleOp = {
+      type: 'SET_RULE',
+      ref: '/consensus/whitelist',
```
platfowner, line 223: /${ConsensusDbPaths.CONSENSUS}/${ConsensusDbPaths.WHITELIST}?

blockchain/block.js
```diff
@@ -209,14 +218,37 @@ class Block {
...
+      value: `auth === '${ownerAccount.address}'`
+    }
+    const whitelistOwnerOp = {
+      type: 'SET_OWNER',
+      ref: '/consensus/whitelist',
```
platfowner, line 228: ditto

consensus/index.js
```diff
@@ -97,7 +86,7 @@ class Consensus {
     this.epochInterval = setInterval(async () => {
       this.tryFinalize();
       let currentTime = Date.now();
-      if (this.state.epoch % 100 === 0) {
+      if (this.state.epoch % 100 === 0 && HOSTING_ENV === 'gcp') {
```
platfowner, line 89: Question: Is this needed only in gcp env? If so, some comment would be very helpful.

consensus/index.js
```diff
@@ -155,12 +144,8 @@ class Consensus {
...
       logger.error(`[${LOG_PREFIX}:${LOG_SUFFIX}] Empty seedBlock (${this.state.epoch} / ${lastNotarizedBlock.hash})`);
     }
-    let validators;
-    if (nextNumber === 1) {
-      validators = STAKE > 0 ? { [this.node.account.address] :  STAKE } : {};
-    } else {
-      validators = lastNotarizedBlock.validators;
-    }
+
+    const validators = this.node.bc.lastBlockNumber() < 4 ? lastNotarizedBlock.validators : this.getWhitelist();
```
platfowner, line 148: Some comment about number 4 would be very helpful for new readers?

consensus/index.js
```diff
@@ -279,14 +257,8 @@ class Consensus {
...
     const myAddr = this.node.account.address;
-    let validators;
-    if (lastBlock.number === 0) {
-      if (!STAKE) throw Error('First node must stake some AIN');
-      validators = {[myAddr]: STAKE};
-    } else {
-      validators = this.getValidatorsVotedFor(lastBlock.hash);
-    }
-    if (!validators || !(Object.keys(validators).length)) throw Error('No validators voted')
+    const validators = this.node.bc.lastBlockNumber() < 4 ? lastBlock.validators : this.getWhitelist();
```
platfowner, line 260: ditto

blockchain/index.js
liayoo: Yes, it's initializing with an empty array. I think github had a glitch in their system and was showing the old version..

db/index.js
liayoo: Removed the condition.

blockchain/block.js
liayoo: Hmm seem to have missed this.. updated!

blockchain/block.js
liayoo: Done

blockchain/block.js
liayoo: Done

consensus/index.js
liayoo: I was experimenting with it and forgot to remove the condition. Thanks for the question!

consensus/index.js
liayoo: Changed the number to 1, because I realized it doesn't matter if it takes +3 blocks for the block#1 to be finalized. I just need to know if it's included in the finalized chain or not. Plus, I guess there is no real guarantee that block#1 will be finalized by block#4.

liayoo: Oh and also added a comment about this :)

consensus/index.js
liayoo: ditto ditto

blockchain/index.js
platfowner: Now I see the updated. Thanks.

node/index.js
```diff
@@ -12,7 +12,7 @@ class Node {
   constructor() {
     this.bc = new Blockchain(String(PORT));
     this.tp = new TransactionPool();
-    this.db = new DB();
+    this.db = new DB(this.bc, null);
```
platfowner, line 15: just DB(this.bc); would be enough?

node/index.js
```diff
@@ -29,7 +29,7 @@ class Node {
   init(isFirstNode) {
     logger.info(`[${NODE_PREFIX}] Initializing node..`);
     const lastBlockWithoutProposal = this.bc.init(isFirstNode);
-    this.bc.setBackupDb(new DB());
+    this.bc.setBackupDb(new DB(this.bc, null));
```
platfowner, line 32: ditto

liayoo:
Thanks for all your helpful comments! Now merging..

Issue #25, opened 2020-07-27
Tracker server log error


pengin7384:
![image](https://user-images.githubusercontent.com/32831939/88503885-e8f8e780-d00d-11ea-8cb8-a60c6a21d396.png)
There are undefined log messages in tracker server.

pengin7384:
I'm working on this issue

minsulee2:
Can you close the issue if it's been resolved?

pengin7384:
Closed! Thanks

PR #26, opened 2020-07-27
Feature/tracker log

pengin7384:
Done 
- Fix undefined logs
- Change byte formula

resolve: #25

### DIFF
--- a/tracker-server/index.js
+++ b/tracker-server/index.js
@@ -45,7 +45,7 @@ function printNodesInfo() {
   });
   for (let i = 0; i < nodeList.length; i++) {
     const node = nodeList[i];
-    const diskAvailableMb = Math.floor(node.diskUsage.available / 1000000);
+    const diskAvailableMb = Math.floor(node.diskUsage.available / 1024 / 1024);
     const memoryFreeMb = Math.round(node.memoryUsage.free / 1024 / 1024);
     logger.info(`    Node[${i}]: ${node.getNodeSummary()} ` +
         `LastBlock: ${node.lastBlock.number}, ` +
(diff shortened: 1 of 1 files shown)


tracker-server/index.js
```diff
@@ -45,11 +45,14 @@ function printNodesInfo() {
   });
   for (let i = 0; i < nodeList.length; i++) {
     const node = nodeList[i];
-    const diskAvailableMb = Math.floor(node.diskUsage.available / 1000000);
+    const diskAvailableMb = Math.floor(node.diskUsage.available / 1024 / 1024);
```
platfowner, line 48: It seem that MB is acquired by dividing by 1,000,00 and MiB is by dividing by 1024*1.24.
https://en.wikipedia.org/wiki/Megabyte

pengin7384: That's amazing... I thought `1MB = 1024KB`. I edit it!

tracker-server/index.js
```diff
@@ -45,11 +45,14 @@ function printNodesInfo() {
   });
   for (let i = 0; i < nodeList.length; i++) {
     const node = nodeList[i];
-    const diskAvailableMb = Math.floor(node.diskUsage.available / 1000000);
-    const memoryFreeMb = Math.round(node.memoryUsage.free / 1024 / 1024);
+    const diskAvailableMb = Math.floor(node.diskUsage.available / 1000 / 1000);
+    const memoryFreeMb = Math.round(node.memoryUsage.free / 1000 / 1000);
```
minsulee2, line 49: Question: What makes it difference from / 1024 / 1024?

tracker-server/index.js
platfowner: Thanks!

tracker-server/index.js
platfowner: The problem is the difference gets bigger and bigger as the unit become bigger.
So it looks like people use MB and MiB more carefully nowadays: https://en.wikipedia.org/wiki/Megabyte

pengin7384: Some use 1000 and some use 1024... and it created confusion. It seems to created a MiB to solve this problem

pengin7384:
Thanks for reviews!

PR #30, opened 2020-07-30
Feature/consensus/heartbeat

minsulee2:
This will be finished after p2p network updates.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -34,7 +34,8 @@ class P2pServer {
     this.managedPeersInfo = {};
     this.sockets = [];
     this.consensus = new Consensus(this, node);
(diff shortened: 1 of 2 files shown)


server/index.js
```diff
@@ -33,6 +34,7 @@ class P2pServer {
     this.managedPeersInfo = {};
     this.sockets = [];
     this.consensus = new Consensus(this, node);
+    this.isAlive = true;
```
platfowner, line 37: is this variable used?

minsulee2: I commented out this for the next updates!
PTAL

minsulee2:
Thanks for the review!

PR #31, opened 2020-07-30
Feature/consensus/add status apis

liayoo:
Added JSON RPC APIs for consensus states
- net_consensusState
    - Returns minimal consensus state information to determine whether the blockchain is healthy (is creating and finalizing blocks) 
- net_rawConsensusState
    - Dumps the whole consensus state as well as its block pool's state

Also added the identical APIs for testing purposes in client. (/get_raw_consensus_state, /get_consensus_state)


minsulee2:
I really want to see the block-pool status! Thanks for the updates!
LGTM

PR #35, opened 2020-08-04
Feature/consensus/whitelist env var

liayoo:
- Added a NUM_VALIDATORS env var
- Changed prng seed to genesis hash + epoch as a temporary fix
- And some disk i/o optimization and minor fixes


consensus/index.js
```diff
@@ -137,17 +142,12 @@ class Consensus {
...
-    const seedBlock = nextNumber <= ConsensusConsts.MAX_CONSENSUS_STATE_DB ? lastNotarizedBlock
-        : this.node.bc.getBlockByNumber(nextNumber - ConsensusConsts.MAX_CONSENSUS_STATE_DB);
-    if (!seedBlock) {
-      logger.error(`[${LOG_PREFIX}:${LOG_SUFFIX}] Empty seedBlock (${this.state.epoch} / ${lastNotarizedBlock.hash})`);
-    }
     // Need the block#1 to be finalized to have the deposits reflected in the state
     const validators = this.node.bc.lastBlockNumber() < 1 ? lastNotarizedBlock.validators : this.getWhitelist();
-    const seed = seedBlock.hash + this.state.epoch;
+    // FIXME(lia): make the seeds more secure and unpredictable
+    const seed = this.genesisHash + this.state.epoch;
```
platfowner, line 148: It would be clearer if more specific operation is used (string concatenation or number addition).

consensus/index.js
```diff
@@ -423,14 +423,7 @@ class Consensus {
...
     }
-    let seedBlock = number <= ConsensusConsts.MAX_CONSENSUS_STATE_DB ? prevBlock
-        : this.node.bc.getBlockByNumber(number - ConsensusConsts.MAX_CONSENSUS_STATE_DB);
-    if (!seedBlock) {
-      // FIXME: what to do if finalization doesn't happen within MAX_CONSENSUS_STATE_DB blocks?
-      logger.error(`[${LOG_PREFIX}:${LOG_SUFFIX}] No (${number} - ${ConsensusConsts.MAX_CONSENSUS_STATE_DB})th block to calculate the seed from`);
-      return false;
-    }
-    const seed = seedBlock.hash + epoch;
+    const seed = this.genesisHash + epoch;
```
platfowner, line 426: ditto

consensus/index.js
liayoo: You mean something like .concat() ? I think it's recommended to use + instead of .concat for performance reasons. Would it be better if I added '' + at the beginning to make sure they're concatenated as strings?

platfowner: Yes, that would be nice. Thanks.

PR #36, opened 2020-08-04
Bugfix/consensus/reunion

minsulee2:


chain-util.js
```diff
@@ -30,6 +31,10 @@ class ChainUtil {
     return (typeof data === 'object' && data !== null && !Array.isArray(data));
   }
 
+  static isNonEmptyObject(data) {
+    return !!data && ChainUtil.isDict(data) && !_.isEmpty(data);
```
liayoo, line 35: is !_.isEmpty() check needed?

minsulee2: I've seen some cases that only {} comes, it will be in trouble with the if (!VAR) case where VAR is {}.

liayoo: I see! Thanks for the explanation.

PR #37, opened 2020-08-04
Feature/lia/streamlet consensus

liayoo:
- Merging streamlet-consensus branch into master


blockchain/block.js
```diff
@@ -216,14 +218,37 @@ class Block {
...
+    }
 
     // Transaction
     const firstTxData = {
       nonce: -1,
       timestamp,
       operation: {
         type: 'SET',
-        op_list: [ tokenOp, balanceOp, functionsOp, rulesOp, ownersOp ]
+        op_list: [ tokenOp, balanceOp, whitelistValOp, functionsOp, rulesOp, ownersOp, whitelistRuleOp, whitelistOwnerOp ]
```
platfowner, line 251: minor: It would be nice if whitelistRuleOp comes earlier than ownersOp.

chain-util.js
```diff
@@ -32,8 +33,16 @@ class ChainUtil {
     return ruleUtil.isBool(value);
   }
 
-  static isNumber(value) {
-    return ruleUtil.isNumber(value);
+  static isNonEmptyObject(data) {
+    return !!data && ruleUtil.isDict(data) && !_.isEmpty(block);
```
platfowner, line 37: Could we avoid dependency on outer libraries in primitive function as much as possible?
Could ruleUtil.isEmptyNode() be used here?

chain-util.js
```diff
@@ -32,8 +33,16 @@ class ChainUtil {
...
+  }
+
+  static isNonEmptyObject(data) {
```
platfowner, line 40: duplicated?

chain-util.js
minsulee2: Oh, I see. I made it since I've seen some cases that {} are not dealt with.
Shall we remove it for now and discuss about it later then?

chain-util.js
minsulee2: really weird. It needs to be removed.

chain-util.js
liayoo: SGTM, will replace with isEmptyNode() and remove the two isNonEmptyObject()'s.

blockchain/block.js
liayoo: Updated

chain-util.js
liayoo: Removed both (see the comments above).

chain-util.js
platfowner: It looks like this function is still used in a couple of other places. As discussed, could you replace them as well?

liayoo: My bad, I replaced them all but removing the functions themselves wasn't included in the commit. All fixed now!

liayoo:
Thanks for all your help & contributions! Now merging..

Issue #42, opened 2020-08-06
Broken unit tests with "TypeError: Cannot read property 'last_hash' of undefined"


platfowner:
```
./node_modules/mocha/bin/mocha --timeout 320000 -R  spec ./test/tx-pool.test.js

      1) "before each" hook for "transactions are correctly numbered"


  1 passing (56ms)
  1 failing

  1) TransactionPool
       sorting transactions by nonces
         "before each" hook for "transactions are correctly numbered":
     TypeError: Cannot read property 'last_hash' of undefined
      at Function.hasRequiredFields (blockchain/block.js:107:19)
      at Function.parse (blockchain/block.js:99:16)
      at Function.isValidChain (blockchain/index.js:163:30)
      at Function.loadChain (blockchain/index.js:329:20)
      at Blockchain.init (blockchain/index.js:59:33)
      at Node.init (node/index.js:31:46)
      at setDbForTesting (test/test-util.js:11:8)
      at Context.beforeEach (test/tx-pool.test.js:52:7)


./node_modules/mocha/bin/mocha --timeout 320000 -R  spec ./test/transaction.test.js

    1) "before each" hook for "assigns nonces correctly"


  0 passing (19ms)
  1 failing

  1) Transaction
       "before each" hook for "assigns nonces correctly":
     TypeError: Cannot read property 'last_hash' of undefined
      at Function.hasRequiredFields (blockchain/block.js:107:19)
      at Function.parse (blockchain/block.js:99:16)
      at Function.isValidChain (blockchain/index.js:163:30)
      at Function.loadChain (blockchain/index.js:329:20)
      at Blockchain.init (blockchain/index.js:59:33)
      at Node.init (node/index.js:31:46)
      at setDbForTesting (test/test-util.js:11:8)
      at Context.beforeEach (test/transaction.test.js:14:5)

```

platfowner:
Now fixed in develop.

PR #43, opened 2020-08-06
Bugfix/platfowner/fix unit tests

platfowner:
Summary:
- Fix broken unit tests with "TypeError: Cannot read property 'last_hash' of undefined" error
- Set NUM_VALIDATORS env variable when starting servers
- Delete block files before initializing node servers in tests 

#42

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,12 +90,12 @@ yarn install
 ```
 - Run blockchain nodes
 ```
-NUM_VALIDATORS=5 ACCOUNT_INDEX=0 HOSTING_ENV=local DEBUG=false node client/index.js
-NUM_VALIDATORS=5 ACCOUNT_INDEX=1 HOSTING_ENV=local DEBUG=false node client/index.js 
-NUM_VALIDATORS=5 ACCOUNT_INDEX=2 HOSTING_ENV=local DEBUG=false node client/index.js 
-NUM_VALIDATORS=5 ACCOUNT_INDEX=3 HOSTING_ENV=local DEBUG=false node client/index.js 
-NUM_VALIDATORS=5 ACCOUNT_INDEX=4 HOSTING_ENV=local DEBUG=false node client/index.js 
+NUM_VALIDATORS=4 ACCOUNT_INDEX=0 HOSTING_ENV=local DEBUG=false node client/index.js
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -90,11 +90,11 @@ yarn install
...
-ACCOUNT_INDEX=0 HOSTING_ENV=local DEBUG=false node client/index.js
-ACCOUNT_INDEX=1 HOSTING_ENV=local DEBUG=false node client/index.js 
-ACCOUNT_INDEX=2 HOSTING_ENV=local DEBUG=false node client/index.js 
-ACCOUNT_INDEX=3 HOSTING_ENV=local DEBUG=false node client/index.js 
-ACCOUNT_INDEX=4 HOSTING_ENV=local DEBUG=false node client/index.js 
+NUM_VALIDATORS=5 ACCOUNT_INDEX=0 HOSTING_ENV=local DEBUG=false node client/index.js
+NUM_VALIDATORS=5 ACCOUNT_INDEX=1 HOSTING_ENV=local DEBUG=false node client/index.js 
+NUM_VALIDATORS=5 ACCOUNT_INDEX=2 HOSTING_ENV=local DEBUG=false node client/index.js 
+NUM_VALIDATORS=5 ACCOUNT_INDEX=3 HOSTING_ENV=local DEBUG=false node client/index.js 
+NUM_VALIDATORS=5 ACCOUNT_INDEX=4 HOSTING_ENV=local DEBUG=false node client/index.js
```
liayoo, line 97: This should be unnecessary, since the default value is already 5 and plus only the first node needs to specify NUM_VALIDATORS.

start_node_gcp.sh
```diff
@@ -24,6 +24,7 @@ if [ "$2" -lt 0 ] || [ "$2" -gt 4 ]; then
     exit
 fi
 
+export NUM_VALIDATORS=5
```
liayoo, line 27: ditto

README.md
platfowner: Tweaked the case and added some comments about the default value.

start_node_gcp.sh
platfowner: I think having explicit value would help the readability.

Issue #45, opened 2020-08-06
Broken integration tests with new consensus 


platfowner:
It looks like the block generation is stuck by some reasons.

```
./node_modules/mocha/bin/mocha --timeout 320000 -R  spec ./integration/afan_client.test.js


  aFan Client Test
    tx_invest
Initial last block number: 0
      1) "before each" hook for "send_one"
Initial last block number: 0
      2) "after each" hook for "send_one"


  0 passing (11m)
  2 failing

  1) aFan Client Test
       "before each" hook for "send_one":
     Error: Timeout of 320000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/seo/git/comcomai/blockchain/ain-blockchain/integration/afan_client.test.js)
      at module.exports.runLoopOnce (node_modules/deasync/index.js:65:10)
      at module.exports (node_modules/deasync-promise/index.js:13:9)
      at module.exports.interval (node_modules/system-sleep/module.js:24:36)
      at waitForNewBlocks (integration/afan_client.test.js:93:7)
      at set_value.then (integration/afan_client.test.js:117:19)

  2) aFan Client Test
       "after each" hook for "send_one":
     Error: Timeout of 320000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/seo/git/comcomai/blockchain/ain-blockchain/integration/afan_client.test.js)
      at module.exports.runLoopOnce (node_modules/deasync/index.js:65:10)
      at module.exports (node_modules/deasync-promise/index.js:13:9)
      at module.exports.interval (node_modules/system-sleep/module.js:24:36)
      at waitForNewBlocks (integration/afan_client.test.js:93:7)
      at set_value.then (integration/afan_client.test.js:122:19)
      at process._tickCallback (internal/process/next_tick.js:68:7)
      at module.exports.runLoopOnce (node_modules/deasync/index.js:64:10)
      at module.exports (node_modules/deasync-promise/index.js:13:9)
      at module.exports.interval (node_modules/system-sleep/module.js:24:36)
      at waitForNewBlocks (integration/afan_client.test.js:93:7)
      at set_value.then (integration/afan_client.test.js:117:19)

```

minsulee2:
@liayoo has it been updated issue, init?! You can close!

liayoo:
Closing this issue as the problem has been fixed in #47.

Issue #44, opened 2020-08-06
Broken unit tests with new consensus


platfowner:
```
./node_modules/mocha/bin/mocha --timeout 80000 -R  spec ./test/block-pool.test.js

...
    1) Returns a finalizable chain
...
    2) Correctly returns the longest notarized chain


  2 passing (261ms)
  2 failing

  1) BlockPool
       Returns a finalizable chain:
     AssertionError: expected undefined to equal true
      at Context.it (test/block-pool.test.js:126:65)

  2) BlockPool
       Correctly returns the longest notarized chain:

      AssertionError: expected [ Array(1) ] to deeply equal [ Array(1) ]
      + expected - actual

               "0x03AAb7b6f16A92A1dfe018Fe34ee420eb098B98A": 250
               "0x04A456C92A880cd59D7145C457475515a6f6E0f2": 250
             }
           }
      +    {
      +      "epoch": 1
      +      "hash": "0x061063b295628fc0e8bcffee5455c50a8287ccc38909f847f211dc6176bf637d"
      +      "last_hash": "0x559e24b3e9410e522e2804067dc7104fccbdaf84397fb5024645759f2b9df977"
      +      "last_votes": []
      +      "last_votes_hash": "0x853fb99c831d4952ff90b897bd7d7c5c2f3747e8eda8ad13e7359b731eadc299"
      +      "number": 1
      +      "proposer": "0x00ADEc28B6a845a085e03591bE7550dd68673C1C"
      +      "size": 0
      +      "timestamp": 1596701907175
      +      "transactions": []
      +      "transactions_hash": "0x853fb99c831d4952ff90b897bd7d7c5c2f3747e8eda8ad13e7359b731eadc299"
      +      "validators": {
      +        "0x00ADEc28B6a845a085e03591bE7550dd68673C1C": 250
      +      }
      +    }
      +    {
      +      "epoch": 3
      +      "hash": "0x30f1bd1a246ea74e603dce7741c722434884f4c2b853bd4d37299521b8dc65b2"
      +      "last_hash": "0x061063b295628fc0e8bcffee5455c50a8287ccc38909f847f211dc6176bf637d"
      +      "last_votes": []
      +      "last_votes_hash": "0x853fb99c831d4952ff90b897bd7d7c5c2f3747e8eda8ad13e7359b731eadc299"
      +      "number": 2
      +      "proposer": "0x00ADEc28B6a845a085e03591bE7550dd68673C1C"
      +      "size": 0
[...]

PR #47, opened 2020-08-07
Bugfix/tests/fix broken tests

liayoo:
- Fixed broken unit tests (block pool, blockchain)
- Fixed block not producing issue in the integration test
- Fixed the transfer test case in the integration test
- Fixed db snapshot bug in consensus module
- Fixed consensus default paths
- Fixed an edge case where peer nodes are not getting first node's staking tx
- Fixed some testing data

#44 #45

### DIFF
--- a/blockchain/index.js
+++ b/blockchain/index.js
@@ -145,17 +145,20 @@ class Blockchain {
     if (!(newBlock instanceof Block)) {
       newBlock = Block.parse(newBlock);
     }
-    const block = this.chain.shift();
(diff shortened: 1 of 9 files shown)


minsulee2:
Does it include both issues #44 and #45 ?

liayoo:
> Does it include both issues #44 and #45 ?

@minsulee2 Yes, it does.

minsulee2:
@liayoo Ok! I've linked 'em all! Thanks! LGTM 👍

PR #48, opened 2020-08-14
Make smoke test deterministic

liayoo:
- Used waitUntilTxFinalized to make sure all nodes have finalized the target transaction
- Made npm scripts simpler 
- Added tx_hash field in 'set' client API responses

### DIFF
--- a/integration/afan_client.test.js
+++ b/integration/afan_client.test.js
@@ -6,11 +6,11 @@ const sleep = require('system-sleep');
 const spawn = require('child_process').spawn;
 const syncRequest = require('sync-request');
 const AfanClient = require('../afan_client');
-const { BLOCKCHAINS_DIR, TransactionStatus } = require('../constants');
+const { BLOCKCHAINS_DIR } = require('../constants');
+const { waitUntilTxFinalized } = require('../test/test-util');
 const PROJECT_ROOT = require('path').dirname(__filename) + '/../';
 const TRACKER_SERVER = PROJECT_ROOT + 'tracker-server/index.js';
 const APP_SERVER = PROJECT_ROOT + 'client/index.js';
-const LAST_BLOCK_NUMBER_ENDPOINT = '/last_block_number';
 
 const ENV_VARIABLES = [
   {
@@ -85,21 +85,6 @@ describe('aFan Client Test', () => {
     rimraf.sync(BLOCKCHAINS_DIR);
   });
 
-  waitUntilTxFinalized = (txHash) => {
-    const unchecked = new Set(servers);
-    while (true) {
-      if (!unchecked.size) return;
-      unchecked.forEach(server => {
-        const txTracker = JSON.parse(syncRequest('GET', server + '/tx_tracker').body.toString('utf-8')).result;
-        const txStatus = _.get(txTracker, `${txHash}.status`);
-        if (txStatus === TransactionStatus.BLOCK_STATUS) {
-          unchecked.delete(server);
-        }
-      });
-      sleep(1000);
-    }
-  }
-
   set_value = (ref, value) => {
     return Promise.resolve(JSON.parse(syncRequest('POST', server1 + '/set_value',
                            {json: {ref, value}}).body.toString('utf-8')));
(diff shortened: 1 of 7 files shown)


integration/afan_client.test.js
```diff
@@ -83,18 +85,19 @@ describe('aFan Client Test', () => {
...
-          .body.toString('utf-8'))['result'];
-    let updatedLastBlockNumber = initialLastBlockNumber;
-    console.log(`Initial last block number: ${initialLastBlockNumber}`)
-    while (updatedLastBlockNumber < initialLastBlockNumber + numNewBlocks) {
+  waitUntilTxFinalized = (txHash) => {
+    const unchecked = new Set(servers);
+    while (true) {
+      if (!unchecked.size) return;
+      unchecked.forEach(server => {
+        const txTracker = JSON.parse(syncRequest('GET', server + '/tx_tracker').body.toString('utf-8')).result;
```
platfowner, line 93: /tx_tracker could be a very heavy operation as time goes on, so I think we need to add an api for a specific transaction. In this change, could you add a TODO here and in client.js?

integration/afan_client.test.js
```diff
@@ -83,18 +85,19 @@ describe('aFan Client Test', () => {
...
-  waitForNewBlocks = (server, numNewBlocks = 1) => {
-    const initialLastBlockNumber =
-        JSON.parse(syncRequest('GET', server + LAST_BLOCK_NUMBER_ENDPOINT)
-          .body.toString('utf-8'))['result'];
-    let updatedLastBlockNumber = initialLastBlockNumber;
-    console.log(`Initial last block number: ${initialLastBlockNumber}`)
-    while (updatedLastBlockNumber < initialLastBlockNumber + numNewBlocks) {
+  waitUntilTxFinalized = (txHash) => {
+    const unchecked = new Set(servers);
+    while (true) {
```
platfowner, line 90: while (true) looks very computation heavy. what about adding some sleep time between iterations?

integration/afan_client.test.js
```diff
@@ -83,18 +85,19 @@ describe('aFan Client Test', () => {
...
   });
 
-  waitForNewBlocks = (server, numNewBlocks = 1) => {
-    const initialLastBlockNumber =
-        JSON.parse(syncRequest('GET', server + LAST_BLOCK_NUMBER_ENDPOINT)
-          .body.toString('utf-8'))['result'];
-    let updatedLastBlockNumber = initialLastBlockNumber;
-    console.log(`Initial last block number: ${initialLastBlockNumber}`)
-    while (updatedLastBlockNumber < initialLastBlockNumber + numNewBlocks) {
+  waitUntilTxFinalized = (txHash) => {
```
platfowner, line 88: Could this be added as a util function to test-util.js?

integration/afan_client.test.js
liayoo: Thanks! There already was an "ain_getTransactionByHash" json-rpc api, so I just added a /get_transaction in client.js that returns the same information.

integration/afan_client.test.js
liayoo: Hmm, there's a "sleep(1000);" already in the while loop. Do you think there should be more delays / it should be done differently?

integration/afan_client.test.js
liayoo: Great idea! Moved to test-util.js.

liayoo:
Addressed all comments, PTAL.

test/test-util.js
```diff
@@ -1,9 +1,12 @@
 const path = require('path');
-const fs = require("fs")
+const fs = require("fs");
+const syncRequest = require('sync-request');
+const sleep = require('system-sleep');
 const Transaction = require('../tx-pool/transaction');
 const { Block } = require('../blockchain/block');
 const { GenesisAccounts } = require('../constants');
 const { ConsensusDbPaths } = require('../consensus/constants');
+const CURRENT_PROTOCOL_VERSION = require('../package.json').version;
```
platfowner, line 9: minor: not used in this file?

integration/afan_client.test.js
platfowner: sorry, please ignore my previous comment.

test/test-util.js
liayoo: removed

liayoo:
Thanks for the review!

Issue #56, opened 2020-08-19
Genesis sharding config


platfowner:
As a base of the sharding implementation, we need genesis sharding config (see AIP-008 Sharding).

PR #59, opened 2020-08-20
Feature/platfowner/genesis sharding

platfowner:
Summary:
- Support sharding config with genesis_sharding.json
- Support custom genesis configs with GENESIS_CONFIGS_DIR env variable
- Refactor genesis block generation
- Let default genesis config when no custom config is given (minimize code duplication)
- Add getJsObject(), setJsObject() util functions
- Fix bugs in genesis functions / rules / owners config 
- Add tests for sharding

Result:
- We can run shard nodes using e.g.:
```
GENESIS_CONFIGS_DIR=blockchain/afan_shard NUM_VALIDATORS=1 ACCOUNT_INDEX=0 HOSTING_ENV=local DEBUG=false node client/index.js
```

#56


blockchain/block.js
```diff
@@ -329,17 +236,14 @@ class Block {
     // This is a temporary fix for the genesis block. Code should be modified after
     // genesis block broadcasting feature is implemented.
     const ownerAccount = GenesisAccounts.owner;
-    const timestamp = GenesisAccounts.timestamp;
+    const timestamp = Date.now();
```
minsulee2, line 239: Finally! We've got now here :) What about genesisTime or something?

chain-util.js
```diff
@@ -87,12 +87,50 @@ class ChainUtil {
...
     return (formatted.startsWith('/') ? '' : '/') + formatted;
   }
 
+  static getJsObject(obj, path) {
+    if (!ChainUtil.isArray(path)) {
+      return null;
+    }
+    let ref = obj;
+    for (let i = 0; i < path.length; i++) {
+      const key = String(path[i]);
```
minsulee2, line 102: What about using static toString() function here?

chain-util.js
```diff
@@ -87,12 +87,50 @@ class ChainUtil {
...
+    }
+    if (!ChainUtil.isDict(obj)) {
+      return false;
+    }
+    if (path.length == 0) {
+      return false;
+    }
+    let ref = obj;
+    for (let i = 0; i < path.length - 1; i++) {
+      const key = String(path[i]);
```
minsulee2, line 123: ditto

chain-util.js
```diff
@@ -87,12 +87,50 @@ class ChainUtil {
...
+      if (!ChainUtil.isDict(ref[key])) {
+        ref[key] = {};
+      }
+      ref = ref[key];
+    }
+    const key = String(path[path.length - 1]);
```
minsulee2, line 129: ditto

blockchain/block.js
platfowner: Done. Note that the timestamp is used for both block timestamp and transaction timestamp.

chain-util.js
platfowner: Done

chain-util.js
platfowner: ditto

chain-util.js
platfowner: ditto

platfowner:
Comments addressed. PTAL @minsulee2

blockchain/block.js
```diff
@@ -329,18 +236,15 @@ class Block {
     // This is a temporary fix for the genesis block. Code should be modified after
     // genesis block broadcasting feature is implemented.
     const ownerAccount = GenesisAccounts.owner;
-    const timestamp = GenesisAccounts.timestamp;
+    const genesisTime = Date.now();
```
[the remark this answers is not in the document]
minsulee2: Oh, I missed it too. It should be updated at some point tho.

minsulee2 (approved):
LGTM!

platfowner:
Thanks for the quick review. Now merging..

Issue #53, opened 2020-08-19
Database APIs for state proof


platfowner:
We need database APIs for state proof.

We also need client APIs in client/index.js for testing / debugging.

minsulee2:
Thanks for the assigning. I will move on to here after #46 done!

minsulee2:
Done!

PR #61, opened 2020-08-21
Feature/proof api

minsulee2:
Updated getProof func on #53 .
/get_proof route also checked correctly works.

### DIFF
--- a/db/index.js
+++ b/db/index.js
@@ -173,9 +173,8 @@ class DB {
    */
   getProof(dbPath) {
     const fullPath = ChainUtil.parsePath(dbPath);
-    // TODO(minsu): Validate the first label.
-    // TODO(minsu): Implement this.
-    return {};
+    const stateNode = this.getRefForReading(fullPath);
+    return stateNode.getProofHash();
   }
 
   matchFunction(funcPath) {
--- a/db/index.js
+++ b/db/index.js
@@ -174,8 +174,12 @@ class DB {
   getProof(dbPath) {
     const fullPath = ChainUtil.parsePath(dbPath);
     const stateNode = this.getRefForReading(fullPath);
-    if (!!stateNode) return stateNode.getProofHash();
-    else return null;
+    if (stateNode) {
+      return stateNode.getProofHash();
+    }
+    else {
+      return null;
+    }
   }
 
   matchFunction(funcPath) {
--- a/db/index.js
+++ b/db/index.js
@@ -57,6 +57,12 @@ class DB {
     this.writeDatabase([PredefinedDbPaths.RULES_ROOT, ...ChainUtil.parsePath(rulesPath)], rules);
   }
 
+  // For testing purpose only.
+  setFunctionsForTesting(functionsPath, functions) {
+    this.writeDatabase([PredefinedDbPaths.FUNCTIONS_ROOT,
+      ...ChainUtil.parsePath(functionsPath)], functions);
+  }
+
   /**
    * Returns reference to the input path for reading if exists, otherwise null.
    */
--- a/db/index.js
+++ b/db/index.js
@@ -174,7 +174,8 @@ class DB {
   getProof(dbPath) {
     const fullPath = ChainUtil.parsePath(dbPath);
     const stateNode = this.getRefForReading(fullPath);
-    return stateNode.getProofHash();
+    if (!!stateNode) return stateNode.getProofHash();
+    else return null;
   }
 
   matchFunction(funcPath) {
--- a/integration/dev.test.js
+++ b/integration/dev.test.js
@@ -259,12 +259,14 @@ describe('API Tests', () => {
       const valuesProof = valuesBody.result;
       const functionProof = functionsBody.result;
       // TODO(minsu): it needs to be updated after some updates on db/index.js
+      // proof is somehow not updated when new blocks are added. After fixing this bug,
+      // the if statement will be removed.
       if (!valuesProof && !ownersProof && !rulesProof && !functionProof) {
         assert.deepEqual(body, {code: 1, result: null});
       } else {
-        const preimage = `owners${HASH_DELIMITER}${ownersProof}` + `${HASH_DELIMITER}`
-          + `rules${HASH_DELIMITER}${rulesProof}` + `${HASH_DELIMITER}`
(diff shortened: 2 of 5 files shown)


db/index.js
```diff
@@ -173,9 +173,9 @@ class DB {
    */
   getProof(dbPath) {
     const fullPath = ChainUtil.parsePath(dbPath);
-    // TODO(minsu): Validate the first label.
-    // TODO(minsu): Implement this.
-    return {};
+    const stateNode = this.getRefForReading(fullPath);
+    if (!!stateNode) return stateNode.getProofHash();
```
platfowner, line 177: In the case we need to convert the value to boolean e.g. `const isNonNull = !!stateNode;`, this `!!` expression would be useful but if we need to check non-null only, `if (stateNode)` would be just enough?

db/index.js
```diff
@@ -173,9 +173,9 @@ class DB {
...
+    else return null;
```
platfowner, line 178: Could you use brace ({}) for if-else statements even though it's simple?

test/db.test.js
```diff
@@ -2030,18 +2039,14 @@ describe("Test Writedatabase with proof", () => {
     rimraf.sync(BLOCKCHAINS_DIR);
   });
 
-  describe("Check proof", () => {
-    it("checks jsObject is correctly set", () => {
-      assert.deepEqual(node.db.getValue("test"), jsObject);
-    });
-
+  describe("Check proof on writeDatabase and setValue", () => {
```
platfowner, line 2042: minor: setValue means database write so just 'Check proof for setValue()'?

test/db.test.js
```diff
@@ -2030,18 +2039,14 @@ describe("Test Writedatabase with proof", () => {
...
     it("checks proof hash of /values/test", () => {
       const testNode = node.db.getRefForReading(['values', 'test']);
       const childLabels = testNode.getChildLabels();
       const preimage = `${childLabels[0]}${HASH_DELIMITER}`
           + `${testNode.getChild(childLabels[0]).getProofHash()}`;
       const rootHash = ChainUtil.hashString(ChainUtil.toString(preimage));
```
platfowner, line 2048: In db test, you can just use buildProofHashOfStateNode()?

test/db.test.js
```diff
@@ -2051,7 +2056,20 @@ describe("Test Writedatabase with proof", () => {
...
```
platfowner, line 2058: ditto

test/db.test.js
```diff
@@ -2021,7 +2021,16 @@ describe("Test Writedatabase with proof", () => {
     node = new Node();
     setDbForTesting(node);
 
-    jsObject = { level0: { level1: { level2: { foo: 'bar', baz: 'caz' } } } };
+    jsObject = {
+      level0: {
+        level1: {
```
platfowner, line 2026: Could you add another level1 or level2 nodes? This structure seems to be too simple.

test/db.test.js
```diff
@@ -2051,7 +2056,20 @@ describe("Test Writedatabase with proof", () => {
...
+  describe("getProof", () => {
+    it("tests values, owners, rules and functions", () => {
+      const valuesNode = node.db.getRefForReading(['values', 'test']);
+      const ownersNode = node.db.getRefForReading(['owners', 'test']);
+      const rulesNode = node.db.getRefForReading(['rules', 'test']);
+      const functionNode = node.db.getRefForReading(['functions']);
+      expect(node.db.getProof('/values/test')).to.equal(valuesNode.getProofHash());
+      expect(node.db.getProof('/owners/test')).to.equal(ownersNode.getProofHash());
+      expect(node.db.getProof('/rules/test')).to.equal(rulesNode.getProofHash());
+      expect(node.db.getProof('/functions')).to.equal(functionNode.getProofHash());
```
platfowner, line 2072: functions/test?

test/db.test.js
platfowner, line 2042: Currently we're calling only setValue(), could you call setFunction(), setRule(), setOwner() as well?

db/index.js
minsulee2: Yeah, stateNode can be null is promised here. Updated!

db/index.js
minsulee2: Coding convention checked!

test/db.test.js
minsulee2: Updated!

test/db.test.js
minsulee2: Oh, that's correct!

test/db.test.js
minsulee2: Indeed! it was old one. I've just updated!

test/db.test.js
minsulee2: Updated too!

minsulee2:
PTAL. I've added integration test as well !

platfowner (approved):
LGTM
with some comments.

integration/dev.test.js
```diff
@@ -236,6 +239,38 @@ describe('API Tests', () => {
...
+        .body.toString('utf-8'));
+      const functionsBody = JSON.parse(
+        syncRequest('GET', server1 + `/get_proof?ref=/${PredefinedDbPaths.FUNCTIONS_ROOT}`)
+        .body.toString('utf-8'));
+      const ownersProof = ownersBody.result;
+      const rulesProof = rulesBody.result;
+      const valuesProof = valuesBody.result;
+      const functionProof = functionsBody.result;
+      // TODO(minsu): it needs to be updated after some updates on db/index.js
+      if (!valuesProof && !ownersProof && !rulesProof && !functionProof) {
```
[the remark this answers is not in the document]
minsulee2: I added!

integration/dev.test.js
```diff
@@ -236,6 +239,38 @@ describe('API Tests', () => {
...
+        assert.deepEqual(body, {code: 1, result: null});
+      } else {
+        const preimage = `owners${HASH_DELIMITER}${ownersProof}` + `${HASH_DELIMITER}`
```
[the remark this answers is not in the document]
minsulee2: That's right! updated!

minsulee2:
As discussed, I will investigate proof bugs first and then update it as soon as possible.
Thanks for the review!

PR #69, opened 2020-08-26
Bugfix/proof updates

minsulee2:


platfowner (commented):
Nice catch!
Could you also add a test case for makeCopyOfStateTree()?
e.g. when a state tree is copied the proof hash of the root node is also copied.

minsulee2:
PTAL. Not only have I updated makeCopyOfStateTree(), but I also updated setting null when deleting stateTree!!

minsulee2:
PTAL, I am waiting for this :) and conflict resolved!

platfowner (approved):
LGTM
Thanks!

minsulee2:
Thanks for the review! I will merge it as soon as leaving this comment!

PR #72, opened 2020-08-28
Make consensusStatus more concise.

liayoo:
For smaller peer node info in tracker server.

Before:
```
{
  ...
  consensusStatus: {
    consensus; { health, status, epoch },
    blockPool: { ... },
    longestNotarizedChainTips: [ ... ]
  },
  ...
}
```
After:
```
{
  ...
  consensusStatus: {
    health: true,
    status: 1,
    epoch: 123,
    longestNotarizedChainTipsSize: 2
  },
  ...
}
```

### DIFF
--- a/blockchain/index.js
+++ b/blockchain/index.js
@@ -125,6 +125,14 @@ class Blockchain {
     return lastBlock.number;
   }
 
+  lastBlockEpoch() {
+    const lastBlock = this.lastBlock();
+    if (!lastBlock) {
+      return -1;
+    }
+    return lastBlock.epoch;
+  }
+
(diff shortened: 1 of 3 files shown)


platfowner:
Would it be just okay with only epoch without block number?

platfowner (approved):
LGTM
Thanks for fixing this.

one question: Is the epoch in consensusStatus for the block number in lastBlock?

liayoo:
@platfowner Thanks for the review! It's the current epoch, but that reminded me to add lastBlock's epoch to the lastBlock field of node status info 🙂

PR #73, opened 2020-08-31
Feature/proof transaction api

minsulee2:


minsulee2:
I've just added 'ain_getProof' and both 'ain_get' and 'ain_getProof' are checked thru Postman.
If I am on the right track, then I will make some test cases. (As getProof() has built on #61 ain_get also supported now. I've just checked it works). If it is, please leave some comments that I can move on to next step! Thanks!

liayoo:
Are you planning to open another PR for the test cases?

minsulee2:
@liayoo I will build the test cases in this PR and then will ask PTAL

liayoo:
@minsulee2 I see, SGTM! (sounds good to me) please do proceed :)

db/index.js
```diff
@@ -183,14 +183,22 @@ class DB {
...
+    let node = this.stateTree;
     const fullPath = ChainUtil.parsePath(dbPath);
-    const stateNode = this.getRefForReading(fullPath);
-    if (stateNode) {
-      return stateNode.getProofHash();
-    }
-    else {
-      return null;
+    const proof = { proof_hash: node.getProofHash() };
+    let child = proof;
```
platfowner, line 189: It looks like 'child' is actually the proof for the current node. What about renaming it 'proof' and save the root proof as 'rootProof' separately for returning?

db/index.js
```diff
@@ -183,14 +183,22 @@ class DB {
...
   getProof(dbPath) {
+    let node = this.stateTree;
     const fullPath = ChainUtil.parsePath(dbPath);
-    const stateNode = this.getRefForReading(fullPath);
-    if (stateNode) {
-      return stateNode.getProofHash();
-    }
-    else {
-      return null;
+    const proof = { proof_hash: node.getProofHash() };
```
platfowner, line 188: As normal label also can have 'proof_hash' value, it looks like we need to rename it e.g. '.proof_hash'? 
I just updated the design doc accordingly.
And could you define '.proof_hash' as a constant in constants.js? I think you can add StateProofProperties object there.

db/index.js
```diff
@@ -183,14 +183,22 @@ class DB {
    * @param {*} dbPath full database path to the state node to be proved.
    */
   getProof(dbPath) {
```
platfowner, line 185: As https://github.com/ainblockchain/ain-blockchain/pull/75 is already merged, I guess getGenesisStateProofHash() needs to be update as well?

test/db.test.js
```diff
@@ -2122,14 +2118,43 @@ describe("Test proof with database", () => {
...
-      expect(node.db.getProof('/owners/test')).to.equal(ownersNode.getProofHash());
-      expect(node.db.getProof('/rules/test')).to.equal(rulesNode.getProofHash());
-      expect(node.db.getProof('/functions/test')).to.equal(functionNode.getProofHash());
+      const rootNode = node.db.stateTree;
+      const ownersNode = node.db.getRefForReading(['owners']);
+      const rulesNode = node.db.getRefForReading(['rules']);
+      const valuesNode = node.db.getRefForReading(['values']);
+      const functionNode = node.db.getRefForReading(['functions']);
+      const rootProof = { proof_hash: rootNode.getProofHash() };
+      const secondProof = JSON.parse(JSON.stringify(rootProof));
```
platfowner, line 2127: minor: what about 'secondLevelProof'? It wasn't easy for me to catch the meaning at a glance.

platfowner (commented):
Thanks for updating quickly.
Leaving some comments.

db/index.js
platfowner: fyi, we're trying to get references using the constant value e.g. obj[StateProofProperties.PROOF_HASH] in the non-test code, but I think it's just fine to directly referencing it in the test code e.g. obj['.proof_hash']. cc: @liayoo

db/index.js
minsulee2: During the #75 updates, Code review has just done :) Thanks for reminding it. It will be updated right after leaving this comment!

test/db.test.js
minsulee2: Updated!!

db/index.js
minsulee2: Done!

db/index.js
minsulee2: Updated! Now, ProofProperties.PROOF_HASH does that :)

minsulee2:
PTAL! All addressed

test/db.test.js
```diff
@@ -2122,14 +2119,43 @@ describe("Test proof with database", () => {
...
+      functionNode.getChildLabels().forEach(label => {
+        Object.assign(functionsProof.functions,
+          { [label]: { [ProofProperties.PROOF_HASH]: functionNode.getChild(label).getProofHash() } });
+      });
+      assert.deepEqual(rootProof, node.db.getProof('/'));
+      assert.deepEqual(ownersProof, node.db.getProof('/owners/test'));
+      assert.deepEqual(rulesProof, node.db.getProof('/rules/test'));
+      assert.deepEqual(valuesProof, node.db.getProof('/values/test'));
+      assert.deepEqual(functionsProof, node.db.getProof('/functions/test'));
+      assert.deepEqual(null, node.db.getProof('/test/test'));
```
platfowner, line 2158: Can we have a separate 'it' for this case?

db/index.js
platfowner: I'm afraid some tests will break without the update. Could you include it in this PR. (So, early merge has benefits :) )

test/db.test.js
minsulee2: Updated!

db/index.js
minsulee2: Test cases were also added at #75 and I've double checked all correctly running and all test were fine.
Do I need to do something more? (This PR has pulled origin develop branch)

minsulee2:
PTAL!

test/db.test.js
platfowner: Oh, I just meant for the case of non-existing path.
Could you separate it?

db/index.js
platfowner: I mean getGenesisStateProofHash() is assuming that the return value of getProof('/') is just a hash value, but it actually an object now. cc: @liayoo

blockchain/block.js
```diff
@@ -248,7 +249,7 @@ class Block {
         return null;
       }
     }
-    return tempGenesisState.getProof('/');
+    return tempGenesisState.getProof('/')[ProofProperties.PROOF_HASH];
```
minsulee2, line 252: As I mentioned before, I pulled origin develop branch after #75 merged. Now, getProof('/') at getGenesisStateProofHash just returns hash value. cc: @liayoo @platfowner

test/db.test.js
minsulee2: I got this!! Updated!

db/index.js
minsulee2: Please check blockchain/block.js file above. I left a comment :)

blockchain/block.js
platfowner: Oh I see. Thanks for the explanation.

platfowner (approved):
LGTM
Thanks!

minsulee2:
Finally! Thanks for the review!

Issue #74, opened 2020-09-01
Add state proofs to blocks


liayoo:
Now that we have getProof(), we should actually include the state proofs in blocks.

PR #75, opened 2020-09-01
Add stateProof to block.

liayoo:
#74 
- Add stateProof to block
- Check stateProofs when validating proposal blocks
- Update tests
- TODO: check stateProofs in merge()

### DIFF
--- a/blockchain/block.js
+++ b/blockchain/block.js
@@ -62,7 +62,7 @@ class Block {
             number:            ${this.number}
             epoch:             ${this.epoch}
             timestamp:         ${this.timestamp}
-            stateProof:         ${this.stateProof}
+            stateProof:        ${this.stateProof}
             proposer:          ${this.proposer}
             validators:        ${this.validators}
             size:              ${this.size}
--- a/blockchain/block.js
+++ b/blockchain/block.js
@@ -150,19 +150,6 @@ class Block {
     return true;
   }
 
-  static getGenesisProofHash() {
-    const tempGenesisState = new DB(null, -1);
-    const genesisTransactions = Block.getGenesisBlockData(GenesisAccounts[AccountProperties.TIMESTAMP]);
-    for (const tx of genesisTransactions) {
-      const res = tempGenesisState.executeTransaction(tx);
-      if (ChainUtil.transactionFailed(res)) {
-        logger.error(`[${LOG_PREFIX}] Genesis transaction failed:\n${JSON.stringify(tx, null, 2)}\nRESULT: ${JSON.stringify(res)}`)
-        return null;
-      }
-    }
-    return tempGenesisState.getProof('/');
-  }
-
   static getDbSetupTransaction(timestamp, keyBuffer) {
     const opList = [];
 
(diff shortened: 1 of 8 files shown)


blockchain/index.js
```diff
@@ -195,6 +197,26 @@ class Blockchain {
...
+    const tempGenesisState = new DB(null, -1);
+    const genesisTransactions = Block.getGenesisBlockData(GenesisAccounts[AccountProperties.TIMESTAMP]);
+    genesisTransactions.forEach(tx => {
+      const res = tempGenesisState.executeTransaction(tx);
+      if (ChainUtil.transactionFailed(res)) {
+        logger.error(`Genesis transaction failed:\n${JSON.stringify(tx, null, 2)}\nRESULT: ${JSON.stringify(res)}`)
+        return;
+      }
+    });
+    return tempGenesisState.getProof();
```
platfowner, line 217: fyi, getProof() currently just returns one hash value, but it needs to be changed to return partial proof hash of the given path (see https://github.com/ainblockchain/ain-blockchain/issues/76).

platfowner, line 217: Could you add an explicit path value e.g. getProof('/') here and other places?

blockchain/block.js
```diff
@@ -20,7 +20,7 @@ const BlockFilePatterns = require('./block-file-patterns');
 const LOG_PREFIX = 'BLOCK';
 
 class Block {
-  constructor(lastHash, lastVotes, transactions, number, epoch, timestamp, proposer, validators) {
+  constructor(lastHash, lastVotes, transactions, number, epoch, timestamp, stateProof, proposer, validators) {
```
platfowner, line 23: What about 'proofHash' instead here and other places? 'state proof' sound like it's an actual proof not a single hash value (see AIP-007 doc)?

blockchain/index.js
```diff
@@ -195,6 +197,26 @@ class Blockchain {
     return true;
   }
 
+  setGenesisStateProof() {
```
platfowner, line 200: Could you make this a part of block.js?

Is there any circular dependency issue with db/index.js?

platfowner (commented):

blockchain/index.js
minsulee2: I will handle it after some updates!

blockchain/index.js
liayoo: I chose to set genesisStateProof in Blockchain because it can be easily reused, but for now, it seems to be used only in initializing consensus and Blockchain. I don't think there would be any circular dependency issues! Will move getGenesisStateProof() to block.js and remove setGenesisStateProof().

blockchain/index.js
liayoo: done!

blockchain/index.js
liayoo: done!

blockchain/block.js
```diff
@@ -246,9 +263,9 @@ class Block {
     const number = 0;
     const epoch = 0;
     const proposer = ownerAddress;
-   const validators = GenesisWhitelist;
+    const validators = GenesisWhitelist;
     return new this(lastHash, lastVotes, transactions, number, epoch, genesisTime,
-        proposer, validators);
+      stateProof, proposer, validators);
```
platfowner, line 268: As getGenesisStateProof() is a static function, now it can be called here?

blockchain/block.js
platfowner: or stateProofHash?

liayoo: Hmm, saw the second comment a little too late, but I do think having 'state' in front makes it much clearer.

platfowner (approved):
LGTM
Thanks!

blockchain/block.js
liayoo: changed!

blockchain/block.js
liayoo: Thanks, updated.

liayoo:
Thanks for the quick review!

Issue #57, opened 2020-08-19
Shard proof hash reporting


liayoo:
We need a mechanism for periodic proof hash reporting as a part of sharding as well as a native function for updating the latest reported block number (see AIP-008 Sharding).

PR #79, opened 2020-09-02
Feature/sharding/periodic reporting

liayoo:
#57 
- Periodically report state proof hash
- Tightened the write rule at sharding paths
- Added test cases
- Moved waitForNewBlocks to test-util.js
- TODO: better handle the case where the reporter has missed a lot of reports (e.g. when restarting)

### DIFF
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -2,6 +2,7 @@ const seedrandom = require('seedrandom');
 const ainUtil = require('@ainblockchain/ain-util');
 const _ = require('lodash');
 const ntpsync = require('ntpsync');
+const sizeof = require('object-sizeof');
 const logger = require('../logger');
 const { Block } = require('../blockchain/block');
 const BlockPool = require('./block-pool');
@@ -9,10 +10,26 @@ const DB = require('../db');
 const Transaction = require('../tx-pool/transaction');
 const PushId = require('../db/push-id');
 const ChainUtil = require('../chain-util');
-const { DEBUG, WriteDbOperations, PredefinedDbPaths, MessageTypes } = require('../constants');
+const {
+  DEBUG,
+  WriteDbOperations,
+  ReadDbOperations,
+  PredefinedDbPaths,
+  MessageTypes,
+  GenesisSharding,
+  ShardingProperties,
+  ShardingProtocols,
+  MAX_TX_BYTES
+} = require('../constants');
 const { ConsensusMessageTypes, ConsensusConsts, ConsensusStatus, ConsensusDbPaths }
   = require('./constants');
+const { sendTxAndWaitForConfirmation, sendGetRequest } = require('../server/util');
 const LOG_PREFIX = 'CONSENSUS';
+const parentChainEndpoint = GenesisSharding[ShardingProperties.PARENT_CHAIN_POC] + '/json-rpc';
+const shardingPath = GenesisSharding[ShardingProperties.SHARDING_PATH];
+const isShardChain = GenesisSharding[ShardingProperties.SHARDING_PROTOCOL] !== ShardingProtocols.NONE;
+const reportingPeriod = GenesisSharding[ShardingProperties.REPORTING_PERIOD];
+const txSizeThreshold = MAX_TX_BYTES * 0.9;
 
 class Consensus {
   constructor(server, node) {
@@ -25,6 +42,7 @@ class Consensus {
     this.epochInterval = null;
     this.startingTime = 0;
     this.timeAdjustment = 0;
+    this.isShardReporter = false;
     this.state = {
       // epoch increases by 1 every EPOCH_MS, and at each epoch a new proposer is pseudo-randomly selected.
       epoch: 1,
@@ -41,8 +59,14 @@ class Consensus {
       return;
     }
     this.genesisHash = genesisBlock.hash;
+    const myAddr = this.node.account.address;
     try {
-      const myAddr = this.node.account.address;
+      if (isShardChain) {
+        this.isShardReporter = ainUtil.areSameAddresses(
+          GenesisSharding[ShardingProperties.SHARD_REPORTER],
+          myAddr
+        );
+      }
       const currentStake = this.getValidConsensusDeposit(myAddr);
       logger.info(`[${LOG_PREFIX}:${LOG_SUFFIX}] Current stake: ${currentStake}`);
       if (!currentStake) {
@@ -603,6 +627,7 @@ class Consensus {
       }
     }
     this.blockPool.cleanUpAfterFinalization(finalizableChain[finalizableChain.length - 2]);
+    this.reportStateProofHash();
   }
 
   catchUp(blockList) {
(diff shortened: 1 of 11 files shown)


consensus/index.js
```diff
@@ -775,16 +800,78 @@ class Consensus {
...
+    if (this.isReporting) {
+      return;
+    }
+    const lastFinalizedBlock = this.node.bc.lastBlock();
+    const lastFinalizedBlockNumber = lastFinalizedBlock ? lastFinalizedBlock.number : -1;
+    if (lastFinalizedBlock.number % reportingPeriod !== 0) {
+      return;
+    }
+    
+    this.isReporting = true;
```
platfowner, line 819: what about setting the default value in constructor() for readability?

consensus/index.js
```diff
@@ -775,16 +800,78 @@ class Consensus {
...
+    const lastReportedBlockNumber = await this.getLastReportedBlockNumber();
+    let numToReport = lastReportedBlockNumber ? lastReportedBlockNumber + reportingPeriod : 0;
+    const opList = [];
+    while (numToReport <= lastFinalizedBlockNumber) {
```
platfowner, line 823: 'num to report' somehow sounds it's a number of the reports. Could you consider other names e.g. blockNumberToReport?

consensus/index.js
```diff
@@ -775,16 +800,78 @@ class Consensus {
...
+    }
+    const lastFinalizedBlock = this.node.bc.lastBlock();
+    const lastFinalizedBlockNumber = lastFinalizedBlock ? lastFinalizedBlock.number : -1;
+    if (lastFinalizedBlock.number % reportingPeriod !== 0) {
+      return;
+    }
+    
+    this.isReporting = true;
+    const lastReportedBlockNumber = await this.getLastReportedBlockNumber();
+    let numToReport = lastReportedBlockNumber ? lastReportedBlockNumber + reportingPeriod : 0;
```
platfowner, line 821: why not lastReportedBlockNumber + 1?

consensus/index.js
```diff
@@ -775,16 +800,78 @@ class Consensus {
...
+    const opList = [];
+    while (numToReport <= lastFinalizedBlockNumber) {
+      if (sizeof(opList) >= txSizeThreshold) {
+        break;
+      }
+      const block = numToReport === lastFinalizedBlockNumber ?
+        lastFinalizedBlock : this.node.bc.getBlockByNumber(numToReport);
```
platfowner, line 828: minor: 4-space indentation?

consensus/index.js
```diff
@@ -775,16 +800,78 @@ class Consensus {
...
+      if (!block) {
+        logger.error(`[${LOG_PREFIX}] Failed to fetch block of number ${numToReport} while reporting`);
+        break;
+      }
+      opList.push({
+        type: WriteDbOperations.SET_VALUE,
+        ref: `${shardingPath}/${numToReport}/${PredefinedDbPaths.SHARDING_PROOF_HASH}`,
+        value: block.stateProofHash
+      })
+      numToReport += reportingPeriod;
```
platfowner, line 838: Why not += 1? In my memory we were willing to report every block hash while it's period is given from the sharding config?

consensus/index.js
platfowner, line 819: Could you move this statement just after the if (this.isReporting) block for safety?

consensus/index.js
```diff
@@ -775,16 +800,78 @@ class Consensus {
...
+    const tx = {
+      operation: {
+        type: WriteDbOperations.SET,
+        op_list: opList,
+      },
+      timestamp: Date.now(),
+      nonce: -1
+    };
+    try {
+      await sendTxAndWaitForConfirmation(
```
platfowner, line 850: In my memory, we're willing to save the transaction hash to somewhere in the child blockchain (see AIP-008)?
Or could you add a TODO for now?

consensus/index.js
liayoo: thanks, done!

consensus/index.js
liayoo: blockNumberToReport it is!

consensus/index.js
liayoo: Fixed! I mistakenly thought that we only need to report some of the blocks as long as it's periodically done..

consensus/index.js
liayoo: fixed

consensus/index.js
liayoo: Right, thanks! Updated to reflect the design.

consensus/index.js
liayoo: Thanks! Done.

consensus/index.js
liayoo: Left a todo. Will record them at /sharding/reports in child state, but in parent state.. need to think more about this. 
How about we store them at /$sharding_path/$block_number/tx_hash using maybe a native function or a ServerValue like you suggested offline? I think we can create a native function similar to _updateLatestShardReport, or even extend the functionality of the existing function.
So the flow would be:
1. _updateLatestShardReport() triggered by SET_VALUE at /$sharding_path/$block_number/proof_hash
2. Updates /$sharding_path/latest
3. SET_VALUE at /$sharding_path/$block_number/tx_hash

consensus/index.js
```diff
@@ -782,16 +804,87 @@ class Consensus {
...
+    if (!this.isShardReporter) {
+      return;
+    }
+    if (this.isReporting) {
+      return;
+    }
+    this.isReporting = true;
+    const lastFinalizedBlock = this.node.bc.lastBlock();
+    const lastFinalizedBlockNumber = lastFinalizedBlock ? lastFinalizedBlock.number : -1;
+    if (lastFinalizedBlock.number % reportingPeriod !== 0) {
```
minsulee2, line 820: Q: is it promising that lastFinalizedBlock.number rises by one? If it is not, then it seems it may not reach at 0 for sometime.

consensus/index.js
platfowner: Basically, I think we can put the tx_hash stuff out of the Open SW Project's scope as it's a feature for user's convenience, which means it's not urgent for now.
Once we have the SERVER_VALUE feature, I think this function (reportStateProofHash()) can directly set the value on both parent and shard sides as this function is one who knows well about the transaction?

platfowner (approved):
LGTM
Thanks!

consensus/index.js
minsulee2: Oh, it's been updated. Never mind my question!

minsulee2 (approved):
Thru sharding doc and your code, I got the point! LGTM

minsulee2:
Reminder: Is it also included one of milestones?

consensus/index.js
liayoo: Thanks! Updated to check if there's a gap between lastFinalizedBlockNumber and lastReportedBlockNumber that's greater than the reporting period.

consensus/index.js
liayoo: I agree with you about the priorities! We should discuss this + the server value feature sometime in the future then :)

liayoo:
Thanks for the review!
@minsulee2 are you talking about setting a milestone for this PR?

minsulee2:
I saw a lot sharding milestones and thought it is one of em and not included. If it is not, then never mind and mergeeeeee :)

Issue #60, opened 2020-08-21
Support global paths for cross-shard transactions


platfowner:
To read / write database in a shard, the global paths in the requests need to be mapped to local paths e.g.
- sharding path: /apps/afan
- global path: /apps/afan/users/\<uid\>
- local path: /users/\<uid\>

PR #81, opened 2020-09-03
Feature/platfowner/global path

platfowner:
Summary:
- Add is_global option to all DB operations except getProof()
- Add is_global option to all DB read/write APIs except getProof()
- Make sharding is testable in DB level by adding setShardingPath(), setValuesForTesting(), and setShardingForTesting()
- Refactor transaction sanitization logic
- Refactor tests in dev.test.js
- Fix test flakyness by using waitUntilTxFinalized()
- Fix flaky test case by removing unnecessary permission checks

#60

### DIFF
--- a/blockchain/genesis_rules.json
+++ b/blockchain/genesis_rules.json
@@ -60,7 +60,7 @@
     "shard": {
       ".write": false,
       "$sharding_path": {
-        ".write": "(data === null && util.isDict(newData) && util.isString(newData.sharding_path) && util.isString(newData.parent_chain_poc) && util.isNumber(newData.reporting_period) && util.isValAddr(newData.shard_owner) && util.isValAddr(newData.shard_reporter) && util.isValShardProto(newData.sharding_protocol)) || (auth === data.shard_owner && newData === null)"
+        ".write": "(data === null && util.isDict(newData) && util.isString(newData.sharding_path) && util.isString(newData.parent_chain_poc) && util.isNumber(newData.reporting_period) && util.isValAddr(newData.shard_owner) && util.isValAddr(newData.shard_reporter) && util.isValShardProto(newData.sharding_protocol) && evalOwner(newData.sharding_path, 'write_owner', auth) && evalOwner(newData.sharding_path, 'write_rule', auth) && evalOwner(newData.sharding_path, 'write_function', auth)) || (auth === data.shard_owner && newData === null)"
       }
     }
   },
--- a/blockchain/genesis_rules.json
+++ b/blockchain/genesis_rules.json
@@ -60,7 +60,7 @@
     "shard": {
       ".write": false,
       "$sharding_path": {
-        ".write": "(data === null && util.isDict(newData) && util.isString(newData.sharding_path) && util.isString(newData.parent_chain_poc) && util.isNumber(newData.reporting_period) && util.isValAddr(newData.shard_owner) && util.isValAddr(newData.shard_reporter) && util.isValShardProto(newData.sharding_protocol) && evalOwner(newData.sharding_path, 'write_owner', auth) && evalOwner(newData.sharding_path, 'write_rule', auth) && evalOwner(newData.sharding_path, 'write_function', auth)) || (auth === data.shard_owner && newData === null)"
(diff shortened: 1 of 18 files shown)


client/index.js
```diff
@@ -58,31 +58,31 @@ app.get('/', (req, res, next) => {
 });
 
 app.get('/get_value', (req, res, next) => {
-  const result = node.db.getValue(req.query.ref);
+  const result = node.db.getValue(req.query.ref, !!req.query.is_global);
```
minsulee2, line 61: It seems 'false' can also be passed as true. What about using
req.query.is_global === 'true'
or
query-string package like const parsedQuery = query-string.parse(req.query) and then use parsedQuery.is_global

db/index.js
```diff
@@ -154,41 +184,62 @@ class DB {
...
   }
 
   /**
    * Returns a proof of a state node.
    * 
-   * @param {*} dbPath full database path to the state node to be proved.
+   * @param {string} fullPath full database path to the state node to be proved.
    */
-  getProof(dbPath) {
+  // TODO(seo): Consider supporting global path for getProof().
```
minsulee2, line 236: I will think about it! am glad to see TODO here :)

blockchain/genesis_rules.json
```diff
@@ -60,7 +60,7 @@
     "shard": {
       ".write": false,
       "$sharding_path": {
-        ".write": "(data === null && util.isDict(newData) && util.isString(newData.sharding_path) && util.isString(newData.parent_chain_poc) && util.isNumber(newData.reporting_period) && util.isValAddr(newData.shard_owner) && util.isValAddr(newData.shard_reporter) && util.isValShardProto(newData.sharding_protocol) && evalOwner(newData.sharding_path, 'write_owner', auth) && evalOwner(newData.sharding_path, 'write_rule', auth) && evalOwner(newData.sharding_path, 'write_function', auth) && evalRule(newData.sharding_path, null, newData.shard_reporter, currentTime)) || (auth === data.shard_owner && newData === null)"
```
liayoo, line 63: Are the evals removed because they are set by the system?

platfowner: Right, I think it's not necessary any more. Also, I saw some broken tests in the updated test cases.

db/index.js
platfowner: fyi, I tried but it was not straightforward as we have root level labels like 'values' and 'rules' with the input path.

platfowner:
All comments addressed. PTAL @liayoo @minsulee2

blockchain/genesis_rules.json
liayoo: Hmm I actually removed the evalRule() in my recent [PR](https://github.com/ainblockchain/ain-blockchain/pull/79/files#diff-1a37496b59f98bb429371efea6263a6dR63), but I think we still need evalOwner()'s. Even though the transactions are created by the system, it's _signed_ by the shard owner; what's more, anyone can write at /sharding/shard/$sharding_path and create shards without proper permissions at /$sharding_path just by using our transaction API.

db/index.js
minsulee2: Indeed. As getProof() is supposed to deal with path in a global glance, it may not necessarily adopt is_global here.

blockchain/genesis_rules.json
platfowner: You're right. Just reverted the change. Thanks.

platfowner:
PTAL

blockchain/genesis_rules.json
liayoo: Thanks! I'll also look into any test cases breaking.

liayoo:
LGTM. Thanks for the new feature!

liayoo (approved):

test/rule-util.test.js
```diff
@@ -196,6 +196,32 @@ describe("RuleUtil", () => {
     })
   })
 
+  describe("toBool", () => {
```
minsulee2, line 199: Hurray!! Goot to see toBool :)

minsulee2 (approved):
LGTM!

blockchain/genesis_rules.json
platfowner: Ah, I see that the case is gone. Maybe I was misunderstanding something.

PR #95, opened 2020-10-07
add line break in integration.test.js#L439

gmlee329:
As 100byte code limit, add line break in integration.test.js#L439

### DIFF
--- a/integration/integration.test.js
+++ b/integration/integration.test.js
@@ -436,8 +436,8 @@ describe('Integration Tests', () => {
           for (let j = 0; j < len; j++) {
             const block = blocks[j];
             if (block.hash !== hashBlock(block)) {
-              assert.fail(`Block hash is incorrect for block ${JSON.stringify(block, null, 2)}
-                          \n(hash: ${hashBlock(block)}, node ${i})`);
+              assert.fail(`Block hash is incorrect for block ${JSON.stringify(block, null, 2)}` +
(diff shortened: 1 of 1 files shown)


integration/integration.test.js
```diff
@@ -436,7 +436,8 @@ describe('Integration Tests', () => {
           for (let j = 0; j < len; j++) {
             const block = blocks[j];
             if (block.hash !== hashBlock(block)) {
-              assert.fail(`Block hash is incorrect for block ${JSON.stringify(block, null, 2)}\n(hash: ${hashBlock(block)}, node ${i})`);
+              assert.fail(`Block hash is incorrect for block ${JSON.stringify(block, null, 2)}
+                          \n(hash: ${hashBlock(block)}, node ${i})`);
```
minsulee2, line 440: This will show assert message with a double entered form since `` also reads enter as well.
It needs to be ``+`` convention to avoid generating new line thru enter key.

gmlee329: I've modified that part!

platfowner:
```
              assert.fail(`Block hash is incorrect for block ${JSON.stringify(block, null, 2)}\n` +
                  `(hash: ${hashBlock(block)}, node ${i})`);
```
이렇게 바꿔주실 수 있으신가요?

platfowner (approved):
LGTM
Thanks!

PR #97, opened 2020-10-07
Rename Node BlockchainNode to distinguish it from StateNode.

platfowner:
Summary:
- Rename Node BlockchainNode to avoid confusion with StateNode


liayoo (approved):
LGTM, thanks!

PR #100, opened 2020-10-15
Feature/platfowner/global path

platfowner:
Summary:
- Add `.shard: { sharding_enabled: true }` config to the sharding path of the parent blockchain in the shard initialization
- In setValue() (along with incValue() and decValue()) operation, it's checked whether the path-to-write path includes any sharding path
  - If it's in the case, the setValue() operation returns error with isGlobal = false and do nothing with isGlobal = true
  - Otherwise it goes with the normal set value operation process 
  - Writing into the shard config (.shard) is allowed
- Move the path for proof hashes to <sharding_path>/.shard/proof_hash_map from <sharding_path>
- Move config util functions from db/index.js to db/state-util.js
- Add port number of the blockchain node to the prefix of the log messages

Background:
- This is needed for cross-shard token swap implementation: AIP-009

### DIFF
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -26,7 +26,7 @@ const {
 } = require('../constants');
 const { ConsensusMessageTypes, ConsensusConsts, ConsensusStatus, ConsensusDbPaths }
   = require('./constants');
-const { signAndSendTx, sendGetRequest } = require('../server/util');
+const { signAndSendTxList, sendGetRequest } = require('../server/util');
 const { sleep } = require('sleep');
 const LOG_PREFIX = 'CONSENSUS';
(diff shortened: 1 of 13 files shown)


liayoo:
LGTM, thanks! I guess it's not our top priorities right now, but would we need similar functionalities for other set operations (SET_RULE, SET_FUNCTION, SET_OWNER) as well?

liayoo (approved):

platfowner:
@liayoo Right, thanks for the reminder, I was thinking to add a TODO for the task item.

platfowner:
@liayoo  Note that the validity checking logic was slightly changed afterwards to fix a bug with the proof hash reporting, and the path for the proof hashes were changed to <sharding_path>/.shard/proof_hash_map.
PTAL

liayoo:
Thanks for letting me know! Checked the updated code and looks good to me. 👍

PR #102, opened 2020-10-16
Bugfix/remove failed txs

liayoo:
- Remove failed transactions from tx pool (during block creation)
- Update log messages and levels
- Update deploy_gcp.sh to kill all processes at the beginning

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -255,6 +255,7 @@ const TransactionStatus = {
   BLOCK_STATUS: 'BLOCK',
   POOL_STATUS: 'POOL',
   TIMEOUT_STATUS: 'TIMEOUT',
+  FAILED_STATUS: 'FAILED'
 };
 
 /**
--- a/constants.js
+++ b/constants.js
@@ -255,7 +255,7 @@ const TransactionStatus = {
   BLOCK_STATUS: 'BLOCK',
   POOL_STATUS: 'POOL',
   TIMEOUT_STATUS: 'TIMEOUT',
-  FAILED_STATUS: 'FAILED'
+  FAIL_STATUS: 'FAIL'
 };
 
 /**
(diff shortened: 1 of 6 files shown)


tx-pool/index.js
```diff
@@ -175,6 +175,28 @@ class TransactionPool {
...
+    const addrToTxSet = {};
+    txs.forEach(tx => {
+      const { address, hash } = tx;
+      if (!addrToTxSet[address]) {
+        addrToTxSet[address] = new Set();
+      }
+      addrToTxSet[address].add(hash);
+      const tracked = this.transactionTracker[hash];
+      if (tracked && tracked.status !== TransactionStatus.BLOCK_STATUS) {
+        delete this.transactionTracker[hash];
```
platfowner, line 188: Would it be helpful for users to keep them in the trasaction tracker?
Instead, could we tag them with a new status like 'FAILED'?

minsulee2: Oh! Finally, can we get real FAILED status? It would be really great to have it since tx status at insight is always SUCCESS.. (actually hard coded tho haha) I am glad to see them on insight as well.

liayoo: Right, because we don't include failed transactions in blocks like some other blockchains (e.g. ethereum) do.. But for users' sake, I guess keeping failed txs around for a while (until they get timed out) could be useful. Thanks for the suggestion!
Added a FAILED status and updated the transaction APIs as well. PTAL! @platfowner @minsulee2

minsulee2 (approved):
LGTM!

constants.js
```diff
@@ -255,6 +255,7 @@ const TransactionStatus = {
   BLOCK_STATUS: 'BLOCK',
   POOL_STATUS: 'POOL',
   TIMEOUT_STATUS: 'TIMEOUT',
+  FAILED_STATUS: 'FAILED'
```
platfowner, line 258: minor: noun (FAIL or FAILURE) would look more consistent naming as BLOCK, POOL, TIMEOUT are all nouns?

platfowner (approved):
LGTM
with some minor comments.
Thanks!

constants.js
liayoo: Sounds good to me!

liayoo:

Issue #101, opened 2020-10-16
Repeated proof hash report with block number 0 


platfowner:
```
...
2020-10-16T08:50:39.728Z [node-0-8081] info:   ==> Running native function '_updateLatestShardReport' with value '0x2b4edf801e032a4e0cdfaa064660870bf7059f35adbdf0eb0aa3146d7700e5f8', timestamp '1602838239724',
            currentTime '1602838239728', params: {"block_number":"5"} and functionPath: [
  "apps",
  "afan",
  ".shard",
  "proof_hash_map",
  "$block_number",
  "proof_hash"
]
...
2020-10-16T08:50:42.620Z [node-0-8081] info:   ==> Running native function '_updateLatestShardReport' with value '0xb099f0b0e173dfdf6fd1acc9a859d177415ee1b3067762bdc98c17e3a7b86eca', timestamp '1602838239724',
            currentTime '1602838242620', params: {"block_number":"0"} and functionPath: [
  "apps",
  "afan",
  ".shard",
  "proof_hash_map",
  "$block_number",
  "proof_hash"
]
2020-10-16T08:50:42.620Z [node-0-8081] error: [CONSENSUS:createProposal] tx result: failed..
```

platfowner:
Ah, just confirmed that it's actually 'reported' by reportStateProofHashes() in consensus/index.js.

platfowner:
It seems that some unnecessary transactions with block number 0 generated by the failure of getLastReportedBlockNumber(), which is caused by exceptions thrown by axios.post() without any clear reasons.

One finding is that it doesn't happen if we use .then(), e.g.
```
return await axios.post(
  ...
  )
  .then(resp => {
    return resp;
  });
```

PR #103, opened 2020-10-16
Fix axios.post() bug. Set the initial value of proof_hash_map/latest …

platfowner:
Summary:
- Use
```
  return await axios.post(
      ...
  .then(function (resp) {
    return _.get(resp, 'data.result.result');
  });
```
pattern to avoid exceptions with empty {} value
- Set the initial value of proof_hash_map/latest to -1
- Tweak proof hash reporting logic to minimize report conflicts

Background:
- Just found that sometimes axios.post() throws some empty exceptions when used with `const response = await axios.post();` pattern

#101

### DIFF
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -56,6 +56,7 @@ class Consensus {
     // This feature is only used when LIGHTWEIGHT=true.
     this.cache = {};
     this.validatorList = Object.keys(GenesisWhitelist).sort();
+    this.lastReportedBlockNumberSent = -1;
   }
 
   init(lastBlockWithoutProposal, isFirstNode = false) {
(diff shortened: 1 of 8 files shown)


consensus/index.js
```diff
@@ -859,21 +863,21 @@ class Consensus {
     } catch (e) {
       logger.error(`[${LOG_PREFIX}] Failed to report state proof hashes: ${e}`);
     }
+    this.lastReportedBlockNumberSent = lastFinalizedBlockNumber;
```
liayoo, line 866: This may not always be the case, for example, if the size of reports gets too big and we cannot include all the blocks up to lastFinalizedBlockNumber.. although theoretically, that shouldn't happen very often..

server/util.js
```diff
@@ -95,8 +97,7 @@ async function waitUntilTxListFinalize(endpoint, txHashList) {
 async function waitUntilTxFinalize(endpoint, txHash) {
   let numTries = 0;
   while (true) {
-    try {
-      const response = await axios.post(
+    const confirmed = await axios.post(
```
liayoo, line 100: Would this be the pattern you mentioned that we need to avoid?

platfowner (commented):
It's weird that I cannot see a reply button to the first comment, so let me reply here:
Right, for the case there are too may blocks to report, Just changed to set this.lastReportedBlockNumberSent value whenever a block reporting operation is added to the transaction. PTAL

consensus/index.js
platfowner: Right, for the case there are too may blocks to report, Just changed to set this.lastReportedBlockNumberSent value whenever a block reporting operation is added to the transaction. PTAL

server/util.js
liayoo: Ah, I gotcha. It seems like a weird bug on axios part.. Thanks for the explanation!

liayoo:
LGTM. Thanks for handling the issue!

liayoo (approved):

PR #106, opened 2020-10-20
Add placeholders for cross-shard checkin feature

liayoo:
Added native function placeholders, updated rules, and defined some constants.


platfowner (approved):
LGTM
Thanks!

Issue #107, opened 2020-10-20
Need to implement atomic token swap as a POC of cross-shard transactions


platfowner:

platfowner:
Remaining issues:
- Need to integrate with escrow service in checkin
- Need to implement checkout

PR #108, opened 2020-10-20

platfowner:
Summary:
- Add sendCheckinTx.js util script for cross shard transaction testing

#107


tools/cross-shard/sendCheckinTx.js
```diff
@@ -0,0 +1,107 @@
+const _ = require("lodash");
+const axios = require('axios');
+const ainUtil = require('@ainblockchain/ain-util');
+const config = require("./config");
+const ChainUtil = require('../../chain-util');
+const { parentTokenAmount } = require("./config");
```
liayoo, line 6: requiring ./config twice?

tools/cross-shard/sendCheckinTx.js
```diff
@@ -0,0 +1,107 @@
...
+  const endpoint = `${config.endpointUrl}/json-rpc`;
+  const keyBuffer = Buffer.from(config.userPrivateKey, 'hex');
+  const timestamp = Date.now();
+  const payloadTxBody =
+      buildPayloadTx(config.userAddr, config.shardOwnerAddr, config.parentTokenAmount, timestamp);
+  console.log(`payloadTxBody: ${JSON.stringify(payloadTxBody, null, 2)}`);
+  const signedPayloadTx = signTx(payloadTxBody, keyBuffer);
+  console.log(`signedPayloadTx: ${JSON.stringify(signedPayloadTx, null, 2)}`);
+  const triggerTxBody =
+      buildTriggerTx(config.userAddr, JSON.stringify(signedPayloadTx.signedTx), timestamp);
```
liayoo, line 88: Could it be better to use the payload tx object itself instead of the stringified version? Some of the advantages I can think of are less stringifying/parsing & it's slightly easier to verify the payload contents. What do you think?

tools/cross-shard/sendCheckinTx.js
platfowner: Removed. Thanks.

tools/cross-shard/sendCheckinTx.js
platfowner: Agree that there are no evident advantages of stringifying the payload. Just done. Thanks.

platfowner:

liayoo (approved):
LGTM, thanks for making this tool!

PR #109, opened 2020-10-21

platfowner:
Summary:
- Add is_confirmed, confirmed_at properties to transactionTracker
- Tweak return value of get_transaction client API and ain_getTransactionByHash json rpc API
- Add transaction confirmation logic to sendCheckinTx.js tool script

#107


liayoo (approved):
LGTM, thanks!

PR #110, opened 2020-10-22

platfowner:
Summary:
- Add addRemoteTransaction(), checkRemoteTransactions(), doAction() to check tx confirmation (tx-pool/index.js)
- Add payload transaction forwarding logic (db/functions.js)
- Simplify API call util functions (server/util.js)
- Rename is_confirmed -> is_finalized

Limitation:
- Currently writeDbAndTriggerFunctions() is used in doAction() in tx-pool/index.js and in this case action results may not remain in DB in multi-node shard cases (more specifically when the shard block proposer is not the shard reporter). One solution is to replace writeDbAndTriggerFunctions() with real transaction to itself. @liayoo 

#107

### DIFF
--- a/json_rpc/index.js
+++ b/json_rpc/index.js
@@ -154,7 +154,10 @@ module.exports = function getMethods(
         const index = Number(args.index);
         const block = node.bc.getBlockByHash(args.block_hash);
         if (block.transactions.length > index && index >= 0) {
-          result = Object.assign({}, block.transactions[index], { is_finalized: true });
+          result = {
+            transaction: block.transactions[index],
+            is_finalized: true
+          };
         }
       }
       done(null, addProtocolVersion({ result }));
@@ -166,7 +169,10 @@ module.exports = function getMethods(
         const index = Number(args.index);
         const block = node.bc.getBlockByNumber(args.block_number);
         if (block.transactions.length > index && index >= 0) {
-          result = Object.assign({}, block.transactions[index], { is_finalized: true });
+          result = {
+            transaction: block.transactions[index],
+            is_finalized: true
+          };
         }
       }
       done(null, addProtocolVersion({ result }));
--- a/json_rpc/index.js
+++ b/json_rpc/index.js
@@ -154,7 +154,7 @@ module.exports = function getMethods(
         const index = Number(args.index);
         const block = node.bc.getBlockByHash(args.block_hash);
         if (block.transactions.length > index && index >= 0) {
-          result = Object.assign({}, block.transactions[index], { is_confirmed: true });
+          result = Object.assign({}, block.transactions[index], { is_finalized: true });
         }
       }
       done(null, addProtocolVersion({ result }));
@@ -166,7 +166,7 @@ module.exports = function getMethods(
         const index = Number(args.index);
         const block = node.bc.getBlockByNumber(args.block_number);
         if (block.transactions.length > index && index >= 0) {
-          result = Object.assign({}, block.transactions[index], { is_confirmed: true });
+          result = Object.assign({}, block.transactions[index], { is_finalized: true });
         }
       }
       done(null, addProtocolVersion({ result }));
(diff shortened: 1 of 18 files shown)


platfowner:
@liayoo PR updated as discussed offline. PTAL

server/util.js
```diff
@@ -13,12 +13,12 @@ const ChainUtil = require('../chain-util');
 
 const CURRENT_PROTOCOL_VERSION = require('../package.json').version;
 
-async function sendTxListAndWaitForConfirmation(endpoint, txList, keyBuffer) {
-  const res = await signAndSendTxList(endpoint, txList, keyBuffer);
-  if (res.errMsg || !res.txHashList) {
+async function sendTxAndWaitForConfirmation(endpoint, tx, keyBuffer) {
```
pengin7384, line 16: How about 'sendTxAndWaitForFinalization'?

tx-pool/index.js
```diff
@@ -41,7 +51,7 @@ class TransactionPool {
       address: tx.address,
       index: this.transactions[tx.address].length - 1,
       timestamp: tx.timestamp,
-      is_confirmed: false,
+      is_finalized: false,
       confirmed_at: -1,
```
pengin7384, line 55: 'finalized_at' seems to be more consistent. What do you think?

json_rpc/index.js
```diff
@@ -154,7 +154,7 @@ module.exports = function getMethods(
         const index = Number(args.index);
         const block = node.bc.getBlockByHash(args.block_hash);
         if (block.transactions.length > index && index >= 0) {
-          result = Object.assign({}, block.transactions[index], { is_confirmed: true });
+          result = Object.assign({}, block.transactions[index], { is_finalized: true });
```
liayoo, line 157: (minor) We could actually remove is_finalized here since if we're looking for a tx in a block, it means it's finalized. Or, we could keep the info to match this result to ain_getTransactionByHash's but change to follow this structure? { transaction: {...}, is_finalized: true }

liayoo (approved):

server/util.js
minsulee2: Any difference between finalization and confirmation??

server/util.js
```diff
@@ -38,78 +38,46 @@ function signTx(tx, keyBuffer) {
...
-      const { txHash, signedTx } = signTx(tx, keyBuffer);
-      signedTxList.push(signedTx);
-      txHashList.push(txHash);
-    }
-    params = {
-      protoVer: CURRENT_PROTOCOL_VERSION,
-      tx_list: signedTxList,
-    };
-  }
+async function sendSignedTx(endpoint, signedTxParams) {
```
minsulee2, line 41: Can I ask why it changed to single signedTx from txList?

json_rpc/index.js
minsulee2: but still chance not to finalize?

server/util.js
platfowner: It was once generalized in my previous but reverted as there is not clear use case and its complexity to maintain.

server/util.js
platfowner: @pengin7384 Done.
They are the same AFAIK except finalization is a kind of official term in Stream protocol.

server/util.js
minsulee2: Thanks for the explain!

tx-pool/index.js
platfowner: @pengin7384 Done. Thanks.

json_rpc/index.js
platfowner: @liayoo Just tried to remove is_finalized at all but resulted choosing option 2 after realizing is_finalized == true is very useful from the end user's point of view than status == 'BLOCK_STATUS'.

platfowner:
Comments address. PTAL

pengin7384:
LGTM! Thanks 👍

json_rpc/index.js
liayoo: @platfowner I like option 2 as it's consistent with other get tx apis. Thank you!
@minsulee2 these particular methods look for txs only in the finalized chain. It would be quite difficult to check other unfinalized blocks as well..

platfowner:

PR #111, opened 2020-10-22
Feature/sharding/cross shard checkin

liayoo:
- Implement openCheckin(), closeCheckin() native functions
- Replace writeDbAndTriggerFunctions() with a real transaction

#107

### DIFF
--- a/db/functions.js
+++ b/db/functions.js
@@ -9,10 +9,16 @@ const {
   DefaultValues,
   ShardingProperties,
   GenesisSharding,
-  ACCOUNT_INDEX,
+  WriteDbOperations,
+  ShardingProtocols,
+  GenesisAccounts,
+  AccountProperties,
+  TokenExchangeSchemes,
+  PORT,
 } = require('../constants');
 const ChainUtil = require('../chain-util');
-const { sendSignedTx } = require('../server/util');
+const { sendSignedTx, signAndSendTx } = require('../server/util');
+const Transaction = require('../tx-pool/transaction');
 
 const parentChainEndpoint = GenesisSharding[ShardingProperties.PARENT_CHAIN_POC] + '/json-rpc';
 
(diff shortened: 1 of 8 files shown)


tx-pool/index.js
```diff
@@ -341,18 +344,25 @@ class TransactionPool {
...
-    const fullPath = db.getFullPath(parsedPath, PredefinedDbPaths.VALUES_ROOT);
-    const valueCopy = ChainUtil.isDict(action.value) ?
-        JSON.parse(JSON.stringify(action.value)) : action.value;
-    valueCopy[PredefinedDbPaths.REMOTE_TX_ACTION_RESULT] = {
-      code: success ? FunctionResultCode.SUCCESS : FunctionResultCode.FAILURE
+  doAction(action, success) {
+    const triggerTx = action.transaction;
+    const valueWithResultCode = Object.assign(
+          ChainUtil.isDict(action.value) ? action.value : { value: action.value },
+          { code: success ? FunctionResultCode.SUCCESS : FunctionResultCode.FAILURE });
```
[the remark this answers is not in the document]
platfowner: If we use the txHash included in the path, I think this can be just removed?

tx-pool/index.js
```diff
@@ -341,18 +344,25 @@ class TransactionPool {
...
+      timestamp: triggerTx.timestamp,
+      nonce: -1
     };
-    const transaction = action.transaction;
-    db.writeDbAndTriggerFunctions(
-        parsedPath, fullPath, valueCopy, transaction.timestamp, Date.now(), transaction);
+    const ownerPrivateKey = ChainUtil.getJsObject(
+      GenesisAccounts, [AccountProperties.OWNER, AccountProperties.PRIVATE_KEY]);
+    const keyBuffer = Buffer.from(ownerPrivateKey, 'hex');
+    const endpoint = `http://localhost:${PORT}/json-rpc`;
```
platfowner, line 364: ditto

db/functions.js
```diff
@@ -214,38 +220,152 @@ class Functions {
...
-          transaction: payloadTx.transaction,
+          timestamp: Date.now(),
+          nonce: -1
         };
-        this.tp.addRemoteTransaction(txHash, action);
-      });
-    } else {
-      logger.info(`  =>> Skip sending signed transaction to the parent blockchain: ${txHash}`);
+    // Sign and send transferTx to the node itself
+    const endpoint = `http://localhost:${PORT}/json-rpc`;
```
platfowner, line 318: We need a todo for this?

platfowner (commented):
Nice job done!
Leaving some suggestions and comments.

tx-pool/index.js
```diff
@@ -341,18 +344,25 @@ class TransactionPool {
...
+      operation: {
+        type: WriteDbOperations.SET_VALUE,
+        ref: action.ref,
+        value: valueWithResultCode
+      },
+      timestamp: triggerTx.timestamp,
+      nonce: -1
     };
-    const transaction = action.transaction;
-    db.writeDbAndTriggerFunctions(
```
platfowner, line 354: now writeDbAndTriggerFunctions() can be removed?

db/functions.js
platfowner: Just created a PR to make the node url available from functions and tx pool: https://github.com/ainblockchain/ain-blockchain/pull/112

tx-pool/index.js
```diff
@@ -341,18 +344,25 @@ class TransactionPool {
...
-    const parsedPath = ChainUtil.parsePath(action.ref);
-    const fullPath = db.getFullPath(parsedPath, PredefinedDbPaths.VALUES_ROOT);
-    const valueCopy = ChainUtil.isDict(action.value) ?
-        JSON.parse(JSON.stringify(action.value)) : action.value;
-    valueCopy[PredefinedDbPaths.REMOTE_TX_ACTION_RESULT] = {
-      code: success ? FunctionResultCode.SUCCESS : FunctionResultCode.FAILURE
+  doAction(action, success) {
+    const triggerTx = action.transaction;
+    const valueWithResultCode = Object.assign(
+          ChainUtil.isDict(action.value) ? action.value : { value: action.value },
```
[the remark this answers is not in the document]
liayoo: Right, haha. Then actions would be changed like so? 
as-is: { ref: `/checkin/<userAddr>/<id>/parent_finalize`, value: { txHash } }
to-be: { ref: `/checkin/<userAddr>/<id>/parent_finalize/<txHash>`, value: true } & doAction() always writes at `<actionRef>/<functionResult>` without modifying the action.value. Correct me if I misunderstood your suggestion!

tx-pool/index.js
liayoo: Removed!

db/functions.js
liayoo: Thanks, will change to use the url once the pr is merged!

tx-pool/index.js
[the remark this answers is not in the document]
platfowner: One thing here is that 'value' actually changes depending on whether the transaction was finalized or not (failed or timed out). Maybe we can just pass value generating function e.g.

(success) => {
  return {
    tx_hash: txHash,
    result: !!success,
  };
}

?

tx-pool/index.js
platfowner: Thanks!

db/functions.js
platfowner: Just merged!

tx-pool/index.js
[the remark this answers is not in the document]
liayoo: Ah, I see. I was going to change the writing path depending on the tx result, but you're saying we should keep the ref as-is, use txHash and result as the value? 
Should we then remove the value field from action?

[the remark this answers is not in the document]
platfowner: @liayoo 
Ah, I was confused a little bit, sorry.
Maybe we can just keep changing the ref by adding the tx hash with simpler value generating function like
```
(success) => {
  return !!success;
}
```
?

Yes, this function replaces the value parameter.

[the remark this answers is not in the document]
liayoo: I gotcha. I think we're now on the same page. Let me update this PR accordingly!

tx-pool/index.js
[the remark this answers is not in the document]
liayoo: done

tx-pool/index.js
liayoo: replaced with `urlInternal`

liayoo:

tx-pool/index.js
```diff
@@ -328,31 +337,45 @@ class TransactionPool {
...
-    const valueCopy = ChainUtil.isDict(action.value) ?
-        JSON.parse(JSON.stringify(action.value)) : action.value;
-    valueCopy[PredefinedDbPaths.REMOTE_TX_ACTION_RESULT] = {
-      code: success ? FunctionResultCode.SUCCESS : FunctionResultCode.FAILURE
+  doAction(action, txHash, success) {
+    const triggerTx = action.transaction;
+    const actionTx = {
+      operation: {
+        type: WriteDbOperations.SET_VALUE,
+        ref: this._getRemoteTxActionResultPath(
```
platfowner, line 365: Can we have business logic is handled by the native functions (_openCheckin() & _closeCheckin()) as much as possible? e.g. letting the final value of 'ref' and the value function is passed by _openCheckin() so action would look like: 
```
{
  ref: '.../parent_final/<txHash>',
  valueFunction: (succes) => {
    result: !!success
  },
  transaction: ...
}
```

platfowner (approved):
LGTM
with some comments.
Thanks!

tx-pool/index.js
liayoo: Thanks! That looks much better. Updated as you described :)

liayoo:

PR #112, opened 2020-10-22
Make Node URL available.

platfowner:
Summary:
- Make Node URL available from transaction pool and functions
- Fix undefined 'action' error

#107

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -13,6 +13,7 @@ const { ConsensusStatus } = require('../consensus/constants');
 const { Block } = require('../blockchain/block');
 const Transaction = require('../tx-pool/transaction');
 const {
+  PORT,
   P2P_PORT,
   TRACKER_WS_ADDR,
   HOSTING_ENV,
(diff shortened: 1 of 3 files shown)


server/index.js
```diff
@@ -173,10 +173,15 @@ class P2pServer {
     })
     .then((ipAddr) => {
       this.ipAddress = ipAddr;
+      this.node.setNodeUrl(this.getNodeUrl(ipAddr))
```
liayoo, line 176: Actually, when I tested, I needed to use an internal IP address for HOSTING_ENV === 'gcp'. Here's an URL we can get the IP from: http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/ip

liayoo (commented):
Thanks for the changes! Left a comment regarding the gcp env, PTAL.

server/index.js
platfowner: Thanks, just made both internal and external addresses are available.

platfowner:
PTAL @liayoo

liayoo (approved):
LGTM, thanks!

PR #115, opened 2020-10-26
Feature/sharding/cross shard checkin

liayoo:
- Set is_global to true for check-in related transactions
- tools/cross-shard/config.js → tools/cross-shard/config_local.js
- Add tools/cross-shard/config_gcp.js as an example
- Fix bugs in shard reporting and native functions finalized state check


platfowner (approved):
LGTM
Thanks!

Issue #128, opened 2020-11-02
Better logging with module prefixes


liayoo:
It would be nice to have a programmatic way to set prefixes for logs. Couldn't find an easy way to set each function's name, but I think we can set the file names at least.

For example:
logger.info(\`[${P2P_PREFIX}] Listening to peer-to-peer connections on: ${P2P_PORT}\n\`);
=> logger.info(\`Listening to peer-to-peer connections on: ${P2P_PORT}\n\`);

Refs:
https://stackoverflow.com/questions/42858585/add-module-name-in-winston-log-entries
https://stackoverflow.com/questions/13410754/i-want-to-display-the-file-name-in-the-log-statement

platfowner:
Resolved.

PR #130, opened 2020-11-02
Init logger with log prefix.

platfowner:
Summary:
- Initialize logger with log prefix

#128

### DIFF
--- a/logger/index.js
+++ b/logger/index.js
@@ -1,12 +1,28 @@
 const winston = require('winston');
 const { getWinstonLevels, getWinstonColors, getWinstonTransports } = require('./winston-util');
 
-const logger = new winston.createLogger({
+const winstonLogger = new winston.createLogger({
   levels: getWinstonLevels(),
   transports: getWinstonTransports(),
   exitOnError: false
 });
 
 winston.addColors(getWinstonColors());
+
+const logger = function(prefix) {    
+  const prefixedLogger = {
+      error: function(text) {
+          winstonLogger.error(`[${prefix}] ` + text)
+      },
+      info: function(text) {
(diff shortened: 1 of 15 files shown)


logger/index.js
```diff
@@ -1,12 +1,28 @@
...
   transports: getWinstonTransports(),
   exitOnError: false
 });
 
 winston.addColors(getWinstonColors());
 
+const logger = function(prefix) {    
+  const prefixedLogger = {
+      error: function(text) {
+          winstonLogger.error(`[${prefix}] ` + text)
```
liayoo, line 15: (minor) How about including the text in the string like this? \`[${prefix}] ${text}\`

logger/index.js
```diff
@@ -1,12 +1,28 @@
...
+      },
+      info: function(text) {
+          winstonLogger.info(`[${prefix}] ` + text)
```
liayoo, line 18: ditto

logger/index.js
```diff
@@ -1,12 +1,28 @@
...
+      },
+      debug: function(text) {
+          winstonLogger.debug(`[${prefix}] ` + text)
```
liayoo, line 21: ditto

liayoo (approved):
LGTM, with some minor comments. Thanks!

logger/index.js
platfowner: Done. Thanks!

logger/index.js
platfowner: ditto

logger/index.js
platfowner: ditto

platfowner:
All comments addressed. Thanks. Now merging..

PR #138, opened 2020-11-11
Release/0.3.3

minsulee2:
Now, we can run both 0.3.2 and 0.3.3 together! Hurray!


client/protocol_versions.json
```diff
@@ -1,5 +1,6 @@
 {
-  "0.3.2": {
-    "min": "0.3.2"
+  "0.3": {
```
platfowner, line 2: What about just keeping the old config and adding a config for 0.3.3? 
And leaving max value as blank seems to be just fine because we don't know about higher version yet.
@liayoo Please correct me if I'm wrong.

platfowner (approved):
LGTM
with some minor comments.
Thanks!

client/protocol_versions.json
liayoo: I agree. I don't see a benefit of combining the configs for 0.3.2 and 0.3.3 as 0.3 at this version, and we've decided to leave max values empty until we have updates that make a newer version incompatible with older versions. Maybe the decision wasn't made clear in previous meetings.. Sorry for the confusion!

liayoo (approved):
LGTM, thanks!

client/protocol_versions.json
minsulee2: I've fixed it! Thanks for the reviews!!

minsulee2:
Thanks for the review! I am merging it now!

PR #146, opened 2020-11-29
Update deploy & setup scripts.

liayoo:
- Updated gcp deploy & setup scripts
- Set default LIGHTWEIGHT value to false


deploy_gcp.sh
```diff
@@ -22,7 +22,7 @@ echo "PROJECT_ID=$PROJECT_ID"
 GCP_USER="$2"
 echo "GCP_USER=$GCP_USER"
 
-FILES_FOR_TRACKER="setup_tracker_gcp.sh setup_ubuntu.sh start_tracker_gcp.sh tracker-server/"
+FILES_FOR_TRACKER="blockchain/ chain-util.js consensus/ constants.js db/ logger/ package.json setup_tracker_gcp.sh setup_ubuntu.sh start_tracker_gcp.sh tracker-server/"
```
platfowner, line 25: Do we need to deploy these folders or files for tracker server?

liayoo: Unfortunately, yes.. We need them because the tracker server is now sharing the logger code with nodes. It would be better if we could just copy the files we need (e.g. ./consensus/constants.js instead of ./consensus), but scp wasn't creating the folders for the nested files.

platfowner: If so, we can just remove tracker-server/package.json now?

liayoo: Ah, that's a great idea! Just removed package.json & yarn.lock from ./tracker-server, and updated the setup_tracker_gcp.sh script.

platfowner (approved):
LGTM
Thanks!

PR #147, opened 2020-11-30
Feature/state version control/consensus

liayoo:
- Remove redundant backupDb and replace it with finalizedVersion
- By doing so, we avoid executing transactions to backupDb every time we finalize a block, since we already have a state constructed during the consensus process


platfowner (approved):
LGTM
Thanks!

PR #150, opened 2020-12-03
Fix decreaseNumRef() error. Tweak state node API.

platfowner:
Summary:
- Fix decreaseNumRef() error.
- Tweak state node API.

#143


db/state-util.js
```diff
@@ -246,10 +253,15 @@ function deleteStateTreeVersion(root, version) {
...
       numNodes += deleteStateTreeVersion(childNode, version);
+    } else if (childNode.getNumRef() < 1) {
+      // This shouldn't happen.
+      logger.error(
+          `[${LOG_HEADER}] Invalid numRef value: ${childNode.getNumRef()} with label: ${label}.`);
     }
     root.deleteChild(label);
   }
-  root.reset();
+  root.resetValue();
```
liayoo, line 263: do we need a check for root.numRef == 1 here?

platfowner: Added. Thanks for the comment.

liayoo (approved):
LGTM, thanks!!

PR #153, opened 2020-12-05

platfowner:
Before:
- With multiple state versions in state manager, proof hash values on the paths to roots of different state versions were not updated when database is written

After:
- The proof hash values on all the paths to the affected roots are updated. 

Summary:
- Update proof hash for all paths to roots with different versions when database is written
- Add treeSize APIs to state node and client
- Add parent APIs to state node
- Add equal APIs to state node
- Move state tree <-> js object conversion functions to StateNode class
- Add proofHash and treeSize versions to the output of toJsObject(true)
- Remove some duplicated test cases

#143


db/state-util.js
```diff
@@ -284,45 +283,69 @@ function makeCopyOfStateTree(root) {
...
   }
-  return ChainUtil.hashString(ChainUtil.toString(preimage));
+  if (!root1.equal(root2)) {
+    return false;
+  }
+  root1.getChildLabels().forEach((label) => {
+    const child1 = root1.getChild(label);
+    const child2 = root2.getChild(label);
+    if (!equalStateTrees(child1, child2)) {
+      return false;
```
liayoo, line 300: I don't think returning false in forEach callback would return false for the equalStateTrees function..

platfowner: Right. Fixed and added test cases. Thanks!

platfowner:

And note that some additional changes were made:
- Move state tree <-> js object conversion functions to StateNode class
- Add proofHash and treeSize versions to the output of toJsObject(true)

db/state-node.js
```diff
@@ -122,23 +226,38 @@ class StateNode {
...
+    return ChainUtil.hashString(ChainUtil.toString(preimage));
+  }
+  
+  computeTreeSize() {
+    if (this.getIsLeaf()) {
+      return 1;
     } else {
-      // This shouldn't happen.
-      logger.error(`[${LOG_HEADER}] Failed to decrease numRef value: ${this.numRef}.`);
+      return this.getChildNodes().reduce((acc, cur) => acc + cur.getTreeSize(), 1);
```
minsulee2, line 253: the init value 1 of reduce means the node itself?

minsulee2 (approved):
LGTM with minor comment

db/state-node.js
platfowner: Right. Also it indicates that numeric addition (+) should be applied to the values instead of others (e.g. string concatenation). Actually, if we skip this like:
`return 1 + this.getChildNodes().reduce((acc, cur) => acc + cur.getTreeSize());`

It returns string concatenation results.

node/index.js
```diff
@@ -139,6 +139,10 @@ class BlockchainNode {
     this.syncDb(nodeVersion)
   }
 
+  dumpFinalizedStates(withDetails) {
```
liayoo, line 142: would it be better to make this singular since there's only one finalized state?

db/state-node.js
```diff
@@ -80,13 +180,17 @@ class StateNode {
   }
 
   deleteChild(label) {
```
liayoo, line 182: do we need to update treeSize here or somewhere that calls this function?

platfowner: As treeSize updates and proofHash updates are relatively high-cost operations, they're done from higher level all at once (see ref code below).
https://github.com/ainblockchain/ain-blockchain/blob/master/db/index.js#L182
https://github.com/ainblockchain/ain-blockchain/blob/master/db/index.js#L185

liayoo: Ah, I see. Thanks for the explanation!

node/index.js
platfowner: As discussed offline, Just renames this function dumpFinalizedVersion() and /finalized_states client API /dump_finalized_version.

platfowner:
@liayoo Comments addressed. PTAL

liayoo (approved):
LGTM, thanks for the great work!

Issue #157, opened 2020-12-13
[CONSENSUS] Error while creating a proposal: ReferenceError: txOps is not defined


platfowner:
```
{
  "insertId": ".......1IIz1VzEiQAh3mtLnPbJywZcf",
  "jsonPayload": {
    "metadata": {},
    "message": "[CONSENSUS] Error while creating a proposal: ReferenceError: txOps is not defined"
  },
  "resource": {
    "type": "gce_instance",
    "labels": {
      "instance_id": "4460521773284006620",
      "zone": "asia-east1-b",
      "project_id": "testnet-dev-ground"
    }
  },
  "timestamp": "2020-12-13T19:50:00.316999912Z",
  "severity": "ERROR",
  "logName": "projects/testnet-dev-ground/logs/winston_log",
  "receiveTimestamp": "2020-12-13T19:50:00.373285380Z"
}
```

platfowner:
Fixed by #158.

PR #158, opened 2020-12-13
Fix txOps is not defined bug.

platfowner:
Summary:
- Fix 'txOps is not defined' bug.
- Add /protocol_versions client API

#157


liayoo (approved):
LGTM, thanks!

minsulee2 (approved):
LGTM

Issue #163, opened 2020-12-15
Need to make SET operation atomic 


platfowner:
**Is your feature request related to a problem? Please describe.**
Sub-operations in SET operation can be applied if one of them fails.

**Describe the solution you'd like**
Make it atomic by using state versions.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

platfowner:
Resolved.

PR #164, opened 2020-12-15

platfowner:
Summary:
- Make SET operation atomic
- Fix temp state version conflict bug
- Make BATCH transaction results include tx hashes
- Use parseOrLog() instead of JSON.parse() to detect server-side errors easily
- Rename and move executeTransaciton() to BlockchainNode from P2PServer
- Move constants.js and chain-util.js to common/ directory 

#163


node/index.js
```diff
@@ -232,6 +232,61 @@ class BlockchainNode {
...
+  /**
+   * Try to executes a transaction on the node database. If it was not successful, all changes are
+   * rolled back from the database states.
+   * @param {Object} tx transaction
+   */
+  executeOrRollbackTransaction(tx) {
+    const backupVersion = StateManager.createRandomVersion(`${StateVersions.BACKUP}`);
+    const backupRoot = this.stateManager.cloneVersion(this.db.stateVersion, backupVersion);
+    if (!backupRoot) {
+      logger.error(`[${LOG_HEADER}] Failed to clone state version: ${this.db.stateVersion}`);
```
liayoo, line 244: should we not execute the transaction in this case?

platfowner: I think that's reasonable. Just updated.

platfowner:
@liayoo Comments addressed. PTAL

Also the following changes were added:
- Fix temp state version conflict bug
- Make BATCH transaction results include tx hashes
- Use parseOrLog() instead of JSON.parse() to detect server-side errors easily

db/state-manager.js
```diff
@@ -165,18 +165,26 @@ class StateManager {
...
       logger.error(`[${LOG_HEADER}] Non-existing version: ${version}`);
       return false;
     }
-    this.finalizedVersion = version;
+    this.finalVersion = version;
     return true;
   }
+
+  /**
+   * Returns a random state version with the given previs.
```
liayoo, line 182: minor) typo?

db/state-manager.js
```diff
@@ -165,18 +165,26 @@ class StateManager {
...
+   * @param {*} versionPrefix state version prefix
+   */
+  static createRandomVersion(versionPrefix) {
+    return `${versionPrefix}:${Date.now()}:${Math.floor(Math.random()*10000)}`;
```
liayoo, line 186: spaces between *

liayoo (approved):
LGTM with some minor comments. Thanks for the updates!

db/state-manager.js
platfowner: Fixed, thanks.

db/state-manager.js
platfowner: Fixed. Thanks!

Issue #171, opened 2020-12-25
[BUG] Blockchain node crashes without any error messages


platfowner:
**Describe the bug**
Blockchain node crashes around block number 148xx.

https://www.notion.so/comcom/Stability-debugging-38a0c95751a340038dbb5efb013dd867 

**To Reproduce**
$ sh deploy_gcp.sh summer seo 0 

**Expected behavior**

**System information:**
 - OS & version: Ubuntu 18.04
 - NodeJS: 10.x

**Screenshots**

**Additional context**

PR #172, opened 2020-12-25
Bugfix/platfowner/bugfix

platfowner:
Summary:
- Upgrade nodejs version to v12 from v10 to use --async-stack-traces option 
- Allow chain segment merging while in SERVING (returning to the original logic) 
- Add enableVersionRenaming feature flag (for further debugging) 
- Log stderr logs to error_logs.txt 
- Tweak logging messages

Background:
- To inspect #171 , tried the followings:
  1) Revert effective logical changes since release/R27
  2) Improve logging including nodejs version upgrade to use --async-stack-traces option (see https://stackoverflow.com/questions/55162619/why-do-i-lose-stack-trace-when-using-async-await-in-node-js) 
- As a result, 1) made no changes but it seems that the issue is gone by 2) (not sure, let take a longer look) 

Related memo: https://www.notion.so/comcom/Stability-debugging-38a0c95751a340038dbb5efb013dd867

#171

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -69,6 +69,7 @@ class BlockchainNode {
 
   init(isFirstNode) {
     const LOG_HEADER = 'init';
+
     logger.info(`[${LOG_HEADER}] Initializing node..`);
     const lastBlockWithoutProposal = this.bc.init(isFirstNode);
     const startingDb =
@@ -105,9 +106,10 @@ class BlockchainNode {
 
   syncDb(newVersion) {
     const LOG_HEADER = 'syncDb';
+
     const oldVersion = this.db.stateVersion;
     if (newVersion === oldVersion) {
-      logger.info(`[${LOG_HEADER}] Already sync'ed.`);
+      logger.info(`[${LOG_HEADER}] Already sync'ed with version: ${newVersion}`);
       return false;
     }
     const clonedRoot = this.stateManager.cloneFinalVersion(newVersion);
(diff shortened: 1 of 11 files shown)


node/index.js
```diff
@@ -137,9 +141,11 @@ class BlockchainNode {
     if (!this.stateManager.finalizeVersion(newFinalVersion)) {
       logger.error(`[${LOG_HEADER}] Failed to finalize version: ${newFinalVersion}`);
     }
-    logger.info(`[${LOG_HEADER}] Renaming version: ${version} -> ${newFinalVersion}`);
-    if (!this.stateManager.renameVersion(version, newFinalVersion)) {
-      logger.error(`[${LOG_HEADER}] Failed to replace version: ${version} -> ${newFinalVersion}`);
+    if (FeatureFlags.enableVersionRenaming) {
+      logger.info(`[${LOG_HEADER}] Renaming version: ${version} -> ${newFinalVersion}`);
+      if (!this.stateManager.renameVersion(version, newFinalVersion)) {
+        logger.error(`[${LOG_HEADER}] Failed to replace version: ${version} -> ${newFinalVersion}`);
```
liayoo, line 147: rename?

liayoo (approved):
LGTM with a minor comment. Thanks! I'll take a look at the issue as well.

node/index.js
platfowner: fixed. thanks.

platfowner:
Thanks for the fast review. Now merging..

PR #174, opened 2021-01-02

platfowner:
Summary:
- Fix memory leak issue by changing deleteStateTreeVersion() logic
- Do some code refactoring

#171

### DIFF
--- a/db/state-util.js
+++ b/db/state-util.js
@@ -160,9 +160,8 @@ function setStateTreeVersion(node, version) {
     node.setVersion(version);
     numAffectedNodes++;
   }
-  for (const label of node.getChildLabels()) {
-    const childNode = node.getChild(label);
-    numAffectedNodes += setStateTreeVersion(childNode, version);
+  for (const child of node.getChildNodes()) {
+    numAffectedNodes += setStateTreeVersion(child, version);
   }
 
   return numAffectedNodes;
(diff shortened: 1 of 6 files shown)


db/state-util.js
```diff
@@ -215,45 +213,30 @@ function deleteStateTree(node) {
...
+  node.resetProofHash();
+  numAffectedNodes++;
 
   for (const label of node.getChildLabels()) {
-    const childNode = node.getChild(label);
+    const child = node.getChild(label);
     node.deleteChild(label);
-    if (childNode.numParents() === 0) {
-      numAffectedNodes += deleteStateTreeVersion(childNode, version);
-    } else if (childNode.numParents() < 0) {
```
liayoo, line 238: question) do we not need to check/log this case anymore?

platfowner: We used to be using a separate count for this value so we were checking this. But now we're using a set and its 'size' for numParents() so it cannot be negative any more.

minsulee2 (approved):
LGTM

db/state-util.js
liayoo: I see, thanks for the explanation!

liayoo (approved):
LGTM, thanks!

platfowner:

PR #179, opened 2021-01-11
Feature/change timeout logic

cshcomcom:
Done
- Delete blocking timed out transaction logic (executeTransactionAndAddToPool)
- Use received_at to delete old transactions

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -234,7 +234,7 @@ class BlockchainNode {
           txList.push(createdTx);
         }
       }
-      return { tx_list: txList };
+      return {tx_list: txList};
     }
     const createdTx = this.createSingleTransaction(txBody, isNoncedTransaction);
     if (createdTx === null) {
@@ -300,11 +300,6 @@ class BlockchainNode {
     const LOG_HEADER = 'executeTransactionAndAddToPool';
 
     logger.debug(`[${LOG_HEADER}] EXECUTING TRANSACTION: ${JSON.stringify(tx, null, 2)}`);
-    if (this.tp.isTimedOutFromPool(tx.tx_body.timestamp, this.bc.lastBlockTimestamp())) {
-      return ChainUtil.logAndReturnError(
-          logger, 2, `[${LOG_HEADER}] Timeouted transaction: ${JSON.stringify(tx, null, 2)}`,
-          0);
-    }
     if (this.tp.isNotEligibleTransaction(tx)) {
       return ChainUtil.logAndReturnError(
           logger, 3,
--- a/node/index.js
+++ b/node/index.js
@@ -234,7 +234,7 @@ class BlockchainNode {
           txList.push(createdTx);
         }
       }
-      return {tx_list: txList};
+      return { tx_list: txList };
     }
     const createdTx = this.createSingleTransaction(txBody, isNoncedTransaction);
     if (createdTx === null) {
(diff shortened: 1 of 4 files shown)


tx-pool/index.js
```diff
@@ -48,6 +48,7 @@ class TransactionPool {
     if (!(tx.address in this.transactions)) {
       this.transactions[tx.address] = [];
     }
+    tx.received_at = Date.now();
```
platfowner, line 51: I think it's not a good idea to change tx itself?

platfowner: Instead, what about adding 'created_at' field here?
https://github.com/ainblockchain/ain-blockchain/blob/develop/tx-pool/transaction.js#L16

cshcomcom: Looks good to me! I work it and open this pr again

cshcomcom:
@platfowner All comments are addressed! PTAL

liayoo (approved):
LGTM, thanks! Please wait for @platfowner's review.

node/index.js
```diff
@@ -234,7 +234,7 @@ class BlockchainNode {
           txList.push(createdTx);
         }
       }
-      return { tx_list: txList };
+      return {tx_list: txList};
```
platfowner, line 237: minor: Personally I see that the previous pattern is more readable.

tx-pool/index.js
```diff
@@ -56,6 +56,7 @@ class TransactionPool {
       timestamp: tx.tx_body.timestamp,
       is_finalized: false,
       finalized_at: -1,
+      tracked_at: Date.now(),
```
platfowner, line 59: `tracked_at: tx.created_at,` ?

tx-pool/transaction.js
```diff
@@ -1,18 +1,19 @@
 const _ = require('lodash');
 const ainUtil = require('@ainblockchain/ain-util');
 const logger = require('../logger')('TRANSACTION');
-const { WriteDbOperations } = require('../common/constants');
+const {WriteDbOperations} = require('../common/constants');
```
platfowner, line 4: ditto

tx-pool/index.js
cshcomcom: Good! Done

cshcomcom:
PTAL! I deleted object-curly-spacing according to the google style guide and also changed lint. I think we have to discuss coding convention and change lint before add object-curly-spacing! @platfowner

platfowner:
@cshcomcom  what about just turning the option off until we get any agreement so that authors can choose depending their own preference?
https://eslint.org/docs/rules/object-curly-spacing#when-not-to-use-it

cshcomcom:
Good idea. I turned off option and reverted it! Done

tx-pool/transaction.js
```diff
@@ -201,7 +202,7 @@ class Transaction {
 
   static isValidFormat(txBody) {
     const sanitized = Transaction.sanitizeTxBody(txBody);
-    return _.isEqual(JSON.parse(JSON.stringify(sanitized)), txBody, { strict: true });
```
platfowner, line 204: ditto

platfowner (approved):
LGTM
w/ some minor comments.
Thanks for the upgrade!

cshcomcom:
Thanks for the review! Now merging

PR #184, opened 2021-01-22

liayoo:
As-is
- /consensus/whitelist set only once at genesis
- validators == whitelist (does not change)

To-be
- /consensus/whitelist can be modified by the root / shard owner
- consensus deposits & whitelist values obtained from previous block's state
- validators = whitelisted && staked
- Fixed proposal tx execution order in checkProposal()
- Fixed withdraw lockup checking logic
- Increased DEPOSIT_LOCKUP_DURATION_MS from 1 month to 6 months

TODOs
- Revive STAKE env var
- Replace NUM_VALIDATORS with GENESIS_WHITELIST and MIN_NUM_VALIDATORS
- Change the format of whitelist from { \<address>: \<amount> } to { \<address>: true }
- Add MIN_STAKE_AMOUNT_PER_VALIDATOR parameter
- Recover from the case where there's not enough validators (or, do not allow this case)

### DIFF
--- a/blockchain/genesis_rules.json
+++ b/blockchain/genesis_rules.json
@@ -40,6 +40,9 @@
           }
         }
       }
+    },
+    "whitelist": {
+      ".write": "lastBlockNumber < 1 || (auth === getValue('/sharding/config/sharding_path') === '/' ? '0xAAAf6f50A0304F12119D218b94bea8082642515B' : getValue('/sharding/config/shard_owner'))"
     }
   },
   "deposit": {
--- a/blockchain/genesis_rules.json
+++ b/blockchain/genesis_rules.json
@@ -42,7 +42,7 @@
       }
     },
     "whitelist": {
-      ".write": "lastBlockNumber < 1 || (auth === getValue('/sharding/config/sharding_path') === '/' ? '0xAAAf6f50A0304F12119D218b94bea8082642515B' : getValue('/sharding/config/shard_owner'))"
+      ".write": "auth === util.getOwnerAddr()"
     }
   },
   "deposit": {
--- a/common/constants.js
+++ b/common/constants.js
@@ -321,7 +321,7 @@ const TransactionStatus = {
  * Default values
  */
 const DefaultValues = {
-  DEPOSIT_LOCKUP_DURATION_MS: 15552000000 // 180 days
+  DEPOSIT_LOCKUP_DURATION_MS: moment.duration(180, 'days').as('milliseconds')
 }
 
 /**
--- a/common/constants.js
+++ b/common/constants.js
@@ -321,7 +321,7 @@ const TransactionStatus = {
  * Default values
  */
 const DefaultValues = {
-  DEPOSIT_LOCKUP_DURATION_MS: 2592000000 // 30 days
+  DEPOSIT_LOCKUP_DURATION_MS: 15552000000 // 180 days
 }
 
 /**
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -9,7 +9,7 @@ const BlockPool = require('./block-pool');
 const Transaction = require('../tx-pool/transaction');
 const PushId = require('../db/push-id');
 const ChainUtil = require('../common/chain-util');
-const StateManager = require('../db/state-manager');
+const StateNode = require('../db/state-node');
 const {
   WriteDbOperations,
   ReadDbOperations,
@@ -853,44 +853,28 @@ class Consensus {
   }
 
   getWhitelist() {
-    const lastBlockNumber = this.node.bc.lastBlockNumber();
-    const baseVersion = this.node.stateManager.getFinalVersion();
-    const snapVersion = this.node.stateManager.createUniqueVersionName(
-      `${StateVersions.SNAP}:${lastBlockNumber}`);
-    const snapDb = this.node.createTempDb(baseVersion, snapVersion, lastBlockNumber);
-    const whitelist = snapDb.getValue(ChainUtil.formatPath([
-      ConsensusDbPaths.CONSENSUS,
-      ConsensusDbPaths.WHITELIST
-    ])) || {};
-    this.node.destroyDb(snapDb);
-    return whitelist;
+    const whitelist = this.node.getValueFromStateWithVersion(this.node.stateManager.getFinalVersion(),
+        `/${ConsensusDbPaths.CONSENSUS}/${ConsensusDbPaths.WHITELIST}`, false);
+    logger.error(`[getWhitelist] whitelist: ${JSON.stringify(whitelist, null, 2)}`);
+    return whitelist || {};
   }
 
   getValidators(blockHash, blockNumber) {
     const LOG_HEADER = 'getValidators';
-    let db = this.blockPool.hashToDb.get(blockHash);
-    const isSnapDb = this.node.bc.lastBlock().hash === blockHash;
-    if (isSnapDb) {
-      const baseVersion = this.node.stateManager.getFinalVersion();
-      const snapVersion = this.node.stateManager.createUniqueVersionName(
-        `${StateVersions.SNAP}:${blockNumber}`);
-      db = this.node.createTempDb(baseVersion, snapVersion, blockNumber);
-    }
-    if (!db) {
-      const err = `[${LOG_HEADER}] No db found for block ${blockHash}, ${blockNumber}`;
+    const db = this.blockPool.hashToDb.get(blockHash);
+    const stateVersion = this.node.bc.lastBlock().hash === blockHash ?
+        this.node.stateManager.getFinalVersion() : (db ? db.stateVersion : null);
+    if (!stateVersion) {
+      const err = `[${LOG_HEADER}] No stateVersion found for block ${blockHash}, ${blockNumber}`;
       logger.error(err);
       throw Error(err);
     }
-    const whitelist = db.getValue(ChainUtil.formatPath([
-          ConsensusDbPaths.CONSENSUS,
-          ConsensusDbPaths.WHITELIST
-        ])) || {};
+    const whitelist = this.node.getValueFromStateWithVersion(stateVersion,
+        `/${ConsensusDbPaths.CONSENSUS}/${ConsensusDbPaths.WHITELIST}`, false) || {};
     const validators = {};
     Object.keys(whitelist).forEach((address) => {
-      const deposit = db.getValue(ChainUtil.formatPath([
-            PredefinedDbPaths.DEPOSIT_ACCOUNTS_CONSENSUS,
-            address
-          ]));
+      const deposit = this.node.getValueFromStateWithVersion(stateVersion,
+        `/${PredefinedDbPaths.DEPOSIT_ACCOUNTS_CONSENSUS}/${address}`, false) || {};
       if (deposit && deposit.value === whitelist[address] &&
           deposit.expire_at >= Date.now() + ConsensusConsts.DAY_MS) {
         validators[address] = deposit.value;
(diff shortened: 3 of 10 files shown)


common/constants.js
```diff
@@ -321,7 +321,7 @@ const TransactionStatus = {
  * Default values
  */
 const DefaultValues = {
-  DEPOSIT_LOCKUP_DURATION_MS: 2592000000 // 30 days
+  DEPOSIT_LOCKUP_DURATION_MS: 15552000000 // 180 days
```
minsulee2, line 324: What about using **moment.duration(180, 'days').as('milliseconds')** ?

consensus/index.js
```diff
@@ -518,49 +524,50 @@ class Consensus {
...
     this.node.destroyDb(tempDb);
     const createdTx = Transaction.create(proposalTx.tx_body, proposalTx.signature);
     if (!createdTx) {
       logger.error(`[${LOG_HEADER}] Failed to create a transaction with a proposal: ` +
           `${JSON.stringify(proposalTx, null, 2)}`);
+      this.node.destroyDb(newVersion);
       return false;
     }
     this.node.tp.addTransaction(createdTx);
-    const newVersion = this.node.stateManager.createUniqueVersionName(
```
minsulee2, line 553: Question: What makes this code lines moving up difference?

consensus/index.js
```diff
@@ -849,17 +853,69 @@ class Consensus {
...
+          ConsensusDbPaths.WHITELIST
+        ])) || {};
+    const validators = {};
+    Object.keys(whitelist).forEach((address) => {
+      const deposit = db.getValue(ChainUtil.formatPath([
+            PredefinedDbPaths.DEPOSIT_ACCOUNTS_CONSENSUS,
+            address
+          ]));
+      if (deposit && deposit.value === whitelist[address] &&
+          deposit.expire_at >= Date.now() + ConsensusConsts.DAY_MS) {
```
minsulee2, line 895: This means changing deposit values can only be applied once a day?

minsulee2 (commented):
I ran the first round :)

consensus/index.js
liayoo: Not really, it's currently implemented so that changing deposit value will update the deposit's expire_at value to current time + DEPOSIT_LOCKUP_DURATION_MS.

consensus/index.js
liayoo: Previously, it was trying to execute the proposal tx on the previous block's db state, but we're now executing the proposal tx on the proposal block's db state (after proposal block's last_votes and transactions are applied).

common/constants.js
liayoo: Done, thanks!

liayoo:
Addressed all R1 comments, PTAL @minsulee2

blockchain/genesis_rules.json
```diff
@@ -40,6 +40,9 @@
           }
         }
       }
+    },
+    "whitelist": {
+      ".write": "lastBlockNumber < 1 || (auth === getValue('/sharding/config/sharding_path') === '/' ? '0xAAAf6f50A0304F12119D218b94bea8082642515B' : getValue('/sharding/config/shard_owner'))"
```
platfowner, line 45: What about adding a rule util function that returns the owner address? e.g. getOwnerAddr()

consensus/index.js
```diff
@@ -849,17 +853,69 @@ class Consensus {
   }
 
   getWhitelist() {
-    return LIGHTWEIGHT ? GenesisWhitelist : this.node.db.getValue(ChainUtil.formatPath([
+    const lastBlockNumber = this.node.bc.lastBlockNumber();
+    const baseVersion = this.node.stateManager.getFinalVersion();
+    const snapVersion = this.node.stateManager.createUniqueVersionName(
+      `${StateVersions.SNAP}:${lastBlockNumber}`);
+    const snapDb = this.node.createTempDb(baseVersion, snapVersion, lastBlockNumber);
```
platfowner, line 860: Do we need this temp db for the read-only operations?

consensus/index.js
```diff
@@ -849,17 +853,69 @@ class Consensus {
...
+
+  getValidators(blockHash, blockNumber) {
+    const LOG_HEADER = 'getValidators';
+    let db = this.blockPool.hashToDb.get(blockHash);
+    const isSnapDb = this.node.bc.lastBlock().hash === blockHash;
+    if (isSnapDb) {
+      const baseVersion = this.node.stateManager.getFinalVersion();
+      const snapVersion = this.node.stateManager.createUniqueVersionName(
+        `${StateVersions.SNAP}:${blockNumber}`);
+      db = this.node.createTempDb(baseVersion, snapVersion, blockNumber);
```
platfowner, line 877: ditto

consensus/index.js
```diff
@@ -849,17 +853,69 @@ class Consensus {
...
+    return validators;
   }
 
   getValidConsensusDeposit(address) {
-    const deposit = this.node.db.getValue(ChainUtil.formatPath([
+    const lastBlockNumber = this.node.bc.lastBlockNumber();
+    const baseVersion = this.node.stateManager.getFinalVersion();
+    const snapVersion = this.node.stateManager.createUniqueVersionName(
+      `${StateVersions.SNAP}:${lastBlockNumber}`);
+    const snapDb = this.node.createTempDb(baseVersion, snapVersion, lastBlockNumber);
```
platfowner, line 913: ditto

consensus/index.js
minsulee2: Oh, I see. Thanks.

minsulee2 (approved):
LGTM except for @platfowner 's comments. I leave Approve tho, will see platfowner's approve as well. Thanks!

consensus/index.js
minsulee2: Check! check!

blockchain/genesis_rules.json
liayoo: Sounds great! That really simplified the rule. Thanks!

consensus/index.js
liayoo: Added getValueFromStateWithVersion() (and its counterparts for functions, rules, & owners) in node. It takes version, readingPath, and isGlobal for the parameters. I think we can use this to get states from specific versions and deprecate the getters in DB class in the future. What do you think?

node/index.js
```diff
@@ -22,6 +22,7 @@ const Transaction = require('../tx-pool/transaction');
 
 const isShardChain =
     GenesisSharding[ShardingProperties.SHARDING_PROTOCOL] !== ShardingProtocols.NONE;
+const shardingPath = ChainUtil.parsePath(GenesisSharding[ShardingProperties.SHARDING_PATH]);
```
platfowner, line 25: this.db.shardingPath can be used instead?

node/index.js
```diff
@@ -22,6 +22,7 @@ const Transaction = require('../tx-pool/transaction');
 
 const isShardChain =
```
platfowner, line 23: What about adding this as a member variable?

db/index.js
```diff
@@ -207,57 +206,38 @@ class DB {
     return this.removeEmptyNodesRecursive(fullPath, 0, this.stateRoot);
   }
 
-  readDatabase(fullPath) {
-    const stateNode = this.getRefForReading(fullPath);
-    if (stateNode === null) {
+  static getterUtil(rootNode, readingPath, isGlobal, shardingPath, rootLabel) {
```
platfowner, line 209: Thanks for the refactoring.
What about 'readFromStateRoot' or something? 
And would it be more intuitive to tweak the order or the names of the params a little bit? like:
```
static readFromStateRoot(stateRoot, rootLabel, refPath, isGlobal, shardingPath) {
...
```

Also, could you leave another layer between getValue() and this function like:
```
readDatabase(refPath, rootLabel, isGlobal) {
  return DB.readFromStateRoot(this.stateRoot, PredefinedDbPaths.VALUES_ROOT, refPath, isGlobal, this.shardingPath);
}
```

db/rule-util.js
```diff
@@ -69,6 +69,11 @@ class RuleUtil {
       return '';
     }
   }
+
+  getOwnerAddr() {
+    const { GenesisAccounts, AccountProperties } = require('../common/constants');
```
platfowner, line 74: I think this can be moved to the top?

consensus/index.js
```diff
@@ -518,49 +524,50 @@ class Consensus {
...
+    }
+
+    // Try executing the proposal tx on the proposal block's db state
+    const tempVersion = this.node.stateManager.createUniqueVersionName(
+      `${StateVersions.CONSENSUS_PROPOSE}:${prevBlock.number}:${number}`);
+    const tempDb = this.node.createTempDb(newVersion, tempVersion, prevBlock.number - 1);
     if (ChainUtil.transactionFailed(tempDb.executeTransaction(proposalTx))) {
       logger.error(`[${LOG_HEADER}] Failed to execute the proposal tx`);
+      this.node.destroyDb(tempDb);
+      this.node.destroyDb(newVersion);
```
platfowner, line 559: newDb?

consensus/index.js
```diff
@@ -518,49 +524,50 @@ class Consensus {
...
       return false;
     }
     this.node.destroyDb(tempDb);
     const createdTx = Transaction.create(proposalTx.tx_body, proposalTx.signature);
     if (!createdTx) {
       logger.error(`[${LOG_HEADER}] Failed to create a transaction with a proposal: ` +
           `${JSON.stringify(proposalTx, null, 2)}`);
+      this.node.destroyDb(newVersion);
```
platfowner, line 567: newDb?

consensus/index.js
platfowner: Using getValueFromStateWithVersion() looks good to me.
Currently, node.db is used as a default db for in-bounding use transactions to serve the temporary states before confirmation with getValue() operation. Would it be better to remove it? (Is the block proposing is totally independent of node.db now?)  I think we need to consider the simplicity of the test code as well? (We can talk more about this later)

node/index.js
```diff
@@ -168,6 +169,30 @@ class BlockchainNode {
     return this.stateManager.getFinalRoot().toJsObject(withDetails);
   }
 
+  getValueFromStateWithVersion(version, readingPath, isGlobal) {
```
platfowner, line 172: Would 'getValueWithStateVersion()' be simpler?

And waht about s/readingPath/refPath?

node/index.js
liayoo: done

node/index.js
liayoo: done

db/index.js
liayoo: Thanks for the suggestions! Done.

db/rule-util.js
liayoo: done

consensus/index.js
liayoo: Wow.. my mistake. Thanks for catching it. Done.

consensus/index.js
liayoo: done

node/index.js
liayoo: done!

liayoo:

db/functions.js
```diff
@@ -440,7 +440,7 @@ class Functions {
   }
 
   _getFullValuePath(parsedPath) {
-    return this.db.getFullPath(parsedPath, PredefinedDbPaths.VALUES_ROOT);
+    return this.db.constructor.getFullPath(parsedPath, PredefinedDbPaths.VALUES_ROOT);
```
platfowner, line 443: What about redefining getFullPath() static?

platfowner: Ah, it's already done. Then `DB.getFullPath()`?

node/index.js
```diff
@@ -168,6 +166,30 @@ class BlockchainNode {
     return this.stateManager.getFinalRoot().toJsObject(withDetails);
   }
 
+  getValueWithStateVersion(version, refPath, isGlobal) {
```
platfowner, line 169: minor: (sorry) the function name sounds like the version param comes at the end like `getValueWithStateVersion(refPath, isGlobal, version)`?

platfowner:
Thanks!

platfowner (approved):

db/functions.js
liayoo: Do you mean writing it like DB.getFullPath()? This is another way of accessing a static function, actually. I think due to some circular dependencies, requiring DB class and accessing the function like this (DB.getFullPath()) wasn't working here, and so had to resort to using the DB instance's constructor.

node/index.js
liayoo: I agree! Changed the params' order. Thanks!

liayoo:

consensus/index.js
liayoo: Sorry for my delayed reply! Somehow didn't see this comment..
First off, yes, consensus is now independent of node.db!! (yay) 
I think we can keep the node.db for executing in-bound txs on top of the final state, and I think it's time we consider implementing the "read concern" concept that we roughly designed a long time ago? Not the most pressing issue, but we can talk about this in one of our meetings :)

platfowner: Right, that'd be a good agenda for our next discussion. Thanks.

Issue #185, opened 2021-01-22
Improve load balancer's health check


platfowner:
**Is your feature request related to a problem? Please describe.**
The health check in the load balancer needs to consider both node state and consensus state.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

platfowner:
Updated load balancer's health checking with '/health_check returns true'.

PR #186, opened 2021-01-22

platfowner:
Summary:
- Make health check reflect both consensus status and node status.
- Use a separate health check path /health_check.

#185


liayoo (approved):
LGTM, thanks!

PR #189, opened 2021-01-25

### LINKED ISSUE #182: [BUG] error: [BLOCKCHAIN] First block is not the Genesis block
**Describe the bug**
When the first node is started and nobody comes, after that, terminates the first node and restart! then it faces error :(

**To Reproduce**
1. [Tracker] node tracker-server/index.js
2. [Node1] NUM_VALIDATORS=2 ACCOUNT_INDEX=0 HOSTING_ENV=local node client/index.js
3. [Node2] DO NOTHING
4. [Node1] ctrl + c (terminate the node)
5. [Node1] NUM_VALIDATORS=2 ACCOUNT_INDEX=0 HOSTING_ENV=local node client/index.js again

**Expected behavior**
The [Node1] should be wait for other nodes in a peaceful manner.

liayoo:
- Sanitize tx_body before hashing Transaction
- Sanitize transactions before hashing Block
- Fix logger message level

Fixes #182

### DIFF
--- a/unittest/blockchain.test.js
+++ b/unittest/blockchain.test.js
@@ -37,6 +37,7 @@ describe('Blockchain', () => {
     node1.addNewBlock(Block.create(
         lastBlock.hash, [], [tx], lastBlock.number + 1, lastBlock.epoch + 1, '',
         node1.account.address, {}));
+    delete tx.created_at;
     assert.deepEqual(node1.bc.chain[node1.bc.chain.length -1].transactions[0], tx);
   });
 
--- a/blockchain/block.js
+++ b/blockchain/block.js
@@ -25,11 +25,11 @@ class Block {
   constructor(lastHash, lastVotes, transactions, number, epoch, timestamp,
       stateProofHash, proposer, validators) {
     this.last_votes = lastVotes;
-    this.transactions = transactions;
+    this.transactions = Block.sanitizeTransactions(transactions);
     // Block's header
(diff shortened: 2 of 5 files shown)


unittest/blockchain.test.js
```diff
@@ -37,6 +37,7 @@ describe('Blockchain', () => {
     node1.addNewBlock(Block.create(
         lastBlock.hash, [], [tx], lastBlock.number + 1, lastBlock.epoch + 1, '',
         node1.account.address, {}));
+    delete tx.created_at;
```
platfowner, line 40: Q: Would it be useful to keep this value in the block?

liayoo: Hmm so you're saying, we should keep the information even after it's included in a block? But without including in the tx & block's hashes?

platfowner: The value was added later so the current code might not being considering it.
I'm just wondering if the creation time has value for future reference. (It might be enough to have the block generation time though.) If not, then let's just move forward.

platfowner (approved):
LGTM
Thanks!

unittest/blockchain.test.js
liayoo: Right, I think it's enough to have the tx's and block's timestamps for now. Let's see if we have a need for this after block generation!

liayoo:
Thanks for the quick review! Now merging..

Issue #188, opened 2021-01-25
Need to support multiple functions for the same ref path.


platfowner:
**Is your feature request related to a problem? Please describe.**
Currently, we support only 1 native function for the same ref path.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #190, opened 2021-01-26

platfowner:
Summary:
- Support multiple functions per path
- Use __dirname instead of process.cwd()
- Fix circular code import issue in rule-util.js

#188

### DIFF
--- a/db/functions.js
+++ b/db/functions.js
@@ -53,8 +53,8 @@ class Functions {
    * @param {Number} currentTime current time
    * @param {Object} transaction transaction
    */
-  // TODO(seo): Support multiple-functions per path.
   // TODO(seo): Trigger subtree functions.
+  // TODO(seo): Handle async function calls properly.
   triggerFunctions(parsedValuePath, value, timestamp, currentTime, transaction) {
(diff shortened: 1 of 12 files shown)


db/functions.js
```diff
@@ -52,52 +53,105 @@ class Functions {
...
+    let newFuncMap = ChainUtil.getJsObject(newFunction, [FunctionProperties.FUNCTION]);
+    if (!newFuncMap || !ChainUtil.isDict(newFunction)) {
+      // Add a place holder.
+      ChainUtil.setJsObject(newFunction, [FunctionProperties.FUNCTION], {});
+      newFuncMap = ChainUtil.getJsObject(newFunction, [FunctionProperties.FUNCTION]);
+    }
+    for (const functionKey in funcChangeMap) {
+      const functionValue = funcChangeMap[functionKey];
+      if (functionValue === null) {
+        if (functionKey in newFuncMap) {
```
liayoo, line 142: I think 'delete' behaves like this already (only removes if the key is in the object)?

liayoo (approved):

db/functions.js
platfowner: Done. Thanks!

PR #192, opened 2021-01-27

liayoo:
- Change whitelist structure from address => stake to address => true
- Revive STAKE env var
- Add some safety measures to ensure validator set size >= MIN_NUM_VALIDATORS
- Introduce MIN_STAKE_PER_VALIDATOR (its value is subject to change)
- Return false if evalRuleString() failed

TODOs
- Use a separate json file for genesis params (including genesis whitelist & validators)
- Deprecate NUM_VALIDATORS
- Remove genesis block & accounts from the source code (once block server is in place?)

### DIFF
--- a/blockchain/block.js
+++ b/blockchain/block.js
@@ -295,8 +295,7 @@ class Block {
 
     const firstTx = this.buildDbSetupTx(genesisTime, ownerPrivateKey);
     const secondTx = this.buildAccountsSetupTx(ownerAddress, genesisTime, ownerPrivateKey);
-    // NOTE: Need to get these staking txs ahead of time for genesis block (after genesis block is
-    // removed from the source code)
+    // TODO(lia): Change the logic to staking & signing by the current node
     const stakingTxs = this.buildGenesisStakingTxs(genesisTime);
 
     return [firstTx, secondTx, ...stakingTxs];
--- a/blockchain/block.js
+++ b/blockchain/block.js
@@ -10,7 +10,7 @@ const DB = require('../db');
 const {
   PredefinedDbPaths,
   GenesisAccounts,
-  GenesisWhitelist,
+  GenesisValidators,
   GenesisValues,
   GenesisFunctions,
   GenesisRules,
@@ -259,6 +259,34 @@ class Block {
     return Transaction.signTxBody(secondTxBody, privateKey);
   }
 
+  static buildGenesisStakingTxs(timestamp) {
+    const _ = require('lodash');
+    const txs = [];
+    Object.entries(GenesisValidators).forEach(([address, amount], index) => {
+      const privateKey = _.get(GenesisAccounts,
+          `${AccountProperties.OTHERS}.${index}.${AccountProperties.PRIVATE_KEY}`);
+      if (!privateKey) {
+        throw Error(`GenesisAccounts missing values: ${JSON.stringify(GenesisAccounts)}, ${address}`);
+      }
+      const txBody = {
+        nonce: -1,
+        timestamp,
+        operation: {
+          type: 'SET_VALUE',
+          ref: ChainUtil.formatPath([
+            PredefinedDbPaths.DEPOSIT_CONSENSUS,
+            address,
+            1,
+            PredefinedDbPaths.DEPOSIT_VALUE
+          ]),
+          value: amount
+        }
+      };
+      txs.push(Transaction.signTxBody(txBody, privateKey));
+    });
+    return txs;
+  }
+
   static getGenesisBlockData(genesisTime) {
     const ownerAddress = ChainUtil.getJsObject(
         GenesisAccounts, [AccountProperties.OWNER, AccountProperties.ADDRESS]);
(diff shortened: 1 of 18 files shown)


blockchain/block.js
```diff
@@ -267,8 +295,11 @@ class Block {
 
     const firstTx = this.buildDbSetupTx(genesisTime, ownerPrivateKey);
     const secondTx = this.buildAccountsSetupTx(ownerAddress, genesisTime, ownerPrivateKey);
+    // NOTE: Need to get these staking txs ahead of time for genesis block (after genesis block is
+    // removed from the source code)
+    const stakingTxs = this.buildGenesisStakingTxs(genesisTime);
```
minsulee2, line 300: is it gonna wait for all deposit txs when getting started?

minsulee2: I got this. Never mind the first comment!

consensus/index.js
```diff
@@ -76,29 +75,16 @@ class Consensus {
     this.genesisHash = genesisBlock.hash;
     const myAddr = this.node.account.address;
     try {
+      const targetStake = process.env.STAKE ? Number(process.env.STAKE) : 0;
```
minsulee2, line 78: The name seems ambiguous tho, I cannot come up with way better idea :(

consensus/index.js
```diff
@@ -312,9 +298,25 @@ class Consensus {
 
     const myAddr = this.node.account.address;
     // Need the block#1 to be finalized to have the deposits reflected in the state
-    const validators = this.node.bc.lastBlockNumber() < 1 ?
-        lastBlock.validators : this.getValidators(lastBlock.hash, lastBlock.number);
-    if (!validators || !(Object.keys(validators).length)) throw Error('No whitelisted validators')
+    let validators = {};
```
minsulee2, line 301: Can it be const init?

minsulee2 (approved):
LGTM with minor comments!!!
Thanks for the quick PR :)

db/index.js
```diff
@@ -678,14 +678,21 @@ class DB {
...
         this.addPathToValue(value, matched.matchedValuePath, matched.closestRule.path.length);
     const newData =
         this.addPathToValue(newValue, matched.matchedValuePath, matched.closestRule.path.length);
-    return !!this.evalRuleString(
+    let evalRuleRes = false;
+    try {
+      evalRuleRes = !!this.evalRuleString(
         matched.closestRule.config, matched.pathVars, data, newData, address, timestamp);
+    } catch (e) {
+      logger.error(`[${LOG_HEADER}] Failed to eval rule: ${e}`);
```
platfowner, line 693: What about adding the param values of evalRuleString() to the message?

consensus/index.js
liayoo: Done. Thanks!

consensus/index.js
liayoo: I couldn't come up with one either :( I'll try to update later if I think of a better name!

db/index.js

liayoo:

blockchain/block.js
```diff
@@ -267,8 +295,11 @@ class Block {
 
     const firstTx = this.buildDbSetupTx(genesisTime, ownerPrivateKey);
     const secondTx = this.buildAccountsSetupTx(ownerAddress, genesisTime, ownerPrivateKey);
+    // NOTE: Need to get these staking txs ahead of time for genesis block (after genesis block is
```
platfowner, line 298: Question: What are the benefits or necessity of staking for all blockchain nodes in the genesis time?

platfowner: I mean could you share more context?

consensus/index.js
```diff
@@ -312,9 +298,25 @@ class Consensus {
 
     const myAddr = this.node.account.address;
     // Need the block#1 to be finalized to have the deposits reflected in the state
-    const validators = this.node.bc.lastBlockNumber() < 1 ?
-        lastBlock.validators : this.getValidators(lastBlock.hash, lastBlock.number);
-    if (!validators || !(Object.keys(validators).length)) throw Error('No whitelisted validators')
+    const validators = {};
+    if (this.node.bc.lastBlockNumber() < 1) {
+      const whitelist = GenesisWhitelist;
+      Object.keys(whitelist).forEach((address) => {
```
platfowner, line 304: minor: `for (const address in whitelist) { ...`?

blockchain/block.js
liayoo: Sorry for lacking sufficient context for the changes.
Since we consider a node to be a validator if it's whitelisted && is staking, I thought that it would really make more sense to start with the genesis validators staking some of their AIN at the beginning. I was thinking when we take out the genesis block and accounts from the code, we would have to create the txs ahead of time and have a fixed block. 
Also, there was an issue in broadcasting the stake txs - the first node's tx seemed to not propagate properly to the non-first nodes. I think it might be fixed once the p2p changes are applied though. 
I made the choice to just include the txs in the genesis block since it was impeding others' development, but if we think that the genesis block should be variable, the staking txs can be reverted easily & I can try to fix the propagation issue once the network PR is in!

consensus/index.js
liayoo: Done. Thanks!

liayoo:

blockchain/block.js
platfowner: Thanks for the explanation.
It seems to be fine to me to stake in the genesis block, what looks a little bit unnatural to me though is signing the staking transactions using each blockchain node's private key. I think it could be more natural and convenient if it can be done using the owner's private key instead as we can assume that the first node is run by the owner. What do you think? (e.g. extending the deposit function to allow depositing for the third account?) 
I think just leaving some todos or notes would be enough for now though.

platfowner (approved):
LGTM with some questions and suggestions.
Thanks!

blockchain/block.js
liayoo: Right, I do like the idea. We're not using the owner's account for a node now, but if we do in the future (probably when we remove genesis block & other things), that would make sense! 
The deposit / withdraw have a lot to be done.. Extending the deposit also sounds good.
Anyways, added a todo. Thanks!

liayoo:

PR #193, opened 2021-01-28
Feature/network

minsulee2:
Updates:
- Separate p2p server and p2p client
- Refactor p2p server
- Refactor tracker server
- Fix some minor bugs

TODOs:
- Separate server and client more precisely
- Separate execute and broadcast functions
- Do not send msg from consensus
- Heartbeat
- sign and verify


client/index.js
```diff
@@ -558,3 +566,22 @@ function validateVersion(req, res, next) {
     next();
   }
 }
+
+function matchConnections(numConnection, numOutbound, numInbound) {
+  let maxConnection = numConnection;
+  let maxOutbound = numOutbound;
+  let maxInbound = numInbound;
+  if (0 > numConnection || numConnection < 5) {
```
platfowner, line 574: Math.max(numConnection, 5)?

client/index.js
```diff
@@ -558,3 +566,22 @@ function validateVersion(req, res, next) {
...
+    maxConnection = 5;
+  }
+  if (0 > numOutbound || numOutbound < 2) {
```
platfowner, line 577: Math.max(numOutbound, 2)?

client/index.js
```diff
@@ -558,3 +566,22 @@ function validateVersion(req, res, next) {
...
+    maxOutbound = 2;
+  }
+  if (maxConnection < maxOutbound) {
```
platfowner, line 580: Math.min(maxOutbound, maxConnection)?

client/index.js
```diff
@@ -558,3 +566,22 @@ function validateVersion(req, res, next) {
...
+    maxOutbound = maxConnection;
+  }
+  if (0 > numInbound || numInbound < numConnection - numOutbound) {
```
platfowner, line 583: `const maxInbound = Math.max(numInbound, numConnection - numOutbound)`?

common/constants.js
```diff
@@ -53,6 +53,8 @@ const MessageTypes = {
   CHAIN_SEGMENT: 'CHAIN_SEGMENT',
   CHAIN_SEGMENT_REQUEST: 'CHAIN_SEGMENT_REQUEST',
   CONSENSUS: 'CONSENSUS',
+  ACCOUNT: 'ACCOUNT',
```
platfowner, line 56: minor: what about adding '_RESPONSE' suffix to 'ACCOUNT' and 'CHAIN_SEGMENT' for readability?

consensus/constants.js
```diff
@@ -3,11 +3,15 @@ const ConsensusConsts = {
   DAY_MS: 86400000,
   EPOCH_MS: 3000,
   MAX_CONSENSUS_STATE_DB: 1000,
+  HEALTH_THRESHOLD_EPOCH: 600, // 600 epochs = 10 minutes
   GENESIS_STAKE: 250,
   MIN_STAKE_PER_VALIDATOR: 100,
   // TODO(lia): Use a separate genesis json file for consensus params and deprecate NUM_VALIDATORS
   MIN_NUM_VALIDATORS: process.env.NUM_VALIDATORS ? Number(process.env.NUM_VALIDATORS) : 3,
-  HEALTH_THRESHOLD_EPOCH: 600 // 600 epochs = 10 minutes
+  HEALTH_THRESHOLD_EPOCH: 600, // 600 epochs = 10 minutes
```
platfowner, line 11: dup definition?

consensus/constants.js
```diff
@@ -3,11 +3,15 @@ const ConsensusConsts = {
...
+  INITIAL_MAX_CONNECTION: process.env.MAX_CONNECTION ? Number(process.env.MAX_CONNECTION) : 5,
```
platfowner, line 12: Unless these values are related to consensus, what about moving out of this enum to somewhere in common/constants.h?

client/index.js
```diff
@@ -558,3 +566,22 @@ function validateVersion(req, res, next) {
     next();
   }
 }
+
+function matchConnections(numConnection, numOutbound, numInbound) {
```
platfowner, line 570: What about moving this under server/index.js as a static function?

server/index.js
```diff
@@ -792,6 +583,392 @@ class P2pServer {
     await sendTxAndWaitForFinalization(parentChainEndpoint, shardInitTx, ownerPrivateKey);
     logger.info(`setUpDbForSharding success`);
   }
+}
+
+class P2pClient {
```
platfowner, line 588: What about moving this into a separate file?

server/index.js
```diff
@@ -792,6 +583,392 @@ class P2pServer {
...
+  }
+
+  buildManagedPeersInfo() {
+    return {
+      outbound: this.outbound,
+      inbound: this.server.inbound
+    };
+  }
+
+  // TODO(seo): Add sharding status.
```
platfowner, line 626: Could you remove this for me? It's already done.

tracker-server/index.js
```diff
@@ -109,221 +77,131 @@ const server = new WebSocketServer({
...
   }
+}
 
-  addPeer(peer) {
-    if (peer && peer.address !== this.address && !this.managedPeers[peer.address] &&
-        !this.unmanagedPeers[peer.address]) {
-      this.managedPeers[peer.address] = true;
-      peer.unmanagedPeers[this.address] = true;
+function getPeerCandidates(myself, candidates) {
+  Object.keys(PEER_NODES).forEach(address => {
```
platfowner, line 156: minor: `for (const address in PEER_NODES) {`?

tracker-server/index.js
```diff
@@ -109,221 +77,131 @@ const server = new WebSocketServer({
...
-      logger.info(`>> Message to node [${abbrAddr(node.address)}]: ` +
-          `${JSON.stringify(msgToNode, null, 2)}`)
-      ws.send(JSON.stringify(msgToNode));
-      printNodesInfo();
-    } catch (error) {
-      logger.error(error.stack);
+    const nodeInfo = JSON.parse(message);
+    WS_LIST[ws.uuid] = nodeInfo.address;
+    if (PEER_NODES[nodeInfo.address]) {
+      PEER_NODES[nodeInfo.address] = nodeInfo;
```
platfowner, line 92: can be pulled out of if-else block?

platfowner: ah, location is set. please ignore the comment above.

consensus/index.js
```diff
@@ -216,12 +216,13 @@ class Consensus {
         // prevent the node from getting/handling messages properly.
         // this.node.status = BlockchainNodeStatus.SYNCING;
 
-        this.server.requestChainSegment(this.node.bc.lastBlock());
+        // TODO(minsu): FIXME!!
```
liayoo, line 219: Could you be a bit more specific?

tracker-server/index.js
```diff
@@ -4,89 +4,57 @@ const geoip = require('geoip-lite');
 const express = require('express');
 const jayson = require('jayson');
 const _ = require('lodash');
+const { v4: uuidv4 } = require('uuid');
 const logger = require('../logger')('TRACKER_SERVER');
 
 const P2P_PORT = process.env.P2P_PORT || 5000;
 const PORT = process.env.PORT || 8080;
-const MAX_NUM_PEERS = 2;
 const PEER_NODES = {};
```
liayoo, line 12: suggestion) How about using camelCase? Currently, I think it implies that it's 'constant'. Same goes for WS_LIST.

server/index.js
```diff
@@ -556,87 +399,35 @@ class P2pServer {
...
-  broadcastConsensusMessage(msg) {
-    logger.debug(`SENDING: ${JSON.stringify(msg)}`);
-    this.sockets.forEach((socket) => {
-      socket.send(JSON.stringify({
-        type: MessageTypes.CONSENSUS,
-        message: msg,
-        protoVer: CURRENT_PROTOCOL_VERSION
-      }));
-    });
+  removeFromListIfExists(address) {
```
liayoo, line 423: suggestion) removeFromListIfExists -> removeFromInboundListIfExists, removeFromOutboundListIfExists

server/index.js
```diff
@@ -52,22 +51,21 @@ const HEARTBEAT_INTERVAL_MS = 1000;
...
-// TODO(seo): Sign messages to tracker or peer.
+// TODO(minsu): Sign messages to tracker or peer.
 class P2pServer {
-  constructor(node, minProtocolVersion, maxProtocolVersion) {
-    this.internalIpAddress = null;
-    this.externalIpAddress = null;
-    this.trackerWebSocket = null;
+  constructor (node, minProtocolVersion, maxProtocolVersion, maxConnection, maxOutbound, maxInbound)
+  {
     this.server = null;
```
platfowner, line 58: As this is a web socket server, what about giving more specific name like this.wsServer? Just this.server sounds like it's a counterpart of this.client?

client/index.js
platfowner: And if the variables from this function are not used except initializing p2pServer, would it be better to hide this inside the p2pServer logic?

platfowner:
Thanks for the hard work, I have several basic-level questions:
- I see that sendChainSegment() is a part of the client logic.
  What I assumed was that server handles requests from other peers while client sends requests to peers (also handling the responses from the servers).
  If it's correct, should this function a part of the server class? 
  
  Maybe it's caused by trying to separate code for the client and the server. Some functions (e.g. broadcastTransaction()) may lay in an overlapping area? 
  
- For the separation of the server part and the client code, I'm not quite sure about its benefits. It seems that we see clear benefits from web socket channel separation for the client and the server, what would be the benefits of having separate code (i.e., class) for the client and the server?

server/index.js
platfowner: Until the refactoring is on a finalization stage, I think it's just fine to have a todo here.

platfowner:
Discussed offline with the following agreements:
- The server handles requests (CHAIN_SEGMENT_REQUEST, ACCOUNT_REQUEST), while the client sends requests and broad cast messages (TRANSACTION, CONSENSUS)
- Let have the server logic in P2PServer and the client logic in P2PClient
- For further refactoring, let's discuss in a meeting with @liayoo

client/index.js
minsulee2: Done!

client/index.js
minsulee2: Done!

client/index.js
minsulee2: Done!

client/index.js
minsulee2: Done!

common/constants.js
minsulee2: Yeah, this would be better. applied em!

consensus/constants.js
minsulee2: That was , change.

consensus/constants.js
minsulee2: It moved thanks!

client/index.js
minsulee2: It moved to p2p server! thanks!

server/index.js
minsulee2: Separted!

server/index.js
minsulee2: Done!

tracker-server/index.js
minsulee2: Done!

consensus/index.js
minsulee2: Added some comments on it!

tracker-server/index.js
minsulee2: Updated!

server/index.js
minsulee2: Updated!

server/index.js
minsulee2: Sounds gooooood. Updated!! it is now way clearer!

minsulee2:
As discussed off-line, the further refactoring especially connecting to consensus parts will be discussed together and fixed in a next PR.
PTAL @platfowner @liayoo

consensus/constants.js
```diff
@@ -3,6 +3,7 @@ const ConsensusConsts = {
   DAY_MS: 86400000,
   EPOCH_MS: 3000,
   MAX_CONSENSUS_STATE_DB: 1000,
+  HEALTH_THRESHOLD_EPOCH: 600, // 600 epochs = 10 minutes
```
platfowner, line 6: Is this used somewhere?

platfowner: or dup definition?

server/client.js
```diff
@@ -0,0 +1,427 @@
...
+      this.setTrackerEventHandlers();
+      this.setIntervalForTrackerUpdate();
+    });
+    this.trackerWebSocket.on('error', (error) => {
+      logger.error(`Error in communication with tracker (${TRACKER_WS_ADDR}): ` +
+        `${JSON.stringify(error, null, 2)}`);
+    });
+  }
+
+  sendChainSegment(socket, chainSegment, number, catchUpInfo) {
```
platfowner, line 142: As this is called only from the server-side, why don't you move this under P2PServer class?

server/index.js
```diff
@@ -793,20 +586,14 @@ class P2pServer {
...
-  }
-
-  clearIntervalHeartbeat(address) {
-    clearInterval(this.managedPeersInfo[address].intervalHeartbeat);
-    this.managedPeersInfo[address].intervalHeartbeat = null;
+  // NOTE(minsu): the total number of connection is up to more than 5 without limit.
+  // maxOutbound is for now limited equal or less than 2.
+  // maxInbound is a rest of connection after maxOutbound is set.
+  static matchConnections(numConnection, numOutbound, numInbound) {
+    const maxConnection = Math.max(numConnection, 5);
```
platfowner, line 593: What about defining 5 and 2 as constants?

tracker-server/index.js
```diff
@@ -109,221 +77,130 @@ const server = new WebSocketServer({
...
+    printNodesInfo();
   });
 
+  // TODO(minsu): code should be setup ex) code === 1006: SIGINT
   ws.on('close', (code) => {
-    logger.info(`\nDisconnected from node [${node ? abbrAddr(node.address) : 'unknown'}] ` +
+    const address = wsList[ws.uuid];
+    logger.info(`\nDisconnected from node [${address ? abbrAddr(address) : 'unknown'}] ` +
         `with code: ${code}`);
-    PEER_NODES[node.address].isLive = false;
```
platfowner, line 149: We used to be using PEER_NODES to track node status in a non-volatile way, i.e., even when some of them die /peer_nodes API still returns their last status (e.g. memory status, disk status, etc) for debugging purpose. In such circumstance, isLive flag was an indicator for each node's liveness. 
If we just reset the status data when some nodes are disconnected, their status will be lost.

consensus/constants.js
minsulee2: Oh, my apology! It was dup definition due to conflicts.

server/client.js
minsulee2: Right! I was so confusing! Fixed!

server/index.js
minsulee2: Done!

tracker-server/index.js
minsulee2: I'd thought it was next up with heartbeat. But, it was necessary! I've revived the isAlive status and left TODO for isAlive as well.

minsulee2:
@platfowner PTAL. All addressed!

client/index.js
```diff
@@ -39,15 +40,22 @@ if (!semver.valid(CURRENT_PROTOCOL_VERSION)) {
...
+} = P2pServer.matchConnections(PeerConnections.INITIAL_MAX_CONNECTION,
+    PeerConnections.INITIAL_MAX_OUTBOUND, PeerConnections.INITIAL_MAX_INBOUND);
 
 const app = express();
 app.use(express.json()); // support json encoded bodies
 
 const node = new BlockchainNode();
-const p2pServer = new P2pServer(node, minProtocolVersion, maxProtocolVersion);
+const p2pServer = new P2pServer(node, minProtocolVersion, maxProtocolVersion,
+    maxConnection, maxOutbound, maxInbound);
```
platfowner, line 58: minor: Unless we have a plan to alter these values for some cases (e.g. testing) in near future, it seems that these values can just be set implicitly in the constructor of P2pServer?

common/constants.js
```diff
@@ -351,7 +354,18 @@ const FeatureFlags = {
   enableStateVersionOpt: true,
   // Enables state version renaming.
   enableVersionRenaming: true,
-}
+};
+
+/**
+ * Connection info.
+ */
+const PeerConnections = {
```
platfowner, line 362: what about defining only initial and default values here and handle the actual values outside of this enum definition? e.g.,

const PeerConnections = {
  DEFAULT_MAX_CONNECTION: 5,
  ...
  INITIAL_MAX_CONNECTION: 5,
  ...
}

  static matchConnections() {
    const numConnection = process.env.MAX_CONNECTION ? Number(process.env.MAX_CONNECTION) : PeerConnections.INITIAL_MAX_CONNECTION;
    ...
    const maxConnection = Math.max(numConnection, PeerConnections.DEFAULT_MAX_CONNECTION);
    ...
    return { maxConnection, maxOutbound, maxInbound };
  }

server/index.js
```diff
@@ -793,20 +597,14 @@ class P2pServer {
...
-
-  clearIntervalHeartbeat(address) {
-    clearInterval(this.managedPeersInfo[address].intervalHeartbeat);
-    this.managedPeersInfo[address].intervalHeartbeat = null;
+  // NOTE(minsu): the total number of connection is up to more than 5 without limit.
+  // maxOutbound is for now limited equal or less than 2.
+  // maxInbound is a rest of connection after maxOutbound is set.
+  static matchConnections(numConnection, numOutbound, numInbound) {
+    const maxConnection = Math.max(numConnection, PeerConnections.DEFAULT_MAX_CONNECTION);
+    const maxOutbound = Math.min(numOutbound, PeerConnections.DEFAULT_MAX_OUTBOUND);
```
platfowner, line 605: It seems that DEFAULT_MAX_CONNECTION and DEFAULT_MAX_OUTBOUND values are used as some limit values. If it's correct, what about naming so? e.g. MAX_CONNECTION_LIMIT, MAX_OUTBOUND_LIMIT

server/index.js
```diff
@@ -793,20 +597,14 @@ class P2pServer {
...
-        ws.ping();
-      });
-    }, HEARTBEAT_INTERVAL_MS);
-  }
-
-  clearIntervalHeartbeat(address) {
-    clearInterval(this.managedPeersInfo[address].intervalHeartbeat);
-    this.managedPeersInfo[address].intervalHeartbeat = null;
+  // NOTE(minsu): the total number of connection is up to more than 5 without limit.
+  // maxOutbound is for now limited equal or less than 2.
```
platfowner, line 601: It seems that the DEFAULT_MAX_OUTBOUND is an upper bound? If so, `Math.max(numOutbound, PeerConnections.DEFAULT_MAX_OUTBOUND)`?

tracker-server/index.js
platfowner: Thanks!

common/constants.js
minsulee2: Updated!

server/index.js
minsulee2: Updated!

server/index.js
minsulee2: Updated!

client/index.js
minsulee2: Oh, I'd missed this part before. I thought providing connection status is good idea thru REST api. What do you think?

minsulee2:
PTAL! @platfowner I feel almost there!

client/index.js
platfowner: Do you want to expose the return values of matchConnections() or the current P2PServer's member variables? e.g.
```
  constructor (node, minProtocolVersion, maxProtocolVersion, maxConnection, maxOutbound, maxInbound)
  {
    ...
    this.maxConnection = maxConnection;
    this.maxOutbound = maxOutbound;
    this.maxInbound = maxInbound;
``` 

(Maybe the values don't change after initial setting) 
If it's the latter case, a new member function like getConnectionStatus() can be added for the purpose, which returns the three values plus some connection related numbers?

minsulee2: Done!

minsulee2: PTAL :)

server/client.js
```diff
@@ -0,0 +1,417 @@
...
+    }, RECONNECT_INTERVAL_MS);
+  }
+
+  clearIntervalForTrackerConnection() {
+    clearInterval(this.intervalConnection);
+    this.intervalConnection = null;
+  }
+
+  // TODO(minsu): maxConnection in client/index.js, maxOutbound in p2pClient, maxInbound in p2pServer
+  getConnectionInfo() {
```
platfowner, line 44: Can be replaced with server.getConnectionInfo()?

platfowner (approved):
Thanks!

server/client.js
minsulee2: Done!

minsulee2:
Thanks for getting through the long journey together! I am merging it now.

PR #201, opened 2021-02-05
Get location always

minsulee2:
Fixed that location set only once.


platfowner (approved):
LGTM
Thanks!

minsulee2:
Super thanks for the quick review!

Issue #200, opened 2021-02-04
Need to cleanup some unused code


platfowner:
e.g. isNoncedTransaction, legacy configs in genesis rules file

platfowner:
Done.

PR #214, opened 2021-02-10
Fix legacy issues 

platfowner:
Summary:
- Move legacy /afan path to /apps/afan
- Deprecate isNoncedTransaction
- Add additional values of transaction to 'extra' field
- Move default tx timestamp & nonce setting to blockchain node from client API

#200


liayoo (approved):
LGTM, thanks!

platfowner:
Thanks for the fast review!

Issue #216, opened 2021-02-11
Need to expose blockchain node status for stability testing/monitoring


platfowner:
e.g. /node_status

platfowner:
Done.

PR #222, opened 2021-02-15
Expose /status client API for blockchain node monitoring

platfowner:
Summary:
- Expose /status client API
- Tweak status structure and names
- Fix some broken tests

#216

### DIFF
--- a/p2p/index.js
+++ b/p2p/index.js
@@ -79,13 +79,6 @@ class P2pClient {
     this.intervalConnection = null;
   }
 
-  getManagedPeersInfo() {
-    return {
-      outbound: ChainUtil.simplifyProperties(this.outbound),
-      inbound: ChainUtil.simplifyProperties(this.server.inbound),
-    };
-  }
-
   getProtocolInfo() {
     return {
       versionMap: PROTOCOL_VERSION_MAP,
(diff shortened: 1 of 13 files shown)


p2p/index.js
```diff
@@ -77,35 +79,74 @@ class P2pClient {
     this.intervalConnection = null;
   }
 
-  buildManagedPeersInfo() {
+  getManagedPeersInfo() {
```
liayoo, line 82: Just noticed, but is it still correct to say managed? How about getPeersInfo?

liayoo (approved):
LGTM with a comment. Thanks!

p2p/index.js
platfowner: Just changed to use networkStatus.connectionInfo instead (they're dup info). Thanks.

p2p/index.js
```diff
@@ -77,35 +79,66 @@ class P2pClient {
...
+  }
+
+  getStatus() {
+    const blockStatus = this.server.getBlockStatus();
+    return {
+      address: this.server.getNodeAddress(),
+      updatedAt: Date.now(),
+      lastBlockNumber: blockStatus.number,
+      networkStatus: this.getNetworkStatus(),
+      blockStatus: this.server.getBlockStatus(),
```
cshcomcom, line 128: minor: blockStatus?

platfowner:
All tests passed. Now merging..

minsulee2:
I was a bit late because of other meetings! LGTM!

p2p/index.js
platfowner: Right. will be fixed in next CLs. Thanks.

PR #223, opened 2021-02-16
Bugfix/restart

minsulee2:
Make some fixes handling smooth restart especially for the case of early block number.

### DIFF
--- a/blockchain/index.js
+++ b/blockchain/index.js
@@ -5,9 +5,9 @@ const glob = require('glob');
 const zipper = require('zip-local');
 const naturalSort = require('node-natural-sort');
 const logger = require('../logger')('BLOCKCHAIN');
-const {Block} = require('./block');
+const { Block } = require('./block');
 const BlockFilePatterns = require('./block-file-patterns');
-const {BLOCKCHAINS_DIR} = require('../common/constants');
+const { BLOCKCHAINS_DIR } = require('../common/constants');
 const CHAIN_SEGMENT_LENGTH = 20;
 const ON_MEM_CHAIN_LENGTH = 20;
 
@@ -35,7 +35,6 @@ class Blockchain {
         logger.info('## Starting NON-FIRST-NODE blockchain with EMPTY blocks... ##');
         logger.info('#############################################################');
         logger.info('\n');
-        this.chain = [];
         this.writeChain();
       }
     } else {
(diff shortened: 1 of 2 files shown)


blockchain/index.js
```diff
@@ -196,6 +198,11 @@ class Blockchain {
       if (!(fs.existsSync(directory))) {
```
platfowner, line 198: minor: () is looks redundant.

blockchain/index.js
```diff
@@ -196,6 +198,11 @@ class Blockchain {
...
         fs.mkdirSync(directory);
         created = true;
+      } else {
+        const files = fs.readdirSync(directory);
+        if (!files.length) {
```
platfowner, line 203: could you add some comments why it's deemed to be 'created'?

db/functions.js
```diff
@@ -82,7 +82,7 @@ class Functions {
         if (functionEntry.function_type === FunctionTypes.NATIVE) {
           const nativeFunction = this.nativeFunctionMap[functionEntry.function_id];
           if (nativeFunction) {
-            logger.info(
+            logger.debug(
```
platfowner, line 85: could you revert this? native function calls are often very useful for identifying the current server status.

blockchain/index.js
```diff
@@ -35,7 +35,6 @@ class Blockchain {
         logger.info('## Starting NON-FIRST-NODE blockchain with EMPTY blocks... ##');
         logger.info('#############################################################');
         logger.info('\n');
-        this.chain = [];
```
liayoo, line 38: Is it required that `this.chain` is not an empty array? If so, how about initializing it to `null`?

minsulee2: this.chain is init with `this.chain = [];` in the constructor and this.chain is only treated as array all around the blockchain/index.js. So that, I finally updated the above `this.chain.push(Block.genesis());` What do you think?

blockchain/index.js
minsulee2: Added!

db/functions.js
minsulee2: Oh, my apology. I should've rollback after debugging. Fixed!

blockchain/index.js
minsulee2: Done!

minsulee2:
All addressed PTAL :)

blockchain/index.js
platfowner: Thanks!

platfowner (approved):
LGTM
Thanks!

blockchain/index.js
liayoo: Sounds good to me! Actually, I was confused in my first comment & thought this change was in the constructor 🤦

liayoo (approved):
LGTM, thanks!

minsulee2:
Thanks for the reviews! merging now!

Issue #213, opened 2021-02-10
[BUG] Txs with strictly ordered nonces are dropped


platfowner:
**To Reproduce**
1.  Remove `nonce: -1` line in `setUp()` of sharding.test.js
2. Run ./node_modules/mocha/bin/mocha --timeout 320000 integration/sharding.test.js -g 'set_value with is_global = false'

It shows:
```
Iteration count exceeded its limit before the given tx 0x064be9ad56c513d348e30664ebd151ac0fb29eb3302775322487d944cbc2e3c1 is finalized!
Failed to check finalization of cleanUp() tx.
```

This is due to the transaction sent by `setUp()` is dropped with mismatched nonce.

Also, the following test run shows no such error message:
```
./node_modules/mocha/bin/mocha --timeout 320000 integration/node.test.js -g 'set_value'
```

This is because of another bug that when there is only one element in `orderedUnvalidatedTransactions` variable in `getValidTransactions()` of tx-pool/index.js, the nonce checking is skipped:
```
while (orderedUnvalidatedTransactions.length > 1) {
```

liayoo:
Fixed in #281

Issue #160, opened 2020-12-14
[BUG] Block is not extending one of the longest notarized chains


platfowner:
I don't know why yet but it was not detected while developing and testing new features for a while since `yarn run test_unit` tells nothing about this. I shows the error messages of the failure only with individual test run command:
```
./node_modules/mocha/bin/mocha --timeout 160000 unittest/block-pool.test.js
```

@liayoo , could you take a look (seems not very critical? not sure). I had a quick look but due to the complexity of the code, it wasn't easy to catch something.

```
  2 passing (569ms)
  2 failing

  1) BlockPool
       Returns a finalizable chain:
     AssertionError: expected undefined to equal true
      at Context.it (unittest/block-pool.test.js:128:65)

  2) BlockPool
       Correctly returns the longest notarized chain:

      AssertionError: expected [ Array(1) ] to deeply equal [ Array(1) ]
      + expected - actual

               "0x03AAb7b6f16A92A1dfe018Fe34ee420eb098B98A": 250
               "0x04A456C92A880cd59D7145C457475515a6f6E0f2": 250
             }
           }
      +    {
      +      "epoch": 1
      +      "hash": "0x8997cc89f6e696e2a11fe4801c6908bf7572eb505f803309da2b75b8d3d89795"
      +      "last_hash": "0xdc5e82aca85f691dbe43403379b79c167d9d9acfd62349e69cb8621497868ac4"
      +      "last_votes": []
      +      "last_votes_hash": "0x853fb99c831d4952ff90b897bd7d7c5c2f3747e8eda8ad13e7359b731eadc299"
      +      "number": 1
      +      "proposer": "0x00ADEc28B6a845a085e03591bE7550dd68673C1C"
      +      "size": 0
      +      "stateProofHash": ""
      +      "timestamp": 1607917657536
      +      "transactions": []
      +      "transactions_hash": "0x853fb99c831d4952ff90b897bd7d7c5c2f3747e8eda8ad13e7359b731eadc299"
      +      "validators": {
      +        "0x00ADEc28B6a845a085e03591bE7550dd68673C1C": 250
      +      }
      +    }
      +    {
      +      "epoch": 3
      +      "hash": "0x0b2b8a375d804a60db36f6aff9f6f3c2a3f490d70337e879eb6757bce2aef0a9"
      +      "last_hash": "0x8997cc89f6e696e2a11fe4801c6908bf7572eb505f803309da2b75b8d3d89795"
      +      "last_votes": []
[...]

PR #225, opened 2021-02-16
Bugfix/consensus

liayoo:
- Added { nonce: -1 } to consensus related transactions #213 
- Changed logging level from error to info ("Block is not extending one of the longest notarized chains") #160


platfowner (approved):
LGTM
Thanks!

PR #230, opened 2021-02-17
Set full graph of network

minsulee2:
After this PR, full network connection is set by default. The `NON_COMPLETE_NETWORK` env will handle network connections afterwards.


p2p/index.js
```diff
@@ -55,6 +55,7 @@ class P2pClient {
     this.maxConnection = Math.max(numConnection, MAX_CONNECTION_LIMIT);
     this.maxOutbound = Math.min(numOutbound, MAX_OUTBOUND_LIMIT);
     this.maxInbound = Math.min(numInbound, numConnection - numOutbound);
+    console.log(numConnection, numOutbound, numInbound)
```
liayoo, line 58: remove? or change to logger.debug()?

common/constants.js
```diff
@@ -406,6 +406,24 @@ function overwriteGenesisParams(overwritingParams, type) {
...
 overwriteGenesisParams(OVERWRITING_CONSENSUS_PARAMS, 'consensus');
 
+// Note(minsu): If NON_COMPLETE_NETWORK env is set, it tightly limits the outbound connections.
+// Full network connections are convinced (especially for dev mode) otherwise.
+function initializeNetworkEnvronments() {
+  if (process.env.NON_COMPLETE_NETWORK) {
+    return GenesisParams.network;
+  } else {
+    return {
+      MAX_CONNECTION_LIMIT: 5,
```
liayoo, line 416: How about we use MIN_NUM_VALIDATORS here and below?

platfowner, line 416: What about using MAX_INBOUND_LIMIT instead of MAX_CONNECTION_LIMIT?
And managing maxInBound and maxOutBound separately (i.e., not using INITIAL_MAX_CONNECTION)?
It seems this could make the logic simpler. Do we have other particular use cases of connection limits?

common/constants.js
```diff
@@ -406,6 +406,24 @@ function overwriteGenesisParams(overwritingParams, type) {
...
+      MAX_OUTBOUND_LIMIT: 5,
+      INITIAL_MAX_CONNECTION: 10,
+      INITIAL_MAX_OUTBOUND: 5,
```
platfowner, line 419: Currently the values depend on the values in genesis_parms.json particularly MIN_NUM_VALIDATORS.
What about setting 
{
  MAX_BOUND_LIMIT: MIN_NUM_VALIDATORS,
  ...
} 
?

common/constants.js
```diff
@@ -406,6 +406,24 @@ function overwriteGenesisParams(overwritingParams, type) {
 overwriteGenesisParams(OVERWRITING_BLOCKCHAIN_PARAMS, 'blockchain');
 overwriteGenesisParams(OVERWRITING_CONSENSUS_PARAMS, 'consensus');
 
+// Note(minsu): If NON_COMPLETE_NETWORK env is set, it tightly limits the outbound connections.
+// Full network connections are convinced (especially for dev mode) otherwise.
+function initializeNetworkEnvronments() {
+  if (process.env.NON_COMPLETE_NETWORK) {
```
platfowner, line 412: What about NETWORK_OPTIMIZATION? 

And could you add this to https://github.com/ainblockchain/ain-blockchain/blob/develop/p2p/server.js#L184 ?

p2p/index.js
minsulee2: Removed it!

common/constants.js
minsulee2: MAX_CONNECTION_LIMIT is not necessary at all if we separately deal with inbound and outbound connections. I removed it.

common/constants.js
minsulee2: MAX_BOUND_LIMIT means the max connection of inbound and outbound? Could you check my updates that I am on the right track ?

common/constants.js
minsulee2: Updated!

minsulee2:
PTAL @platfowner I may've been somewhat confused.

p2p/index.js
```diff
@@ -46,20 +44,16 @@ class P2pClient {
   // maxOutbound is for now limited equal or less than 2.
   // maxInbound is a rest of connection after maxOutbound is set.
   initConnections() {
-    const numConnection = process.env.MAX_CONNECTION ?
-        Number(process.env.MAX_CONNECTION) : INITIAL_MAX_CONNECTION;
     const numOutbound = process.env.MAX_OUTBOUND ?
         Number(process.env.MAX_OUTBOUND) : INITIAL_MAX_OUTBOUND;
```
platfowner, line 48: minor: It looks like INITIAL_MAX_OUTBOUND can be renamed DEFAULT_MAX_OUTBOUND? Does it?

p2p/index.js
```diff
@@ -46,20 +44,16 @@ class P2pClient {
...
     const numInbound = process.env.MAX_INBOUND ?
         Number(process.env.MAX_INBOUND) : INITIAL_MAX_INBOUND;
```
platfowner, line 50: ditto

common/constants.js
platfowner: Thanks! Could you update other genesis params files as well?

common/constants.js
platfowner: Thanks.

I think we need to consider the following scenario too:
1) Some one updated MIN_NUM_VALIDATORS to 10 but some values in this code block is still 5

This can be resolved by setting all the values with MIN_NUM_VALIDATORS?

p2p/index.js
minsulee2: Done!

p2p/index.js
minsulee2: Done!

minsulee2:
PTAL. All addressed!

platfowner:
Thanks for the quick update. 
Could you take a look at the comment regarding MIN_NUM_VALIDATORS above?

common/constants.js
minsulee2: Oh, I missed this part. Updated!

minsulee2:
PTAL :) Really all addressed

platfowner (approved):
LGTM
Thanks!

liayoo (approved):
LGTM, thanks!

minsulee2:
Thanks for the reviews! Merging now :)

PR #231, opened 2021-02-17
Init load-test

cshcomcom:
This is simple load test tool from AIN benchmark.
It can be used without worker setup or a config file!


tools/simple-load-test/package.json
```diff
@@ -0,0 +1,14 @@
+{
```
liayoo, line 1: do we need a separate package.json for this tool script?

tools/simple-load-test/index.js
```diff
@@ -0,0 +1,126 @@
+/*
+For simple load testing
+This tool increases '/apps/loadtest/visit_count'
+Please modify numberOfTransactions & duration
```
liayoo, line 4: Could you add an example command here & maybe add an npm script to package.json?

tools/simple-load-test/index.js
```diff
@@ -0,0 +1,126 @@
+/*
```
platfowner, line 1: Could you use jsdoc format (/** ... */)?

tools/simple-load-test/package.json
platfowner: +1

tools/simple-load-test/index.js
```diff
@@ -0,0 +1,126 @@
...
+          }, 0, timestamp + i);
+        }),
+    );
+  }
+
+  await Promise.all(sendTxPromiseList);
+  return sendCnt;
+}
+
+async function main() {
```
platfowner, line 117: Could you add usage() with maybe for extra arguments (e.g. --help)?

liayoo: ref: https://github.com/ainblockchain/ain-blockchain/blob/master/tools/cross-shard/sendCheckinTx.js#L147

tools/simple-load-test/package.json
cshcomcom: Since 'ain-js' is not being used in blockchain, I tried to reduce unnecessary dependency. If we add 'ain-js' to the dependency, it seems inconvenient that 'ain-js' should also be updated every time when protocol version goes up.

tools/simple-load-test/index.js
cshcomcom: Looks good!

tools/simple-load-test/index.js
cshcomcom: Looks good!

tools/simple-load-test/index.js
cshcomcom: Done!

cshcomcom:
PTAL! @platfowner @liayoo

tools/simple-load-test/index.js
```diff
@@ -0,0 +1,136 @@
...
+ Please modify numberOfTransactions & duration
+ Example command line: 'node index.js'
+ */
+const Ain = require('@ainblockchain/ain-js').default;
+const delay = (time) => new Promise((resolve) => setTimeout(resolve, time));
+const testPath = '/apps/loadtest';
+const ainUrl = 'http://localhost:8081';
+const ainPrivateKey = '4207f5dcacb1b601d3a1f8cb10afaca158f6ebe383c0b30d02b39f8d2060cce3';
+const ainAddress = '0xF2be7f1356347a8960630c112AcB6Da61eE94632';
+const numberOfTransactions = 300;
```
platfowner, line 13: What about env-parmeterize these two parameters? e.g. NUM_TXS, DURATION
And ainUrl as well? e.g. TARGET_URL

tools/simple-load-test/index.js
```diff
@@ -0,0 +1,136 @@
+/**
+ For simple load testing
+ This tool increases '/apps/loadtest/visit_count'
+ Please modify numberOfTransactions & duration
+ Example command line: 'node index.js'
+ */
+const Ain = require('@ainblockchain/ain-js').default;
+const delay = (time) => new Promise((resolve) => setTimeout(resolve, time));
+const testPath = '/apps/loadtest';
+const ainUrl = 'http://localhost:8081';
```
platfowner, line 10: minor: What about targetUrl?

tools/simple-load-test/index.js
cshcomcom: I understood it as creating a'.env' file and putting 3 parameters into '.env' file. Is it right? or 'node index.js <TARGET_URL> <DURATION> <NUM_TXS>'? @platfowner

platfowner: I meant using just env variables like:
HOSTING_ENV=local DEBUG=false MIN_NUM_VALIDATORS=2 ACCOUNT_INDEX=0 node client/index.js

I just thought that it would be a good compromise as adding them as commandline arguments can make the interface complicated?

tools/simple-load-test/package.json
liayoo: I see. How about we remove ain-js dependency then? Other tool scripts are using axios & send tx utils can be found in [./tools/util.js](https://github.com/ainblockchain/ain-blockchain/blob/develop/tools/util.js)

tools/simple-load-test/index.js
cshcomcom: I think users can't know what variables exist. Even if it gets a little complicated, how about `node index.js <TARGET_URL> <NUM_TXS> <DURATION>`? I think It is more convenient for users to use!

tools/simple-load-test/index.js
cshcomcom: Looks better

tools/simple-load-test/index.js
platfowner: One drawback is that it's not optional anymore, i.e., always type in three arguments. Or you can try to use https://www.npmjs.com/package/command-line-args to allow optional arguments?

cshcomcom: Looks great!! I'll try it

cshcomcom:
PTAL! @liayoo @platfowner 
Now users can use like this
image

tools/simple-load-test/index.js
```diff
@@ -0,0 +1,197 @@
...
+    type: String,
+    description: 'Target AIN URL (ex: http://node.ainetwork.ai:8080)',
+    defaultOption: true,
+    group: 'required',
+  },
+  {
+    name: 'duration',
+    alias: 'd',
+    type: Number,
+    description: 'Duration of load test, (60 means 1 min, Default: 60)',
```
platfowner, line 35: `Duration of load test in seconds, (Default: 60)`?

tools/simple-load-test/index.js
```diff
@@ -0,0 +1,197 @@
...
+  {
+    name: 'help',
+    description: 'Display this usage guide.',
+    alias: 'h',
+  },
+  {
+    name: 'target_url',
+    alias: 't',
+    type: String,
+    description: 'Target AIN URL (ex: http://node.ainetwork.ai:8080)',
```
platfowner, line 27: To avoid unnecessary traffic to the testnet, could you remove the example?
And what about adding localhost:8081 as a default value?

tools/simple-load-test/index.js
```diff
@@ -0,0 +1,197 @@
+/**
+ For simple load testing
```
platfowner, line 2: minor: jsdoc style is
```
/**
 * Some descritions..
 */
```

tools/simple-load-test/index.js
```diff
@@ -0,0 +1,197 @@
...
+ This tool increases '/apps/loadtest/visit_count'
+ Please modify numberOfTransactions & duration
```
platfowner, line 4: number_txs?

tools/simple-load-test/index.js
cshcomcom: Looks better

tools/simple-load-test/index.js
cshcomcom: Most of users will test in a local environment, so it looks good!

tools/simple-load-test/index.js
cshcomcom: Thanks Done!

tools/simple-load-test/index.js
cshcomcom: Users can see options with `node index.js --help`. So I'm going to delete this comment!

tools/simple-load-test/index.js
platfowner: So for those users, what about `description: 'Target AIN URL (ex: http://localhost:8081)',`?
I think it won't be harmful to consider such users who just clone the repo and run the script without any serious consideration?

platfowner: i.e., to avoid unintentionally misleading users to use this script to send traffic to our testnet servers?

cshcomcom: Oh, I mean, it looks good, so I already changed it!

cshcomcom:
All comments are addressed! PTAL @platfowner

liayoo:
Could we use yarn instead of npm in simple-load-test? (I see package-lock.json!)

liayoo (approved):
LGTM with a comment. Thanks!

tools/simple-load-test/index.js
platfowner: Got it. Thanks!

platfowner (approved):
LGTM
Thanks!

cshcomcom:
Thanks for reviews! I'll merge this PR

PR #234, opened 2021-02-18
Feature/service accounts

liayoo:
DONE
- Added some initial code for /service_accounts feature
  - /service_accounts/\<service type>/\<service name>/\<key>: { admin: { \<address>: true }, balance: \<number> }
    - Example: `/service_accounts/payments/collaborative_ai/0xUSER...ADDR: { admin: { 0xADMIN...ADDR: true }, balance: 100 }`
  - Admin is set when a first transfer is made to the service account
  - When transferring from and/or to a service account, it should be encoded as follows: `<service type>|<service name>|<key>`
    - Example: `/transfer/0xADMIN...ADDR/payments|collaborative_ai|0xUSER...ADDR/123/value` will transfer AIN from `/accounts/0xADMIN...ADDR/balance` to `/service_accounts/payments/collaborative_ai/0xUSER...ADDR/balance`
- Replaced balances for payments service with service accounts
- Added tests for payments (_pay, _claim)
- Added { nonce: -1 } to some requests in ./integration/sharding.test.js.

TODO
- Deprecate deposit_accounts and use service_accounts instead
- Prevent infinite recursion of native function calls
- Decide on service name policy (additional restrictions, etc.)

### DIFF
--- a/common/chain-util.js
+++ b/common/chain-util.js
@@ -68,6 +68,10 @@ class ChainUtil {
     return ruleUtil.isCksumAddr(addr);
   }
 
+  static isServAcntName(name) {
+    return ruleUtil.isServAcntName(name);
+  }
+
   static isValShardProto(value) {
     return ruleUtil.isValShardProto(value);
   }
@@ -90,6 +94,10 @@ class ChainUtil {
 
   static toCksumAddr(addr) {
     return ruleUtil.toCksumAddr(addr);
+  }
+
+  static toServiceAccountName(serviceType, serviceName, user) {
+    return `${serviceType}|${serviceName}|${user}`;
   }
 
   static toString(value) {
@@ -150,6 +158,10 @@ class ChainUtil {
       labels.push(...ChainUtil.parsePath(toAppend));
     }
     return ChainUtil.formatPath(labels);
+  }
+
+  static getBalancePath(addrOrServAcnt) {
+    return ruleUtil.getBalancePath(addrOrServAcnt);
   }
 
   static getJsObject(obj, path) {
--- a/common/chain-util.js
+++ b/common/chain-util.js
@@ -96,6 +96,10 @@ class ChainUtil {
     return ruleUtil.toCksumAddr(addr);
   }
 
+  static parseServAcntName(accountName) {
+    return ruleUtil.parseServAcntName(accountName);
+  }
+
   static toServiceAccountName(serviceType, serviceName, user) {
     return `${serviceType}|${serviceName}|${user}`;
   }
--- a/common/chain-util.js
+++ b/common/chain-util.js
@@ -100,8 +100,8 @@ class ChainUtil {
     return ruleUtil.parseServAcntName(accountName);
   }
 
-  static toServiceAccountName(serviceType, serviceName, user) {
-    return `${serviceType}|${serviceName}|${user}`;
+  static toServiceAccountName(serviceType, serviceName, key) {
+    return `${serviceType}|${serviceName}|${key}`;
   }
 
   static toString(value) {
--- a/db/functions.js
+++ b/db/functions.js
@@ -77,7 +77,7 @@ class Functions {
           parsedValuePath, functionPath, timestamp, execTime, params, value, transaction);
       for (const functionEntry of functionList) {
         if (!functionEntry || !functionEntry.function_type) {
-          continue;  // Does nothing.
+          continue; // Does nothing.
         }
         if (functionEntry.function_type === FunctionTypes.NATIVE) {
           const nativeFunction = this.nativeFunctionMap[functionEntry.function_id];
@@ -85,7 +85,7 @@ class Functions {
             logger.info(
                 `  ==> Triggering NATIVE function '${functionEntry.function_id}' with\n` +
                 formattedParams);
-            const auth = { fid: functionEntry.function_id, };
+            const auth = { fid: functionEntry.function_id };
             // Execute the matched native function.
             nativeFunction(
                 value,
@@ -155,7 +155,6 @@ class Functions {
   /**
    * Returns a new function created by applying the function change to the current function.
    * 
-   * @param {Object} curFunction current function (modified and returned by this function)
    * @param {Object} functionChange function change
    */
   static applyFunctionChange(curFunction, functionChange) {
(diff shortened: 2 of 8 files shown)


common/chain-util.js
```diff
@@ -92,6 +96,10 @@ class ChainUtil {
     return ruleUtil.toCksumAddr(addr);
   }
 
+  static toServiceAccountName(serviceType, serviceName, user) {
```
platfowner, line 99: What about s/user/key/ as its value can be more than just a user address depending on service applications? (e.g. user_addr + address nonce)

db/functions.js
```diff
@@ -650,8 +672,18 @@ class Functions {
     return true;
   }
 
-  _getBalancePath(address) {
-    return `${PredefinedDbPaths.ACCOUNTS}/${address}/${PredefinedDbPaths.BALANCE}`;
+  _getServiceAccountAdminPath(accountName) {
```
platfowner, line 675: Suggestion: What about using function names without '_' prefix for non-native functions?

db/functions.js
```diff
@@ -246,6 +246,33 @@ class Functions {
...
+   * service-related native functions such as payments, deposit, and withdraw.
+   */
+  setServiceAccountTransferOrLog(from, to, value, auth, timestamp, transaction) {
+    const transferPath = this._getTransferValuePath(from, to, timestamp);
+    const transferResult = this.db.setValue(transferPath, value, auth, timestamp, transaction);
+    if (transferResult !== true) {
+      logger.error(
+          `  ==> Failed to setServiceAccountTransferOrLog on '${transferPath}' with error: ${JSON.stringify(transferResult)}`);
+    }
+    if (ChainUtil.isServAcntName(to)) {
```
platfowner, line 260: What about trying to set admin earlier than trying to set transfer value?

db/functions.js
```diff
@@ -246,6 +246,33 @@ class Functions {
...
+      const serviceAccountAdminPath = this._getServiceAccountAdminPath(to);
+      const serviceAccountAdmin = this.db.getValue(serviceAccountAdminPath);
+      if (serviceAccountAdmin === null) {
+        // set admin as the from address of the original transaction
+        const serviceAccountAdminAddrPath = this._getServiceAccountAdminAddrPath(to, transaction.address);
+        const adminSetupResult = this.setValueOrLog(serviceAccountAdminAddrPath, true, auth, timestamp);
+        if (adminSetupResult !== true) {
```
platfowner, line 267: Already logged by setValueOrLog()?

db/functions.js
platfowner: And if it fails, the transfer value setting needs to be skipped?

db/functions.js
```diff
@@ -246,6 +246,33 @@ class Functions {
...
   }
 
+  /**
+   * Adds a transfer entry from a service account to a regular account or vice versa. Used by 
+   * service-related native functions such as payments, deposit, and withdraw.
+   */
+  setServiceAccountTransferOrLog(from, to, value, auth, timestamp, transaction) {
+    const transferPath = this._getTransferValuePath(from, to, timestamp);
+    const transferResult = this.db.setValue(transferPath, value, auth, timestamp, transaction);
+    if (transferResult !== true) {
```
platfowner, line 256: Can be replaced by setValueOrLog()?

db/functions.js
```diff
@@ -650,8 +672,18 @@ class Functions {
     return true;
   }
 
-  _getBalancePath(address) {
-    return `${PredefinedDbPaths.ACCOUNTS}/${address}/${PredefinedDbPaths.BALANCE}`;
+  _getServiceAccountAdminPath(accountName) {
+    const parsed = accountName.split('|');
+    return `/service_accounts/${parsed[0]}/${parsed[1]}/${parsed[2]}/admin`;
```
platfowner, line 677: PredefinedDbPaths.SERVICE_ACCOUNTS?

Could you add admin to PredefinedDbPaths as well?

db/functions.js
```diff
@@ -650,8 +672,18 @@ class Functions {
...
+  }
+
+  _getServiceAccountAdminAddrPath(accountName, adminAddr) {
+    const parsed = accountName.split('|');
+    return `/service_accounts/${parsed[0]}/${parsed[1]}/${parsed[2]}/admin/${adminAddr}`;
```
platfowner, line 682: ditto

platfowner: Suggestion: What about handling exceptional cases to avoid out-of-bound errors just in case e.g. `_.get(parsed, '0', 'null');`?
Or defining separate util function like parseServAcntName() in ruleUtil that returns an array of size 3 with some default values like 'null'?

common/chain-util.js
liayoo: sgtm! done!

db/functions.js
liayoo: done!

db/functions.js
liayoo: I thought it should be the other way around, thinking anyone can try to make a transfer. But I guess only admins (via native functions) can trigger the transfer to/from service accounts by design. Switched the order & made it skip transfer if admin setting fails.

db/functions.js
liayoo: done!

db/functions.js
liayoo: done

db/functions.js
liayoo: done & done

db/functions.js
liayoo: thanks! made a parseServAcntName() util function

liayoo:

db/functions.js
platfowner: Thanks!

db/functions.js
```diff
@@ -246,6 +246,27 @@ class Functions {
...
+  setServiceAccountTransferOrLog(from, to, value, auth, timestamp, transaction) {
+    if (ChainUtil.isServAcntName(to)) {
+      const serviceAccountAdminPath = this.getServiceAccountAdminPath(to);
+      const serviceAccountAdmin = this.db.getValue(serviceAccountAdminPath);
+      if (serviceAccountAdmin === null) {
+        // set admin as the from address of the original transaction
+        const serviceAccountAdminAddrPath = this.getServiceAccountAdminAddrPath(to, transaction.address);
+        const adminSetupResult = this.setValueOrLog(serviceAccountAdminAddrPath, true, auth, timestamp);
+        if (adminSetupResult !== true) {
+          return;
```
platfowner, line 262: Would it be helpful to return adminSetupResult?

db/functions.js
```diff
@@ -650,88 +666,100 @@ class Functions {
...
-    return `${PredefinedDbPaths.ACCOUNTS}/${address}/${PredefinedDbPaths.BALANCE}`;
+  getServiceAccountAdminPath(accountName) {
+    const parsed = ChainUtil.parseServAcntName(accountName);
+    return `${PredefinedDbPaths.SERVICE_ACCOUNTS}/${parsed[0]}/${parsed[1]}/${parsed[2]}/` +
+        `${PredefinedDbPaths.SERVICE_ACCOUNTS_ADMIN}`;
+  }
+
+  getServiceAccountAdminAddrPath(accountName, adminAddr) {
+    const parsed = ChainUtil.parseServAcntName(accountName);
+    return `${PredefinedDbPaths.SERVICE_ACCOUNTS}/${parsed[0]}/${parsed[1]}/${parsed[2]}/` +
```
platfowner, line 677: minor: maybe getServiceAccountAdminPath() can be reused?

platfowner (approved):
LGTM w/ some minor comments.

db/functions.js
liayoo: done!

db/functions.js
liayoo: done!

liayoo:

PR #243, opened 2021-02-23
Feature/payment

liayoo:
- Allow multiple addresses to be service admin for payments
- Remove id fields from payment records
- Update tests

### DIFF
--- a/genesis-configs/base/genesis_rules.json
+++ b/genesis-configs/base/genesis_rules.json
@@ -83,7 +83,7 @@
       "config": {
         "admin": {
           "$admin_addr": {
-            ".write": "evalOwner('/payments/' + $service + '/config', 'write_owner', auth) && util.isBool(newData)"
+            ".write": "evalOwner('/payments/' + $service + '/config', 'write_owner', auth) && (newData === true || newData === null)"
           }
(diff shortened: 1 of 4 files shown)


genesis-configs/base/genesis_rules.json
```diff
@@ -81,23 +81,24 @@
   "payments": {
     "$service": {
       "config": {
-        ".write": "evalOwner('/payments/' + $service + '/config', 'write_owner', auth) && data === null && util.isDict(newData)"
+        "admin": {
+          "$admin_addr": {
+            ".write": "evalOwner('/payments/' + $service + '/config', 'write_owner', auth) && util.isBool(newData)"
```
platfowner, line 86: Suggestion: What about `newData === true` unless `false` value has special meaning?

platfowner: For deletion, `newData === true || newData === null`?

liayoo: Right, actually updated the rule but didn't commit 🤦‍♀️ Thanks!!

platfowner (approved):
LGTM
Thanks!

Issue #254, opened 2021-03-03
Need a stack for chained function calls 


platfowner:
We need a stack for chained function calls:
- To make them atomic (i.e., result consistency)
- To distinguish functionA -> functionB and functionC -> functionB calls in permission eval

platfowner:
Related issue: #240 

Issue resolved.

PR #257, opened 2021-03-03
Support call stack and fill auth.fids for chained function calls.

platfowner:
Summary:
- Add call stack
- Fill auth.fids
- Log function execution result

As a result, both auth.fid and auth.fids are available for .write rule configuration.

#254

### DIFF
--- a/db/functions.js
+++ b/db/functions.js
@@ -83,35 +83,39 @@ class Functions {
         if (functionEntry.function_type === FunctionTypes.NATIVE) {
           const nativeFunction = this.nativeFunctionMap[functionEntry.function_id];
           if (nativeFunction) {
+            // Execute the matched native function.
             logger.info(
                 `  ==> Triggering NATIVE function '${functionEntry.function_id}' with:\n` +
                 formattedParams);
             this.pushCall(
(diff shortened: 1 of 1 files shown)


db/functions.js
```diff
@@ -83,9 +84,13 @@ class Functions {
...
+                `  ==> Triggering NATIVE function '${functionEntry.function_id}' with:\n` +
                 formattedParams);
-            const newAuth = Object.assign({}, auth, { fid: functionEntry.function_id });
+            this.pushCall(
+                ChainUtil.formatPath(parsedValuePath), value, ChainUtil.formatPath(functionPath),
+                functionEntry.function_id);
+            const newAuth =
+                Object.assign({}, auth, { fid: functionEntry.function_id, fids: this.getFids() });
             // Execute the matched native function.
             nativeFunction(
```
liayoo, line 95: how about we wrap a try-catch around the nativeFunction() so popCall() can always be called? (just in case)

liayoo (approved):
LGTM with a comment, Thanks!

db/functions.js
platfowner: Just covered with try-finally blocks. Thanks for the suggestion!

PR #261, opened 2021-03-05
Feature/heartbeat again

minsulee2:
- Add heartbeat in the simplest way via socket
- More update will be next up when some jobs the tracker now doing is transplanted into peer nodes.
- Now, heartbeat just ping and get pong when the corresponding peer is alive.

### DIFF
--- a/p2p/index.js
+++ b/p2p/index.js
@@ -412,7 +412,14 @@ class P2pClient {
   heartbeat() {
     this.intervalHeartbeat = setInterval(() => {
       Object.values(this.outbound).forEach(socket => {
-        socket.ping();
+        if (socket.readyState !== 1) {
+          const account = this.getAccountFromSocket(socket);
+          this.removeFromOutboundIfExists(account);
+          logger.info(`A peer(${account}) is not ready to communicate with. ` +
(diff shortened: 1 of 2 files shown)


p2p/index.js
```diff
@@ -30,9 +30,7 @@ class P2pClient {
         this, node, minProtocolVersion, maxProtocolVersion, this.maxInbound);
     this.trackerWebSocket = null;
     this.outbound = {};
-    // XXX(minsu): The comment out will be revoked when next heartbeat updates.
-    // this.isAlive = true;
-    // this.heartbeat();   // XXX(minsu): it won't run before updating p2p network.
+    this.heartbeat();
```
platfowner, line 33: Suggestion: What about setIntervalHeartbeat()?
Or simply startHeartbeat() / stopHeartbeat()?

platfowner (approved):
LGTM with some comments.
Thanks!

p2p/index.js
minsulee2: I took startHeartbeat() and stopHeartbeat()! Thanks!

minsulee2:
All addressed. I am merging now.

PR #262, opened 2021-03-05
Feature/escrow

liayoo:
- Add escrow service type
  - _openEscrow
    -  ref: `/escrow/<source account>/<target account>/<key>/open`
    - value: `{ admin: { <address>: true } }`
  - _hold
    - ref: `/escrow/<source account>/<target account>/<key>/hold/<recordId>`
    - value: `{ amount: <number> }`
    - moves AIN from source account to escrow service account
  - _release
    - ref: `/escrow/<source account>/<target account>/<key>/release/<recordId>`
    - value: `{ ratio: <number> }`
    - moves AIN from escrow service account to source account and/or target account, according to the ratio (ratio = 1 means all goes to target, ratio = 0 means refund everything to source)
- Extend payments service
  - _pay now accepts 'source' field. If specified, it will try to transfer from 'source' to the payments service account
  - _claim can trigger _hold if 'escrow_key' field is in the value

### DIFF
--- a/db/functions.js
+++ b/db/functions.js
@@ -475,8 +475,7 @@ class Functions {
     const userServiceAccountName = ChainUtil.toServiceAccountName(
         PredefinedDbPaths.PAYMENTS, service, `${user}|${paymentKey}`);
     const transferResult = this.setServiceAccountTransferOrLog(
-        value.source ? value.source : transaction.address, userServiceAccountName, value.amount,
-        auth, timestamp, transaction);
+        transaction.address, userServiceAccountName, value.amount, auth, timestamp, transaction);
     if (transferResult === true) {
       this.saveAndSetExecutionResult(context, resultPath, FunctionResultCode.SUCCESS);
     } else {
@@ -508,7 +507,7 @@ class Functions {
     if (value.escrow_key !== undefined) {
       const escrowHoldPath = this.getEscrowHoldRecordPath(
           userServiceAccountName, value.target, value.escrow_key, timestamp);
-      result = this.setValueOrLog(escrowHoldPath, { amount: value.amount }, auth, timestamp, transaction);
+      result = this.setValueOrLog(escrowHoldPath, value.amount, auth, timestamp, transaction);
     } else {
       result = this.setServiceAccountTransferOrLog(
           userServiceAccountName, value.target, value.amount, auth, timestamp, transaction);
@@ -574,27 +573,6 @@ class Functions {
     }
   }
 
-  releaseInternal(escrowAccount, recipient, amount, auth, timestamp, transaction) {
-    if (ChainUtil.isServAcntName(recipient)) {
-      const parsed = ChainUtil.parseServAcntName(recipient);
-      if (parsed[0] === PredefinedDbPaths.PAYMENTS) {
-        // trigger _pay
-        const paymentAccountNameParsed = parsed[2].split('|');
-        const payPath = this.getPaymentPayRecordPath(
-            parsed[1], paymentAccountNameParsed[0], paymentAccountNameParsed[1], timestamp);
-        return this.setValueOrLog(payPath, { amount, source: escrowAccount }, auth, timestamp, transaction);
-      } else {
-        // other service types not yet supported
-        logger.debug(`  ==> Service type not supported: ${parsed[0]}`);
-        return false;
-      }
-    } else {
-      // transfer to regular accounts
-      return this.setServiceAccountTransferOrLog(
-          escrowAccount, recipient, amount, auth, timestamp, transaction);
-    }
-  }
-
   _release(value, context) {
     const sourceAccount = context.params.source_account;
     const targetAccount = context.params.target_account;
@@ -616,7 +594,7 @@ class Functions {
     logger.debug(`  =>> escrowAmount: ${escrowAmount}, ratio: ${ratio}, ` +
         `targetAmount: ${targetAmount}, sourceAmount: ${sourceAmount}`);
     if (targetAmount > 0) {
-      const result = this.releaseInternal(
+      const result = this.setServiceAccountTransferOrLog(
           escrowServiceAccountName, targetAccount, targetAmount, auth, timestamp, transaction);
       if (result !== true) {
         this.saveAndSetExecutionResult(context, resultPath, FunctionResultCode.INTERNAL_ERROR);
@@ -624,7 +602,7 @@ class Functions {
       }
     }
     if (sourceAmount > 0) {
-      const result = this.releaseInternal(
+      const result = this.setServiceAccountTransferOrLog(
           escrowServiceAccountName, sourceAccount, sourceAmount, auth, timestamp, transaction);
       if (result !== true) {
         this.saveAndSetExecutionResult(context, resultPath, FunctionResultCode.INTERNAL_ERROR);
--- a/db/functions.js
+++ b/db/functions.js
@@ -583,6 +583,7 @@ class Functions {
         return this.setValueOrLog(payPath, { amount, source: escrowAccount }, auth, timestamp, transaction);
       } else {
         // other service types not yet supported
+        logger.debug(`  ==> Service type not supported: ${parsed[0]}`);
         return false;
       }
     } else {
--- a/db/functions.js
+++ b/db/functions.js
@@ -503,6 +503,8 @@ class Functions {
     let result;
     const userServiceAccountName = ChainUtil.toServiceAccountName(
         PredefinedDbPaths.PAYMENTS, service, `${user}|${paymentKey}`);
+    // NOTE: By specifying `escrow_key`, the claimed payment is held in escrow instead of being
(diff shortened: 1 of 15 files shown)


db/functions.js
```diff
@@ -462,16 +465,18 @@ class Functions {
...
       this.saveAndSetExecutionResult(context, resultPath, FunctionResultCode.FAILURE);
       return;
     }
 
-    const userServiceAccountName = ChainUtil.toServiceAccountName('payments', service, `${user}|${paymentKey}`);
+    const userServiceAccountName = ChainUtil.toServiceAccountName(
+        PredefinedDbPaths.PAYMENTS, service, `${user}|${paymentKey}`);
     const transferResult = this.setServiceAccountTransferOrLog(
-      transaction.address, userServiceAccountName, value.amount, auth, timestamp, transaction);
+        value.source ? value.source : transaction.address, userServiceAccountName, value.amount,
```
platfowner, line 478: Question: Has the input format for pay changed?
(It looks like there is no updates in sendPayTx.js)

db/functions.js
```diff
@@ -488,17 +493,25 @@ class Functions {
...
     }
 
-    const userServiceAccountName = ChainUtil.toServiceAccountName('payments', service, `${user}|${paymentKey}`);
-    const transferResult = this.setServiceAccountTransferOrLog(
-        userServiceAccountName, value.target, value.amount, auth, timestamp, transaction);
-    if (transferResult === true) {
+    let result;
+    const userServiceAccountName = ChainUtil.toServiceAccountName(
+        PredefinedDbPaths.PAYMENTS, service, `${user}|${paymentKey}`);
+    if (value.escrow_key !== undefined) {
```
platfowner, line 506: Would it be helpful for new comers if some notes are left that escrow_key switches between direct pay and escrowed pay?

platfowner: And add some use case to sendPayTx.js?

db/functions.js
```diff
@@ -518,6 +531,104 @@ class Functions {
...
+  _openEscrow(value, context) {
+    const sourceAccount = context.params.source_account;
+    const targetAccount = context.params.target_account;
+    const escrowKey = context.params.escrow_key;
+    const { timestamp, auth } = context;
+    const accountKey = ChainUtil.toEscrowAccountName(sourceAccount, targetAccount, escrowKey);
+    const serviceAccountName = ChainUtil.toServiceAccountName(
+        PredefinedDbPaths.ESCROW, PredefinedDbPaths.ESCROW, accountKey);
+    const serviceAccountPath = this.getServiceAccountPath(serviceAccountName);
+    const serviceAccountSetupResult = this.setValueOrLog(serviceAccountPath, value, auth, timestamp);
```
platfowner, line 543: Do we need some validation check for the value?

db/functions.js
```diff
@@ -518,6 +531,104 @@ class Functions {
...
+    if (serviceAccountSetupResult !== true) {
+      return serviceAccountSetupResult;
+    }
+  }
```
platfowner, line 547: If we don't need result saving, could you call setExecutionResult() (ref: _openCheckin, _closeCheckin)?

db/functions.js
```diff
@@ -518,6 +531,104 @@ class Functions {
...
+    if (ChainUtil.isServAcntName(recipient)) {
+      const parsed = ChainUtil.parseServAcntName(recipient);
+      if (parsed[0] === PredefinedDbPaths.PAYMENTS) {
+        // trigger _pay
+        const paymentAccountNameParsed = parsed[2].split('|');
+        const payPath = this.getPaymentPayRecordPath(
+            parsed[1], paymentAccountNameParsed[0], paymentAccountNameParsed[1], timestamp);
+        return this.setValueOrLog(payPath, { amount, source: escrowAccount }, auth, timestamp, transaction);
+      } else {
+        // other service types not yet supported
```
platfowner, line 582: What about leaving some log messages?

db/index.js
```diff
@@ -35,7 +35,7 @@ const RuleUtil = require('./rule-util');
 class DB {
   constructor(stateRoot, stateVersion, bc, tp, isNodeDb, blockNumberSnapshot) {
     this.shardingPath = null;
-    this.isRootBlockchain = null;  // Is this the database of the root blockchain?
+    this.isRootBlockchain = null; // Is this the database of the root blockchain?
```
platfowner, line 38: tip: sometimes there are cases to add some consecutive spaces intentionally to separate comment from code from the readability perspective. :)

tools/util.js
```diff
@@ -20,7 +20,9 @@ function signAndSendTx(endpointUrl, txBody, privateKey) {
       id: 0
     }
   ).then((resp) => {
-    const success = !ChainUtil.transactionFailed(_.get(resp, 'data.result'), null);
+    const success = Array.isArray(_.get(resp, 'data.result.result')) ?
```
platfowner, line 23: question: is there array case?

integration/node.test.js
```diff
@@ -2793,5 +2860,337 @@ describe('Blockchain Node', () => {
...
+            value: {
+              amount: paymentBalanceBefore
+            }
+          }}).body.toString('utf-8'));
+          expect(body.code).to.equals(0);
+          waitUntilTxFinalized(serverList, body.result.tx_hash);
+          // release
+          key = Date.now();
+          const releaseRef = `/escrow/${source}/${target}/1/release/${key}`;
+          paymentBalanceBefore = parseOrLog(syncRequest('GET', server1 +
```
platfowner, line 3162: const?

db/functions.js
```diff
@@ -518,6 +531,104 @@ class Functions {
...
+    } else {
+      this.saveAndSetExecutionResult(context, resultPath, FunctionResultCode.INTERNAL_ERROR);
+    }
+  }
+
+  releaseInternal(escrowAccount, recipient, amount, auth, timestamp, transaction) {
+    if (ChainUtil.isServAcntName(recipient)) {
+      const parsed = ChainUtil.parseServAcntName(recipient);
+      if (parsed[0] === PredefinedDbPaths.PAYMENTS) {
+        // trigger _pay
```
platfowner, line 576: As this is not exactly 'pay' (but 'rebund'), what about just using _transfer?

platfowner: And this can keep the escrow service service-independent?

db/functions.js
liayoo: Right, added some notes and updated sendClaimTx.js to accept --escrow option!

db/functions.js
liayoo: Added the validation check in the write rules, since we can't really rollback state changes made at this point. (Could be a Q2 project..?)

db/functions.js
liayoo: Done, thanks!

db/functions.js
liayoo: This function is replaced with setServiceAccountTransferOrLog().

db/index.js
liayoo: Ah, I didn't realize this was intentional. Thanks for letting me know! Reverted the change.

tools/util.js
liayoo: I got confused with the batch txs (which are supported by a separate API).. Removed the isArray() check.

integration/node.test.js
liayoo: done!

db/functions.js
liayoo: SGTM. Changing it to trigger _transfer made the releaseInternal() unnecessary, so I just removed the function.

db/functions.js
liayoo: As _pay is no longer triggered by _release, there is no use case for having 'source' specified in pay requests, so reverted the feature as well.

liayoo:

tools/payments/sendClaimTx.js
```diff
@@ -32,15 +65,21 @@ async function sendTransaction() {
...
+  if (process.argv.length !== 3 && process.argv.length !== 4) {
+    usage();
+  }
+  if (process.argv.length === 4 && process.argv[3] !== '--escrow') {
+    console.log('Invalid option:', process.argv[3]);
     usage();
   }
   config = require(path.resolve(__dirname, process.argv[2]));
-  await sendTransaction();
+  await sendTransaction(process.argv[3]);
```
platfowner, line 76: minor: what about `!!process.argv[3]`?

platfowner: or `process.argv.length === 4`?

db/functions.js
platfowner: Right, that an issue we have regarding chained function calls. Actually, I have a plan to make the whole chain (the initial tx + ft call chain) to be reverted if any part fails.
Until then, what about adding a TODO for the validation check (e.g. valid admins)?

platfowner (approved):
Thanks for the nice code!

liayoo:
Thanks for the thorough review!! Now merging..

Issue #240, opened 2021-02-23
Need to allow chained triggering of native functions but not to allow circular triggering 


platfowner:
Related issues: #217

platfowner:
Fixed.

PR #270, opened 2021-03-10
Do not allow circular function calls

platfowner:
Summary:
- Do not allow circular function calls
- Add fidList to call stack 

#240 
#254


minsulee2 (approved):
LGTM

liayoo (approved):
LGTM, thanks!

Issue #242, opened 2021-02-23
Do not allow set_function with native functions by users 


platfowner:
Related issues: #217

platfowner:
Idea1: Ref pattern check in native function? e.g. _transfer checks "/transfer/..." 
Idea2: Sanity check with native function ids in SET_FUNCTION

platfowner:
Fixed with ownerOnly bit.

Issue #195, opened 2021-02-03
Need to support explicit function permission 


platfowner:
**Is your feature request related to a problem? Please describe.**
- Control write permissions of native function with explicit rules

e.g. `“.write”: “auth.fid === ‘_transfer’”`

PR #275, opened 2021-03-11
Add test cases for function permission

platfowner:
Summary:
- Add test cases for owner only function bits
- Add test cases for .write rules with auth.fid & auth.fids 
- Add ruleUtil.includes() util function
- Tweak test code and names 

#242 
#195


liayoo (approved):
LGTM, thanks!

Issue #264, opened 2021-03-08
[BUG] Stale state versions 


platfowner:
In prod-summer R37:

![num_versions](https://user-images.githubusercontent.com/46211617/110320010-020d5080-8053-11eb-97a0-babe0a1d4f98.jpg)

node 0:
```
                "stateVersionStatus": {
                    "numVersions": 6,
                    "versionList": [
                        "EMPTY",
                        "POOL:93095:93096:1615205453225:0",
                        "POOL:93096:93097:1615205458878:0",
                        "POOL:93097:93098:1615205459986:0",
                        "FINAL:93096",
                        "NODE:93096"
                    ],
                    "finalVersion": "FINAL:93096"
                }
```

node 1:
```
                "stateVersionStatus": {
                    "numVersions": 35,
                    "versionList": [
                        "EMPTY",
                        "POOL:90075:90076:1615195399617:0",
                        "POOL:90084:90085:1615195428593:0",
                        "POOL:90090:90091:1615195452571:0",
                        "POOL:90171:90172:1615195762972:0",
                        "POOL:90212:90213:1615195917814:0",
                        "POOL:90252:90253:1615196058514:0",
                        "POOL:90260:90261:1615196092951:0",
                        "POOL:90268:90269:1615196116973:0",
                        "POOL:90338:90339:1615196376852:0",
                        "POOL:90357:90358:1615196440020:0",
                        "POOL:90402:90403:1615196612077:0",
                        "POOL:90429:90430:1615196718837:0",
                        "POOL:90519:90520:1615197055743:0",
                        "POOL:90675:90676:1615197634796:0",
                        "POOL:90814:90815:1615198084914:0",
                        "POOL:90826:90827:1615198136952:0",
                        "POOL:90831:90832:1615198157257:0",
                        "POOL:90862:90863:1615198274172:0",
                        "POOL:90864:90865:1615198279923:0",
                        "POOL:90926:90927:1615198496809:0",
                        "POOL:91861:91862:1615201558502:0",
                        "POOL:91933:91934:1615201777350:0",
                        "POOL:91960:91961:1615201864373:0",
                        "POOL:92040:92041:1615202125167:0",
                        "POOL:92057:92058:1615202184722:0",
                        "POOL:92098:92099:1615202332413:0",
                        "POOL:92123:92124:1615202427560:0",
                        "POOL:92144:92145:1615202501634:0",
                        "POOL:92158:92159:1615202546154:0",
[...]

PR #277, opened 2021-03-15
Bugfix/consensus

liayoo:
- Set hashToDb only if addSeenBlock was successful
- Call destroyDb() before returning in checkProposal()

Related: https://github.com/ainblockchain/ain-blockchain/issues/264

### DIFF
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -531,10 +531,12 @@ class Consensus {
     const newDb = this.node.createTempDb(baseVersion, newVersion, prevBlock.number);
     if (!newDb.executeTransactionList(proposalBlock.last_votes)) {
       logger.error(`[${LOG_HEADER}] Failed to execute last votes`);
+      this.node.destroyDb(newDb);
       return false;
     }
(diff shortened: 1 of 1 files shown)


consensus/index.js
```diff
@@ -563,13 +565,14 @@ class Consensus {
         logger.error(`[${LOG_HEADER}] State proof hashes don't match: ` +
             `${newDb.getProof('/')[ProofProperties.PROOF_HASH]} / ` +
             `${proposalBlock.state_proof_hash}`);
+        this.node.destroyDb(newDb);
         return false;
       }
     }
-    this.blockPool.hashToDb.set(proposalBlock.hash, newDb);
     if (!this.blockPool.addSeenBlock(proposalBlock, proposalTx)) {
       return false;
```
platfowner, line 573: Do we need destroyDb() before this return?

liayoo: Right, thanks for catching that! Added destroyDb().

liayoo:

platfowner (approved):
LGTM
Thanks!

PR #278, opened 2021-03-16
Feature/signature for network

minsulee2:


p2p/index.js
```diff
@@ -196,44 +197,70 @@ class P2pClient {
...
-        type: MessageTypes.CONSENSUS,
-        message: msg,
-        protoVer: CURRENT_PROTOCOL_VERSION
-      }));
+    const payload = {
+      type: MessageTypes.CONSENSUS,
+      message: msg,
+      protoVer: CURRENT_PROTOCOL_VERSION
+    };
+    payload.signature = this._signPayload(payload);
```
platfowner, line 221: Could you consider separating signing body from others? e.g.
```
const payload = {
    body: {
        type: MessageTypes.CONSENSUS,
        message: msg,
        protoVer: CURRENT_PROTOCOL_VERSION
    },
    address: <address>,
    signature: <signature>
}
```

p2p/index.js
```diff
@@ -196,44 +197,70 @@ class P2pClient {
     });
   }
 
+  _signPayload(payload) {
```
platfowner, line 200: Could you move util functions to p2p/util.js?

platfowner: Or move to chain-util.js as discussed offline?

minsulee2: As discussed offline, all solved!

p2p/index.js
minsulee2: It is decided that only the address is signed and verified. I couldn't take this.

minsulee2:
PTAL. All addressed discussed offline

p2p/server.js
```diff
@@ -281,29 +287,44 @@ class P2pServer {
...
-                account: this.getNodeAddress(),
+              if (!verifyData(data)) {
+                logger.error('The message is not correctly signed. Discard the message!!');
+                return;
+              }
+              logger.info(`A new websocket(${data.address}) is established.`);
+              this.inbound[data.address] = socket;
+              const payload = {
+                type: MessageTypes.ADDRESS_RESPONSE,
+                address: this.getNodeAddress(),
```
liayoo, line 323: should we maybe add a timestamp to the payload so we can ignore messages that are too old & prevent mitm attacks?

p2p/util.js
```diff
@@ -86,9 +87,37 @@ function sendGetRequest(endpoint, method, params) {
...
+  }
+}
+
+function signPayload(privateKey, payload) {
+  const keyBuffer = Buffer.from(privateKey, 'hex');
+  const stringPayload = JSON.stringify(payload);
+  return ainUtil.ecSignMessage(stringPayload, keyBuffer);
+}
+
+function verifyData(data) {
```
platfowner, line 107: 'data' sound like too broad, what about something like 'verifySignedMessage'?

p2p/util.js
```diff
@@ -86,9 +87,37 @@ function sendGetRequest(endpoint, method, params) {
...
+function removeSocketConnectionIfExists(connectionObj, address) {
+  if (address in connectionObj) {
+    delete connectionObj[address];
+    logger.info(` => Updated managed peers info: ${Object.keys(connectionObj)}`);
+  }
+}
+
+function signPayload(privateKey, payload) {
+  const keyBuffer = Buffer.from(privateKey, 'hex');
+  const stringPayload = JSON.stringify(payload);
```
platfowner, line 103: It seems stringify() can be skipped here as it's done by ecSignMessage().

p2p/util.js
```diff
@@ -86,9 +87,37 @@ function sendGetRequest(endpoint, method, params) {
...
+  return ainUtil.ecSignMessage(stringPayload, keyBuffer);
+}
+
+function verifyData(data) {
+  const signature = data.signature;
+  const address = data.address;
+  delete data.signature;
+  return ainUtil.ecVerifySig(JSON.stringify(data), signature, address);
```
platfowner, line 111: It seems stringify() can skipped here as well.

p2p/util.js
```diff
@@ -86,9 +87,37 @@ function sendGetRequest(endpoint, method, params) {
...
+}
+
+function removeSocketConnectionIfExists(connectionObj, address) {
+  if (address in connectionObj) {
+    delete connectionObj[address];
+    logger.info(` => Updated managed peers info: ${Object.keys(connectionObj)}`);
+  }
+}
+
+function signPayload(privateKey, payload) {
```
platfowner, line 101: What about 'signMessage()'?
And locating 'message' argument first as other functions like ecSignMessage() or signTransaction()?

p2p/util.js
```diff
@@ -86,9 +87,37 @@ function sendGetRequest(endpoint, method, params) {
   });
 }
 
+function getAddressFromSocket(connectionObj, socket) {
+  return Object.keys(connectionObj).filter(address => connectionObj[address] === socket);
```
platfowner, line 91: Doesn't filter() return an array?
What about more explicitly like
```
  const filtered = Object.keys(connectionObj).filter(address => connectionObj[address] === socket);
  if (filtered.length > 0) {
    return filtered[0];
  }
  return null;
```

liayoo: We could use [.find()](https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/Array/find) !

p2p/index.js
```diff
@@ -197,43 +203,48 @@ class P2pClient {
...
-      type: MessageTypes.ACCOUNT_REQUEST,
-      account: account,
+  sendAddress(socket) {
+    const payload = {
+      type: MessageTypes.ADDRESS_REQUEST,
+      address: this.server.getNodeAddress(),
       protoVer: CURRENT_PROTOCOL_VERSION
-    }));
+    };
+    payload.signature = signPayload(this.server.node.account.private_key, payload);
```
platfowner, line 246: Could you consider separating signature from the signed body? e.g.
```
const body = {
  type: MessageTypes.ADDRESS_REQUEST,
  address: this.server.getNodeAddress(),
  timestamp: <timestamp>,
};
const signature = signPayload(this.server.node.account.private_key, body);
const payload = {
  type: MessageTypes.ADDRESS_REQUEST,
  body,
  signature,
  protoVer: CURRENT_PROTOCOL_VERSION
};
```

p2p/server.js
```diff
@@ -281,29 +287,44 @@ class P2pServer {
...
                 protoVer: CURRENT_PROTOCOL_VERSION
-              }));
+              };
+              payload.signature = signPayload(this.node.account.private_key, payload);
```
platfowner, line 326: Could you consider separating signature from signed message body here as well?

And adding timestamp too?

p2p/server.js
minsulee2: The timestamp is added though it seems we need to talk about how to define "old".

p2p/util.js
minsulee2: Done!

p2p/util.js
minsulee2: It cannot be signed if stringify() is not used. Am I doing something wrong?

p2p/util.js
minsulee2: Done!

p2p/index.js
minsulee2: Done!

p2p/server.js
minsulee2: Done!

p2p/util.js
minsulee2: Done!

minsulee2:
PTAL. All addressed.

p2p/util.js
liayoo: Right. signMessage() doesn't stringify() the message, unlike signTransaction().

platfowner: Oh, I was confused with signTransaction() case. Just ignore my previous comment.

p2p/server.js
liayoo: Yup, how about 3 sec? It seems 3s is more than enough to reach other nodes across continents.. 
cc: @platfowner

p2p/util.js
```diff
@@ -86,9 +87,32 @@ function sendGetRequest(endpoint, method, params) {
...
+    logger.info(` => Updated managed peers info: ${Object.keys(connectionObj)}`);
+  }
+}
+
+function signMessage(messageBody, privateKey) {
+  return ainUtil.ecSignMessage(JSON.stringify(messageBody), Buffer.from(privateKey, 'hex'));
+}
+
+function verifySignedMessage(message) {
+  return ainUtil.ecVerifySig(JSON.stringify(message.body), message.signature, message.body.address);
```
platfowner, line 106: The address can be obtained from body hash and signature. (ref: https://github.com/ainblockchain/ain-blockchain/blob/master/tx-pool/transaction.js#L73).
As it's a general function (not specific for `ADDRESS_*` messages) for sign verification, what about taking a similar approach? In that sense, it seems we can skip the `address` field from the message body and add message type `type: MessageTypes.ADDRESS_REQUEST` instead?

p2p/index.js
```diff
@@ -242,24 +258,37 @@ class P2pClient {
...
         socket.close();
         return;
       }
 
       switch (data.type) {
-        case MessageTypes.ACCOUNT_RESPONSE:
-          if (!data.account) {
-            logger.error(`Broken websocket(account unknown) is established.`);
+        case MessageTypes.ADDRESS_RESPONSE:
+          if (!data.body.address) {
```
platfowner, line 276: What about using `_.get()` as we cannot guarantee that the peer is always good one?

p2p/server.js
```diff
@@ -281,29 +291,49 @@ class P2pServer {
...
           socket.close();
           return;
         }
 
         switch (data.type) {
-          case MessageTypes.ACCOUNT_REQUEST:
-            if (!data.account) {
-              logger.error(`Broken websocket(account unknown) is established.`);
+          case MessageTypes.ADDRESS_REQUEST:
+            if (!data.body.address) {
```
platfowner, line 309: ditto

p2p/util.js
platfowner: Discussed offline: Let's keep `address` in the body as it can be hijacked somehow.
Then it needs to be compared with the address from the signature.

minsulee2: Gracefully done! Thanks for the offline discussion!

p2p/index.js
minsulee2: Done!

p2p/server.js
minsulee2: Done!

minsulee2:
PTAL again. All addressed!

platfowner (approved):
LGTM

p2p/util.js
```diff
@@ -86,9 +88,39 @@ function sendGetRequest(endpoint, method, params) {
...
+}
+
+function getAddressFromSignature(message) {
+  const hashedMessage = ainUtil.hashMessage(JSON.stringify(message.body));
+  // TODO(minsu): getAddress should be in the chain-util??
+  return Transaction.getAddress(hashedMessage, message.signature);
+}
+
+function verifySignedMessage(message, address) {
+  return ainUtil.ecVerifySig(JSON.stringify(message.body), message.signature, address);
```
liayoo, line 113: Should we compare the timestamp with the current timestamp here? Or leave a todo for now?

minsulee2: I left TODO comments on both client and server. check check please !! :)

liayoo (approved):
LGTM, thanks!! 👍

minsulee2:
Thanks for the long reviewsssss. merging it now!

Issue #228, opened 2021-02-16
[BUG] Blockchain node should use nonce tracker's nonce in transaction creation


platfowner:
In createSingleTransaction(), nonce tracker's nonce should be used to be consistent.

liayoo:
Fixed in #281

Issue #227, opened 2021-02-16
Need to have a better way of tracking nonces and timestamps


liayoo:
Currently, we only have two types of nonce trackers (committed and pending). A problem arises since we now have blocks in BlockPool that are not yet finalized but could be notarized (voted for) and could be the previous block of a newly proposed block (due to changes in the consensus protocol). 
Without keeping track of nonces & timestamps updated by the blocks in the BlockPool, we can only compare new transactions with the committed / pending nonces, which would be incorrect. I propose that we replace the nonce trackers with the db state, and keep track of nonces and timestamps at paths such as these:
- /accounts/{address}/nonce
- /accounts/{address}/timestamp

Related: https://github.com/ainblockchain/ain-blockchain/issues/213

liayoo:
Implemented in #281

PR #281, opened 2021-03-18
Feature/block level nonce tracking

liayoo:
Summary:
- Migrate nonce tracking to db states for block-level nonce tracking
  - DB path: `/accounts/${address}/nonce`
- Deprecate committedNonceTracker and pendingNonceTracker in tx pool
- Revamp getValidTransactions() in tx pool
- Support loosely ordered noncing (nonce = -2)
  - Ordered by timestamp tracked in the states
  - DB path: `/accounts/${address}/timestamp`
- Update nonce API
- Fix bugs in transaction APIs
- Fix transaction execution bug in consensus checkProposal() by using a backup db & reverting when execution fails
- Remove unnecessary syncDb() in node addNewBlock()

Related issues: 
#227 
#213 
#228

### DIFF
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -316,7 +316,7 @@ class Consensus {
           return null;
         }
       } else {
-        logger.error(`[${LOG_HEADER}] tx: failure\n ${JSON.stringify(txRes)}`);
+        logger.debug(`[${LOG_HEADER}] tx: failure\n ${JSON.stringify(txRes)}`);
         invalidTransactions.push(tx);
         tempDb.setStateVersion(backupRoot, backupVersion);
         this.node.stateManager.deleteVersion(tempVersion);
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -303,8 +303,7 @@ class Consensus {
       } else {
         logger.error(`[${LOG_HEADER}] tx: failure\n ${JSON.stringify(txRes)}`);
         invalidTransactions.push(tx);
-        this.node.stateManager.renameVersion(backupVersion, tempVersion);
-        this.node.stateManager.deleteVersion(backupVersion);
+        tempDb.setStateVersion(backupRoot, backupVersion);
         backupRoot = this.node.stateManager.cloneVersion(tempVersion, backupVersion);
         if (!backupRoot) {
           logger.error(`[${LOG_HEADER}] Failed to clone state version: ${tempVersion}`);
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -239,16 +239,7 @@ class Consensus {
 
   executeLastVoteOrAbort(db, tx) {
     const LOG_HEADER = 'executeLastVoteOrAbort';
-    const dbVersion = db.stateVersion;
-    const backupVersion = this.node.stateManager.createUniqueVersionName(
-      `${StateVersions.BACKUP}:${dbVersion}`);
-    const backupRoot = this.node.stateManager.cloneVersion(dbVersion, backupVersion);
-    if (!backupRoot) {
-      logger.error(`[${LOG_HEADER}] Failed to clone state version: ${dbVersion}`);
-      return false;
-    }
     const txRes = db.executeTransaction(tx);
-    this.node.stateManager.deleteVersion(backupVersion);
     if (!ChainUtil.transactionFailed(txRes)) {
       logger.debug(`[${LOG_HEADER}] tx: success`);
       return true;
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -237,6 +237,52 @@ class Consensus {
     }
   }
 
+  executeLastVoteOrAbort(db, tx) {
+    const LOG_HEADER = 'executeLastVoteOrAbort';
+    const dbVersion = db.stateVersion;
+    const backupVersion = this.node.stateManager.createUniqueVersionName(
+      `${StateVersions.BACKUP}:${dbVersion}`);
+    const backupRoot = this.node.stateManager.cloneVersion(dbVersion, backupVersion);
+    if (!backupRoot) {
+      logger.error(`[${LOG_HEADER}] Failed to clone state version: ${dbVersion}`);
+      return false;
+    }
+    const txRes = db.executeTransaction(tx);
+    this.node.stateManager.deleteVersion(backupVersion);
+    if (!ChainUtil.transactionFailed(txRes)) {
+      logger.debug(`[${LOG_HEADER}] tx: success`);
+      return true;
+    } else {
+      logger.error(`[${LOG_HEADER}] tx: failure\n ${JSON.stringify(txRes)}`);
+      return false;
+    }
+  }
+
+  executeOrRollbackTransaction(db, tx, validTransactions, invalidTransactions) {
+    const LOG_HEADER = 'executeOrRollbackTransaction';
+    const dbVersion = db.stateVersion;
+    const backupVersion = this.node.stateManager.createUniqueVersionName(
+      `${StateVersions.BACKUP}:${dbVersion}`);
+    const backupRoot = this.node.stateManager.cloneVersion(dbVersion, backupVersion);
+    if (!backupRoot) {
+      logger.error(`[${LOG_HEADER}] Failed to clone state version: ${dbVersion}`);
+      return false;
+    }
+    logger.debug(`[${LOG_HEADER}] Checking tx ${JSON.stringify(tx, null, 2)}`);
+    const txRes = db.executeTransaction(tx);
+    if (!ChainUtil.transactionFailed(txRes)) {
+      logger.debug(`[${LOG_HEADER}] tx: success`);
+      validTransactions.push(tx);
+      this.node.stateManager.deleteVersion(backupVersion);
+    } else {
+      logger.debug(`[${LOG_HEADER}] tx: failure\n ${JSON.stringify(txRes)}`);
+      invalidTransactions.push(tx);
+      db.setStateVersion(backupRoot, backupVersion);
+      this.node.stateManager.deleteVersion(dbVersion);
+    }
+    return true;
+  }
+
   // proposing for block #N :
   //    1. create a block (with last_votes)
   //    2. create a tx (/consensus/number/N/propose: { block_hash, ... })
@@ -272,28 +318,10 @@ class Consensus {
       lastVotes.unshift(lastBlockInfo.proposal);
     }
 
-    const backupVersion = this.node.stateManager.createUniqueVersionName(
-      `${StateVersions.BACKUP}:${tempVersion}`);
-    let backupRoot = this.node.stateManager.cloneVersion(tempVersion, backupVersion);
-    if (!backupRoot) {
-      logger.error(`[${LOG_HEADER}] Failed to clone state version: ${tempVersion}`);
-      this.node.destroyDb(tempDb);
-      return null;
-    }
     for (const voteTx of lastVotes) {
-      const txRes = tempDb.executeTransaction(voteTx);
-      this.node.stateManager.deleteVersion(backupVersion);
-      if (!ChainUtil.transactionFailed(txRes)) {
-        logger.debug(`[${LOG_HEADER}] last vote: success`);
-        backupRoot = this.node.stateManager.cloneVersion(tempVersion, backupVersion);
-        if (!backupRoot) {
-          logger.error(`[${LOG_HEADER}] Failed to clone state version: ${tempVersion}`);
-          this.node.stateManager.deleteVersion(tempVersion);
-          return null;
-        }
-      } else {
-        logger.error(`[${LOG_HEADER}] last vote: failure\n ${JSON.stringify(txRes)}`);
-        this.node.stateManager.deleteVersion(tempVersion);
+      const res = this.executeLastVoteOrAbort(tempDb, voteTx);
+      if (!res) {
+        this.node.destroyDb(tempDb);
         return null;
       }
     }
@@ -303,26 +331,12 @@ class Consensus {
     const validTransactions = [];
     const invalidTransactions = [];
     for (const tx of transactions) {
-      logger.debug(`[${LOG_HEADER}] Checking tx ${JSON.stringify(tx, null, 2)}`);
-      const txRes = tempDb.executeTransaction(tx);
-      if (!ChainUtil.transactionFailed(txRes)) {
-        logger.debug(`[${LOG_HEADER}] tx: success`);
-        validTransactions.push(tx);
-        this.node.stateManager.deleteVersion(backupVersion);
-        backupRoot = this.node.stateManager.cloneVersion(tempVersion, backupVersion);
-        if (!backupRoot) {
-          logger.error(`[${LOG_HEADER}] Failed to clone state version: ${tempVersion}`);
-          this.node.stateManager.deleteVersion(tempVersion);
-          return null;
-        }
-      } else {
-        logger.debug(`[${LOG_HEADER}] tx: failure\n ${JSON.stringify(txRes)}`);
-        invalidTransactions.push(tx);
-        tempDb.setStateVersion(backupRoot, backupVersion);
-        this.node.stateManager.deleteVersion(tempVersion);
+      const res = this.executeOrRollbackTransaction(tempDb, tx, validTransactions, invalidTransactions);
+      if (!res) {
+        this.node.destroyDb(tempDb);
+        return null;
       }
     }
-    this.node.stateManager.deleteVersion(backupVersion);
 
     // Once successfully executed txs (when submitted to tx pool) can become invalid
     // after some blocks are created. Remove those transactions from tx pool.
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -277,14 +277,22 @@ class Consensus {
     let backupRoot = this.node.stateManager.cloneVersion(tempVersion, backupVersion);
     if (!backupRoot) {
       logger.error(`[${LOG_HEADER}] Failed to clone state version: ${tempVersion}`);
+      this.node.destroyDb(tempDb);
+      return null;
     }
     for (const voteTx of lastVotes) {
       const txRes = tempDb.executeTransaction(voteTx);
+      this.node.stateManager.deleteVersion(backupVersion);
       if (!ChainUtil.transactionFailed(txRes)) {
         logger.debug(`[${LOG_HEADER}] last vote: success`);
+        backupRoot = this.node.stateManager.cloneVersion(tempVersion, backupVersion);
+        if (!backupRoot) {
+          logger.error(`[${LOG_HEADER}] Failed to clone state version: ${tempVersion}`);
+          this.node.stateManager.deleteVersion(tempVersion);
+          return null;
+        }
       } else {
         logger.error(`[${LOG_HEADER}] last vote: failure\n ${JSON.stringify(txRes)}`);
-        this.node.stateManager.deleteVersion(backupVersion);
         this.node.stateManager.deleteVersion(tempVersion);
         return null;
       }
@@ -300,16 +308,18 @@ class Consensus {
       if (!ChainUtil.transactionFailed(txRes)) {
         logger.debug(`[${LOG_HEADER}] tx: success`);
         validTransactions.push(tx);
-      } else {
-        logger.error(`[${LOG_HEADER}] tx: failure\n ${JSON.stringify(txRes)}`);
-        invalidTransactions.push(tx);
-        tempDb.setStateVersion(backupRoot, backupVersion);
+        this.node.stateManager.deleteVersion(backupVersion);
         backupRoot = this.node.stateManager.cloneVersion(tempVersion, backupVersion);
         if (!backupRoot) {
           logger.error(`[${LOG_HEADER}] Failed to clone state version: ${tempVersion}`);
           this.node.stateManager.deleteVersion(tempVersion);
           return null;
         }
+      } else {
+        logger.error(`[${LOG_HEADER}] tx: failure\n ${JSON.stringify(txRes)}`);
+        invalidTransactions.push(tx);
+        tempDb.setStateVersion(backupRoot, backupVersion);
+        this.node.stateManager.deleteVersion(tempVersion);
       }
     }
     this.node.stateManager.deleteVersion(backupVersion);
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -262,7 +262,6 @@ class Consensus {
     const tempVersion = this.node.stateManager.createUniqueVersionName(
         `${StateVersions.CONSENSUS_CREATE}:${lastBlock.number}:${blockNumber}`);
     const tempDb = this.node.createTempDb(baseVersion, tempVersion, lastBlock.number - 1);
-    logger.debug(`[${LOG_HEADER}] Created a temp state for tx checks`);
     const lastBlockInfo = this.blockPool.hashToBlockInfo[lastBlock.hash];
     logger.debug(`[${LOG_HEADER}] lastBlockInfo: ${JSON.stringify(lastBlockInfo, null, 2)}`);
     // FIXME(minsu or lia): When I am behind and a newly coming node is ahead of me, then I cannot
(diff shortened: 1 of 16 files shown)


consensus/index.js
```diff
@@ -272,27 +271,49 @@ class Consensus {
...
-      if (!ChainUtil.transactionFailed(tempDb.executeTransaction(tx))) {
+      const txRes = tempDb.executeTransaction(tx);
+      if (!ChainUtil.transactionFailed(txRes)) {
         logger.debug(`[${LOG_HEADER}] tx: success`);
         validTransactions.push(tx);
       } else {
-        logger.debug(`[${LOG_HEADER}] tx: failure`);
+        logger.error(`[${LOG_HEADER}] tx: failure\n ${JSON.stringify(txRes)}`);
         invalidTransactions.push(tx);
+        this.node.stateManager.renameVersion(backupVersion, tempVersion);
```
platfowner, line 306: Could you try to use db.setStateVersion() instead? (ref: `executeOrRollbackTransaction()` in https://github.com/ainblockchain/ain-blockchain/blob/master/node/index.js#L289) 

What `renameVersion(fromVersion, toVersion)` actually does is to change the version names of all descendant nodes rooted at `toVersion` named `fromVersion` to `toVersion`. It's target use case is to call it just after cloning with another name (it's because the version name itself cannot easily changed in the current state manager module). Sorry it's confusing.

db/index.js
```diff
@@ -373,12 +374,42 @@ class DB {
     return resultList;
   }
 
+  getAccountsNonceAndTimestamp(address) {
```
platfowner, line 377: minor: What about s/Accounts/Account/ as it returns value of a single account?

db/index.js
```diff
@@ -373,12 +374,42 @@ class DB {
...
+      nonce = 0;
+    }
+    if (timestamp === null) {
+      timestamp = 0;
+    }
+    return { nonce, timestamp };
+  }
+
+  updateAccountNonceAndTimestamp(address, nonce, timestamp) {
+    if (nonce === undefined || timestamp === undefined) return;
```
platfowner, line 392: What about using ChainUtil.isNumber()?

db/index.js
```diff
@@ -393,6 +424,15 @@ class DB {
       // There is nothing to do.
       return true;
     }
+    if (!fromSetOp && transaction && auth && auth.addr && !auth.fid) {
```
platfowner, line 427: What about introducing a separate function for this condition?

db/index.js
```diff
@@ -409,6 +449,9 @@ class DB {
     }
     const valueCopy = ChainUtil.isDict(value) ? JSON.parse(JSON.stringify(value)) : value;
     this.writeDatabase(fullPath, valueCopy);
+    if (!fromSetOp && transaction && auth && auth.addr && !auth.fid) {
+      this.updateAccountNonceAndTimestamp(auth.addr, transaction.tx_body.nonce, transaction.tx_body.timestamp);
+    }
```
platfowner, line 454: Could you update the notes below for me?
I think the first one can be deleted and the second one needs to be updated as now we allow recursive function calls not allowing circular calls.

db/index.js
```diff
@@ -393,6 +424,15 @@ class DB {
       // There is nothing to do.
       return true;
     }
+    if (!fromSetOp && transaction && auth && auth.addr && !auth.fid) {
+      const { nonce, timestamp: accountTimestamp } = this.getAccountsNonceAndTimestamp(auth.addr);
+      if (transaction.tx_body.nonce >= 0 && transaction.tx_body.nonce !== nonce) {
+        return ChainUtil.returnError(105, `Invalid nonce: ${transaction.tx_body.nonce}`);
```
platfowner, line 430: What about renumbering to `103` updating `TX_NONCE_ERROR_CODES` as well as we're on a development stage yet?

db/index.js
platfowner, line 427: Suggestion: What about moving this block and the block of `updateAccountNonceAndTimestamp()` below to `executeTransaction()`? It looks like it can help make things simpler?

integration/blockchain.test.js
```diff
@@ -130,7 +130,7 @@ RANDOM_OPERATION = [
   ['batch', {tx_list: [{operation: {type: 'SET_VALUE', ref: 'test/builed/some/deep',
       value: {'place': {'next': 1, 'level': 'down'}}}}]}],
   ['batch', {tx_list: [{operation: {type: 'SET_VALUE', ref: 'test/b/u/i/l/e/d/hel',
-      value: {'range': [1, 4, 5], 'another': [234]}}}]}],
+      value: {'range': 1, 'another': 2}}}]}],
```
platfowner, line 133: nice catch!

integration/blockchain.test.js
```diff
@@ -196,14 +196,43 @@ function waitUntilNodeStakes() {
...
+          params: {
+            address,
+            from: 'pending',
+            protoVer: CURRENT_PROTOCOL_VERSION
+          }
+        }
+      }).body.toString('utf-8')).result.result;
+    if (randomOperation[0] === 'batch') {
+      for (const tx of value.tx_list) {
+        tx.nonce = nonce++;
```
platfowner, line 223: Isn't the nonce automatically set by `createSingleTransaction()` if it's not set here?

platfowner:
Could you revive this test case for me? https://github.com/ainblockchain/ain-blockchain/blob/develop/integration/node.test.js#L883 

It's be waiting for this moment.

integration/node.test.js
```diff
@@ -1573,31 +1582,35 @@ describe('Blockchain Node', () => {
                 ref: "test/test_value/some202/path",
                 value: "some other202 value",
               },
-              timestamp: Date.now()
+              timestamp: Date.now(),
+              nonce: nonce
```
platfowner, line 1586: Automatically set by `createSingleTransaction()` if the client API is used?

integration/node.test.js
```diff
@@ -1767,36 +1784,43 @@ describe('Blockchain Node', () => {
...
+        .then((nonceRes) => {
+          const nonce = _.get(nonceRes, 'result.result');
+          const txBody = {
+            operation: {
+              type: 'SET_VALUE',
+              value: 'some other value 2',
+              ref: `test/test_value/some/path`
+            },
+            timestamp: Date.now(),
+            nonce,  // strictly ordered nonce
```
platfowner, line 1803: Happy to see this!

node/index.js
```diff
@@ -339,9 +364,6 @@ class BlockchainNode {
   addNewBlock(block) {
     if (this.bc.addNewBlockToChain(block)) {
       this.tp.cleanUpForNewBlock(block);
-      const newVersion = `${StateVersions.NODE}:${block.number}`;
-      this.syncDb(newVersion);
```
platfowner, line 343: Question: Are there reasons they need to be moved?
Above two lines are for updating the block number part of the state version.

integration/blockchain.test.js
platfowner: Oh, they're not auto-incremental. Please ignore the above comment.

integration/node.test.js
platfowner: ditto

node/index.js
```diff
@@ -159,7 +162,9 @@ class BlockchainNode {
       }
     }
     const nodeVersion = `${StateVersions.NODE}:${blockNumber}`;
-    this.syncDb(nodeVersion)
+    this.syncDb(nodeVersion);
+    const newNonce = this.db.getAccountsNonceAndTimestamp(this.account.address).nonce;
+    this.nonce = newNonce;
```
platfowner, line 167: What about moving the above two lines into `syncDb()`?

node/index.js
```diff
@@ -241,13 +261,13 @@ class BlockchainNode {
     *                                        not
     * @return {Transaction} Instance of the transaction class
     */
-  createTransaction(txBody) {
+  createTransaction(txBody, baseDb) {
```
platfowner, line 264: It looks like we don't have any use case of `baseDb` yet? If so, what about just not defining it for now?

tx-pool/index.js
```diff
@@ -107,67 +99,97 @@ class TransactionPool {
           (a, b) => {
             return a.hash === b.hash;
           }
-      );
+        );
```
platfowner, line 102: minor: What about `});`?

tx-pool/index.js
```diff
@@ -107,67 +99,97 @@ class TransactionPool {
...
-            tx.tx_body.operation.op_list[0].ref : undefined;
-        const type = _.get(tx, 'tx_body.operation.type');
-        return (type !== WriteDbOperations.SET_VALUE && type !== WriteDbOperations.SET) ||
-            (ref && !ref.startsWith('/consensus/number')) ||
-            (innerRef && !innerRef.startsWith('/consensus/number'));
-      });
+          const type = _.get(tx, 'tx_body.operation.type');
+          if (type !== WriteDbOperations.SET_VALUE && type !== WriteDbOperations.SET) {
+            return true;
+          }
```
platfowner, line 107: Suggestion: What about defining the above lines as a separate function?

tx-pool/index.js
```diff
@@ -107,67 +99,97 @@ class TransactionPool {
...
+          const ref = _.get(tx, 'tx_body.operation.ref');
+          const innerRef = _.get(tx, 'tx_body.operation.op_list.0.ref');
+          return ((ref && !ref.startsWith('/consensus/number')) ||
+              (innerRef && !innerRef.startsWith('/consensus/number')));
```
platfowner, line 111: Suggestion: A separate function as well?

tx-pool/index.js
```diff
@@ -107,67 +99,97 @@ class TransactionPool {
...
-        } else if (listToTakeValue[0].tx_body.nonce < 0) {
-          newList.push(listToTakeValue.shift());
+      }
+      unvalidatedTransactions[addr] = newTxList;
+    }
+    // Merge list of transactions while ordering by timestamp. Initial ordering by nonce is preserved.
+    return this.mergeMultipleSortedArrays(Object.values(unvalidatedTransactions));
+  }
+
+  mergeMultipleSortedArrays(arrays) {
```
platfowner, line 154: What about adding some test cases for this new function?

tx-pool/index.js
```diff
@@ -107,67 +99,97 @@ class TransactionPool {
...
+          newArr.push(arrays[i]);
         }
       }
+      arrays = newArr;
+    }
+    return arrays.length === 1 ? arrays[0] : [];
+  }
 
-      orderedUnvalidatedTransactions.push(newList);
+  mergeTwoSortedArrays(arr1, arr2) {
```
platfowner, line 169: ditto?

tx-pool/index.js
```diff
@@ -107,67 +99,97 @@ class TransactionPool {
...
-        // Order by noncing if transactions are nonced, else by timestamp
-        unvalidatedTransactions[address].sort((a, b) =>
-            (a.tx_body.nonce < 0 || b.tx_body.nonce < 0) ?
-                ((a.tx_body.timestamp > b.tx_body.timestamp) ?
-                    1 : ((b.tx_body.timestamp > a.tx_body.timestamp) ? -1 : 0)) :
-                (a.tx_body.nonce > b.tx_body.nonce) ?
-                    1 : ((b.tx_body.nonce > a.tx_body.nonce) ? -1 : 0));
+        unvalidatedTransactions[address].sort((a, b) => {
+          if (a.tx_body.nonce === b.tx_body.nonce) {
+            if (a.tx_body.nonce >= -1) { // both ordered / unordered
```
platfowner, line 119: s/ordered/strictly ordered/?

tx-pool/index.js
platfowner: Or the whole block (8 lines) as a separate function?

tx-pool/index.js
```diff
@@ -107,67 +99,97 @@ class TransactionPool {
...
+      const newTxList = [];
+      for (let index = 0; index < txList.length; index++) {
+        const tx = txList[index];
+        const accountNonce = _.get(noncesAndTimestamps, `${addr}.nonce`, 0);
+        const accountTimestamp = _.get(noncesAndTimestamps, `${addr}.timestamp`, 0);
+        if ((tx.tx_body.nonce < 0 || tx.tx_body.nonce === accountNonce) &&
+            (tx.tx_body.nonce !== -2 || tx.tx_body.timestamp > accountTimestamp)) {
+          newTxList.push(tx);
+          if (!noncesAndTimestamps[addr]) noncesAndTimestamps[addr] = {};
+          if (tx.tx_body.nonce >= 0) noncesAndTimestamps[addr].nonce = tx.tx_body.nonce + 1;
```
platfowner, line 144: Consider using `ChainUtil.setJsObject()`.

tx-pool/index.js
```diff
@@ -107,67 +99,97 @@ class TransactionPool {
...
+    // Remove incorrectly nonced / timestamped transactions
+    const noncesAndTimestamps = baseStateVersion ?
+        this.node.getValueWithStateVersion(PredefinedDbPaths.ACCOUNTS, false, baseStateVersion) : {};
+    for (const [addr, txList] of Object.entries(unvalidatedTransactions)) {
+      const newTxList = [];
+      for (let index = 0; index < txList.length; index++) {
+        const tx = txList[index];
+        const accountNonce = _.get(noncesAndTimestamps, `${addr}.nonce`, 0);
+        const accountTimestamp = _.get(noncesAndTimestamps, `${addr}.timestamp`, 0);
+        if ((tx.tx_body.nonce < 0 || tx.tx_body.nonce === accountNonce) &&
```
platfowner, line 140: Could you consider extracting this condition as a separate function?
And as a less efficient but somewhat more readable statement? e.g.
```
return (nonce === -1 || (nonce === -2 && tx_body.timestamp > accountTimestamp) || tx_body.nonce === accountNonce);
```

tx-pool/index.js
```diff
@@ -107,67 +99,97 @@ class TransactionPool {
...
-                (a.tx_body.nonce > b.tx_body.nonce) ?
-                    1 : ((b.tx_body.nonce > a.tx_body.nonce) ? -1 : 0));
+        unvalidatedTransactions[address].sort((a, b) => {
+          if (a.tx_body.nonce === b.tx_body.nonce) {
+            if (a.tx_body.nonce >= -1) { // both ordered / unordered
+              return 0;
+            }
+            return a.tx_body.timestamp - b.tx_body.timestamp; // both loosely ordered
+          }
+          if (a.tx_body.nonce >= 0 && b.tx_body.nonce >= 0) { // both ordered
```
platfowner, line 124: s/ordered/strictly ordered/?

tx-pool/index.js
```diff
@@ -107,67 +99,97 @@ class TransactionPool {
...
       } else {
         unvalidatedTransactions[address] = tempFilteredTransactions;
-        // Order by noncing if transactions are nonced, else by timestamp
-        unvalidatedTransactions[address].sort((a, b) =>
-            (a.tx_body.nonce < 0 || b.tx_body.nonce < 0) ?
-                ((a.tx_body.timestamp > b.tx_body.timestamp) ?
-                    1 : ((b.tx_body.timestamp > a.tx_body.timestamp) ? -1 : 0)) :
-                (a.tx_body.nonce > b.tx_body.nonce) ?
-                    1 : ((b.tx_body.nonce > a.tx_body.nonce) ? -1 : 0));
+        unvalidatedTransactions[address].sort((a, b) => {
```
platfowner, line 117: Is it applicable to use simpler sorting logic? like
```
if (a.tx_body.nonce !== b.tx_body.nonce) {
  return a.tx_body.nonce - b.tx_body.nonce;
}
return a.tx_body.timestamp - b.tx_body.timestamp;
```

consensus/index.js
liayoo: No problem! Thanks for the explanation. Updated to use setStateVersion().

db/index.js

db/index.js

db/index.js
liayoo: Moved the checks to executeOperation(), which reduced code duplication, and thought this no longer needs to be a separate function. Please let me know if you think otherwise!

db/index.js

db/index.js
liayoo: The checks have been moved to executeOperation(), and I thought they would need a different numbering from other error codes in set functions. Went with 900 and 901.

db/index.js
liayoo: This really helped with code duplication & cleaning up the code. Thanks!!

liayoo:
> Could you revive this test case for me? https://github.com/ainblockchain/ain-blockchain/blob/develop/integration/node.test.js#L883
> 
> It's be waiting for this moment.

It is now uncommented! Yay!

node/index.js
liayoo: While running tests, I noticed that the state version list was growing with a lot of `NODE:${number}` versions. Thought this might be the reason why the list was growing (which was actually not..)
Anyways, I don't think this syncDb() is needed since cloneAndFinalizeVersion() is called right after addNewBlock() with the same block number?

node/index.js

node/index.js

tx-pool/index.js
liayoo: Done!

tx-pool/index.js
liayoo: Refactored code into excludeConsensusTransactions().

tx-pool/index.js
liayoo: Refactored code into excludeConsensusTransactions().

tx-pool/index.js
liayoo: Done!

tx-pool/index.js
liayoo: Done!

tx-pool/index.js
liayoo: Done!

tx-pool/index.js
liayoo: Done!

tx-pool/index.js
liayoo: Defined isCorrectlyNoncedOrTimestamped() for this condition checking logic. Thanks for the great suggestion!

tx-pool/index.js
liayoo: Done!

tx-pool/index.js
liayoo: I **really like** the simplicity, but that would mean loosely ordered txs have precedence over unordered txs, and loosely ordered / unorderd txs have precedence over every ordered tx. Also, strictly ordered / unordered txs can be ordered by their timestamps. 

--
I guess the last part (ordering by timestamp) is ok, since we're going to incorporate that into the gas fee logic. I'm not so sure about the first part. What do you think?

liayoo:

db/index.js
```diff
@@ -409,9 +440,7 @@ class DB {
     }
     const valueCopy = ChainUtil.isDict(value) ? JSON.parse(JSON.stringify(value)) : value;
     this.writeDatabase(fullPath, valueCopy);
-    // NOTE(seo): As of now (2021-01), we don't allow recursive function triggering.
-    // NOTE(lia): Allow recursive function triggering for service accounts. Should update this logic
-    // to prevent infinite recursion.
+    // NOTE(lia): Allow recursive function triggering but no circular calls.
```
platfowner, line 443: Ah, right. We do support chained but not recursive function calls yet.

node/index.js
```diff
@@ -214,6 +220,21 @@ class BlockchainNode {
     return nonce;
   }
 
+  getNonceForAddr(address, pending) {
+    if (!isCksumAddr(address)) return -1;
```
platfowner, line 224: Would it be helpful to support non-checksum address?

db/index.js
liayoo: Right, thanks! Updated the comment.

node/index.js
liayoo: Updated to accept non-checksummed but valid addresses as well!

liayoo:

node/index.js
platfowner: Just checked that syncDb() is called at the end of cloneAndFinalizeVersion().
Thanks for addressing the question.

node/index.js
platfowner: Thanks, then what about new naming like `syncDbAndNonce()`? (minor)

tx-pool/index.js
platfowner: Thanks!

tx-pool/index.js
platfowner: As of now, let's just keep the existing logic.
Regarding the Gas Fee Charging design, let's have further discussions. Thanks for addressing the question.

consensus/index.js
```diff
@@ -272,27 +271,48 @@ class Consensus {
...
-    })
+    }
 
-    const transactions = this.node.tp.getValidTransactions(longestNotarizedChain);
+    // TODO(lia): restrict the size / number of txs
+    const transactions = this.node.tp.getValidTransactions(longestNotarizedChain, tempVersion);
     const validTransactions = [];
     const invalidTransactions = [];
-    transactions.forEach((tx) => {
+    for (const tx of transactions) {
```
platfowner, line 297: Does the clone & deletion of the version need to take place for each transaction as it's done so in executeOrRollbackTransaction() (https://github.com/ainblockchain/ain-blockchain/blob/master/node/index.js#L289)? 

In this case, I think the unused one (either this.db.stateVersion or backupVersion) needs to be removed manually. (Maybe it's a bug that the deletion of this.db.stateVersion is missing in the case of rollback?: https://github.com/ainblockchain/ain-blockchain/blob/master/node/index.js#L308)

tx-pool/index.js
```diff
@@ -84,90 +77,139 @@ class TransactionPool {
...
-            (a.tx_body.nonce < 0 || b.tx_body.nonce < 0) ?
-                ((a.tx_body.timestamp > b.tx_body.timestamp) ?
-                    1 : ((b.tx_body.timestamp > a.tx_body.timestamp) ? -1 : 0)) :
-                (a.tx_body.nonce > b.tx_body.nonce) ?
-                    1 : ((b.tx_body.nonce > a.tx_body.nonce) ? -1 : 0));
+        addrToTxList[address] = filteredTransactions;
+        // sort transactions
+        addrToTxList[address].sort((a, b) => {
+          if (a.tx_body.nonce === b.tx_body.nonce) {
+            if (a.tx_body.nonce >= -1) { // both strictly ordered / unordered
```
platfowner, line 129: minor: would it be more deterministic if we apply sorting by timestamp for unordered txs as well by changing it to `a.tx_body.nonce >= 0`?

platfowner (commented):
Just left some comments.
- State rollback logic in block proposal
- Sort by timestamp for unordered txs?

Sorry, it takes time due to the code complexities.

consensus/index.js
liayoo: I was trying to avoid cloning too much if possible, but you're right. The code was not updating the backup as needed.. Updated to clone & delete versions for each tx.
Also, added the missing version deletion in executeOrRollbackTransaction().

tx-pool/index.js
liayoo: Done!

liayoo:
> Just left some comments.
> 
> * State rollback logic in block proposal
> * Sort by timestamp for unordered txs?
> 
> Sorry, it takes time due to the code complexities.

No worries! I mistakenly thought that the review round was finished, my bad. Thanks for taking the time to review, as always! 🙏

node/index.js
```diff
@@ -300,6 +323,9 @@ class BlockchainNode {
           logger.error(`[${LOG_HEADER}] Failed to finalize version: ${backupVersion}`);
         }
       }
+      if (!this.stateManager.deleteVersion(this.db.stateVersion)) {
```
platfowner, line 326: What about moving it just after setStateVersion() to avoid temporary status of db with deleted state version?

consensus/index.js
```diff
@@ -237,6 +237,52 @@ class Consensus {
     }
   }
 
+  executeLastVoteOrAbort(db, tx) {
+    const LOG_HEADER = 'executeLastVoteOrAbort';
+    const dbVersion = db.stateVersion;
+    const backupVersion = this.node.stateManager.createUniqueVersionName(
```
platfowner, line 243: Maybe backupVersion is not actually used so it can be just removed?

node/index.js
liayoo: Done!

consensus/index.js
liayoo: Thanks for catching that! Done.

platfowner (approved):
LGTM
Thanks for introducing such a nice solution.

liayoo:
Thanks for the reviews! Now merging..

Issue #283, opened 2021-03-24
[BUG] Broken integration tests


platfowner:
It seems some integration tests are failing after the nonce tracking PR #281 . @liayoo , could you take a look?

```
./node_modules/mocha/bin/mocha --timeout 160000 integration/node.test.js -g 'batch with a failing transaction'

  Blockchain Node
    APIs (sets)
      /batch
        1) batch with a failing transaction


  0 passing (18s)
  1 failing

  1) Blockchain Node
       APIs (sets)
         /batch
           batch with a failing transaction:

      AssertionError: expected [ Array(8) ] to deeply equal [ Array(8) ]
      + expected - actual

           }
           "tx_hash": "erased"
         }
         {
      -    "result": {
      -      "code": 11
      -      "error_message": "[executeOrRollbackTransaction] Failed to clone state version: BACKUP:4:1616557789709:0"
      -    }
      +    "result": true
           "tx_hash": "erased"
         }
         {
      -    "result": {
      -      "code": 11
      -      "error_message": "[executeOrRollbackTransaction] Failed to clone state version: BACKUP:4:1616557789709:0"
      -    }
      +    "result": true
           "tx_hash": "erased"
         }
         {
      -    "result": {
      -      "code": 11
      -      "error_message": "[executeOrRollbackTransaction] Failed to clone state version: BACKUP:4:1616557789709:0"
      -    }
      +    "result": true
           "tx_hash": "erased"
         }
         {
      -    "result": {
      -      "code": 11
      -      "error_message": "[executeOrRollbackTransaction] Failed to clone state version: BACKUP:4:1616557789709:0"
      -    }
      +    "result": true
           "tx_hash": "erased"
         }
       ]
      
      at Context.<anonymous> (integration/node.test.js:1700:16)
      at processImmediate (internal/timers.js:461:21)
```

```
./node_modules/mocha/bin/mocha --timeout 160000 integration/node.test.js -g 'owner only: set_function with ownerOnly = true'

  Blockchain Node
    Native functions
      Function permission (_saveLastTx, _transfer)
        Owner only
          ✓ owner only: set_function with ownerOnly = true (_transfer)
          1) "after each" hook for "owner only: set_function with ownerOnly = true (_transfer)"


  1 passing (16s)
  1 failing

  1) Blockchain Node
       Native functions
         "after each" hook for "owner only: set_function with ownerOnly = true (_transfer)":
     AssertionError: expected { Object (code, error_message) } to equal true
[...]

PR #285, opened 2021-03-24
Fix state version deletion bug

liayoo:
Fixes: https://github.com/ainblockchain/ain-blockchain/issues/283


platfowner (approved):
LGTM
Thanks!

PR #286, opened 2021-03-25
Feature/proof verification

minsulee2:


consensus/index.js
```diff
@@ -727,6 +730,20 @@ class Consensus {
     this.handleConsensusMessage({value: voteTx, type: ConsensusMessageTypes.VOTE});
   }
 
+  checkStateValidity(finalizableChain) {
+    for (const blockToFinalize of finalizableChain) {
+      if (blockToFinalize.number <= this.node.bc.lastBlockNumber()) {
+        continue;
+      }
+      const versionToFinalize = this.blockPool.hashToDb.get(blockToFinalize.hash).stateVersion;
+      const state = this.node.stateManager.getRoot(versionToFinalize);
```
platfowner, line 739: minor: what about s/state/root/?

db/state-util.js
```diff
@@ -307,6 +307,27 @@ function updateProofHashForAllRootPaths(fullPath, root) {
   return updateProofHashForAllRootPathsRecursive(node);
 }
 
+function verifyProofHashForAllRootPathsRecursive(node) {
+  let verification = true;
+  if (!node.getIsLeaf()) {
+    for (const childNode of node.getChildNodes()) {
+      const decendentVerifications = verifyProofHashForAllRootPathsRecursive(childNode);
```
platfowner, line 314: s/decendent/descendant/

db/state-util.js
```diff
@@ -307,6 +307,27 @@ function updateProofHashForAllRootPaths(fullPath, root) {
...
+      verification = verification && node.verifyProofHash() && decendentVerifications;
```
platfowner, line 315: It seems that early stop can be applied here?

db/state-util.js
```diff
@@ -307,6 +307,27 @@ function updateProofHashForAllRootPaths(fullPath, root) {
   return updateProofHashForAllRootPathsRecursive(node);
 }
 
+function verifyProofHashForAllRootPathsRecursive(node) {
```
platfowner, line 310: It looks like verifyProofHashForAllRootPaths() and verifyProofHashForAllRootPathsRecursive() can be combined into one as their arguments have no difference?

platfowner, line 310: 'ForAllRootPaths' means for the paths to all roots, but in this case we have only one root and the logic is recursively applied to all sub-states. So what about 'verifyProofHashForStateTree()' instead?

consensus/index.js
```diff
@@ -737,6 +754,11 @@ class Consensus {
       logger.debug(`[${LOG_HEADER}] No notarized chain with 3 consecutive epochs yet`);
       return;
     }
+    // XXX(minsu): neeeeeeeeed discussion how to deal with it!
```
platfowner, line 757: What about using TODO() pattern instead? cc: @liayoo

liayoo:
Actually, as I mentioned offline, state proofs are already being checked when the node first receives a block (done in checkProposal()). I believe just checking the root proof hashes should be sufficient since tx execution is performed by the node itself? (unless there's a bug in the proof hashing logic?) 
So I thought we needed proof hash checks for incoming chain segments, not in consensus... (here's a related TODO: https://github.com/ainblockchain/ain-blockchain/blob/develop/node/index.js#L442) Am I understanding this correctly? @minsulee2 @platfowner

consensus/index.js
```diff
@@ -737,6 +754,11 @@ class Consensus {
...
+    if (!this.checkStateValidity(finalizableChain)) {
```
platfowner, line 758: Regarding where to add the verification:

It looks like there are two entry points of new blocks:
- Entry1: MessageTypes.CONSENSUS -> handleConsensusMessage() -> ConsensusMessageTypes.PROPOSE -> checkProposal()

```
    if (!LIGHTWEIGHT) {
      if (newDb.getProof('/')[ProofProperties.PROOF_HASH] !== proposalBlock.state_proof_hash) {
        logger.error(`[${LOG_HEADER}] State proof hashes don't match: ` +
            `${newDb.getProof('/')[ProofProperties.PROOF_HASH]} / ` +
            `${proposalBlock.state_proof_hash}`);
        this.node.destroyDb(newDb);
        return false;
      }
    }
 ```

- Entry2: MessageTypes.CHAIN_SEGMENT_RESPONSE -> mergeChainSegment() -> applyBlocksToDb()

```
    for (const block of blockList) {
      // TODO(lia): validate the state proof of each block
      if (!db.executeTransactionList(block.last_votes)) {
        logger.error(`[${LOG_HEADER}] Failed to execute last_votes of block: ` +
            `${JSON.stringify(block, null, 2)}`);
        return false;
      }
      if (!db.executeTransactionList(block.transactions)) {
        logger.error(`[${LOG_HEADER}] Failed to execute transactions of block: ` +
            `${JSON.stringify(block, null, 2)}`);
        return false;
      }
    }
```

As you can see, we're already applying proof verification to entry1. So I think now we need to add proof verification to entry2. cc: @liayoo

File: consensus/index.js [hunk 7]
```diff
@@ -727,6 +730,20 @@ class Consensus {
...
   }
 
+  checkStateValidity(finalizableChain) {
+    for (const blockToFinalize of finalizableChain) {
+      if (blockToFinalize.number <= this.node.bc.lastBlockNumber()) {
+        continue;
+      }
+      const versionToFinalize = this.blockPool.hashToDb.get(blockToFinalize.hash).stateVersion;
+      const state = this.node.stateManager.getRoot(versionToFinalize);
+      if (!verifyProofHashForAllRootPaths(state)) {
```
platfowner, line 740: It's not clear to me why proof validation needs to be applied to a path?
The states are re-generated by the current blockchain node using the given blocks. So isn't it enough to compare the root proof hash and the proof hash in the block for each block?

platfowner:
@liayoo  Your comments are fully synced with mine.

consensus/index.js
minsulee2: Done!

db/state-util.js
minsulee2: Done!

db/state-util.js
minsulee2: Done!

consensus/index.js
minsulee2: Done!

db/state-util.js
minsulee2: Done!

consensus/index.js
minsulee2: I added the validity check at the entry2! I got confused because of the function name, which I got a hint as addNewblock() (at consensus.js). It was meant to addNewBlockToChain(). Very unfortunate :( No worries. Fixed'em all!
Thanks for the long explain!!

consensus/index.js
minsulee2: Right! left comment regarding this above. Updated! thanks!

minsulee2:
@platfowner @liayoo PTAL. I found myself why I've got a huge confusion. I guess I am on the right track as you guys as well :)

stop_servers.sh
```diff
@@ -1,6 +1,6 @@
 
 BASEDIR=$(dirname "$0")
-rm -rf $BASEDIR/blockchain/blockchains
+rm -rf $BASEDIR/blockchains
```
platfowner, line 3: It seems this PR needs to be synced.

db/state-util.js
```diff
@@ -307,6 +307,19 @@ function updateProofHashForAllRootPaths(fullPath, root) {
   return updateProofHashForAllRootPathsRecursive(node);
 }
 
+function verifyProofHashForStateTree(stateTree) {
+  let verification = true;
+  if (!stateTree.getIsLeaf()) {
+    for (const childNode of stateTree.getChildNodes()) {
+      const decendantVerifications = verifyProofHashForStateTree(childNode);
```
platfowner, line 314: minor: what about s/verification/verified/ and s/descendantVerifications/childVerified/?

db/state-util.js
```diff
@@ -307,6 +307,19 @@ function updateProofHashForAllRootPaths(fullPath, root) {
...
+      verification = verification && stateTree.verifyProofHash() && decendantVerifications;
+      return verification;
```
platfowner, line 316: Early stop needs to be applied only for only failure case?
```
function verifyProofHashForStateTree(node) {
  if (!node.verifyProofHash()) {
    return false;
  }
  if (!node.getIsLeaf()) {
    for (const childNode of node.getChildNodes()) {
      if (!verifyProofHashForStateTree(childNode)) {
        return false;
      }
    }
  }
  return true;
}
```

unittest/state-util.test.js
```diff
@@ -1365,4 +1366,47 @@ describe("state-util", () => {
...
+          another_route: {
+            test: -1000
+          }
+        }
+      };
+      const rootNode = StateNode.fromJsObject(jsObject);
+      const level0Node = rootNode.getChild('level0');
+      const level1Node = level0Node.getChild('level1');
+      setProofHashForStateTree(rootNode);
+      level1Node.setProofHash('0xdeadbeaf');
```
platfowner, line 1408: What about adding level12 node as a sibling of level1 node (renaming level 11 node?) and use level12 node for the failure case to check the early stop logic of verifyProofHashForStateTree()?

liayoo (approved):
LGTM, thanks! Please address @platfowner's comments before merging.

stop_servers.sh
platfowner: I mean synced with `git pull --rebase origin develop`.

minsulee2: Done!

db/state-util.js
minsulee2: As refactoring done, the variables does not exist anymore! Thanks!

db/state-util.js
minsulee2: Done!!

unittest/state-util.test.js
minsulee2: Done!!

minsulee2:
@platfowner PTAL. All addressed.

platfowner (approved):
LGTM

minsulee2:
Thanks thanks! merging it now!

Issue #299, opened 2021-04-05
Need to track critical resource metrics: state tree height 


platfowner:

platfowner:
Fixed.

PR #302, opened 2021-04-05
Add APIs for tree height

platfowner:
Summary:
- Add APIs `/get_state_info`, `ain_getStateInfo` for state info 
```
{
  tree_depth: <Tree depth>,
  tree_size: <Tree size>
}
``` 
- Rename state proof APIs: `/get_proof` -> `/get_state_proof`, `ain_getProof` -> `ain_getStateProof`
- Remove `GET_PROOF` operation as we have separate APIs 
- Tweak `getRefForWriting()` code

Testing:
- Passed unit tests and integration tests

Related issue:
- https://github.com/ainblockchain/ain-blockchain/issues/299


db/state-node.js
```diff
@@ -13,21 +13,23 @@ class StateNode {
     // Used for leaf nodes only.
     this.value = null;
     this.proofHash = null;
+    this.treeDepth = 1;
```
liayoo, line 16: I think a depth of a tree is usually defined as "the length of the path to its root" ([ref](https://en.wikipedia.org/wiki/Tree_(data_structure)#Terminology)). Can we rename this to subtreeDepth or something else if you have a better name?

platfowner: Just replaced its name with 'tree height' and updated related logic accordingly. PTAL @liayoo 
Thanks for the clarification.

liayoo (approved):
LGTM, thanks!

minsulee2:
LGTM tho, I've got a question that in what circumstance the state tree needs to have its height?

minsulee2 (approved):

platfowner:
@minsulee2 The tree height is updated whenever db write happens (writeDatabase()) and will be used for setting limits on critical resources.

platfowner:
Thanks for the reviews, now being merged..

Issue #311, opened 2021-04-06
Need to set limits on critical resources: tree size


platfowner:
Related issue: #303

cshcomcom:
Done! https://github.com/ainblockchain/ain-blockchain/pull/330

PR #315, opened 2021-04-09
Feature/tree size limit

cshcomcom:
# Description
- Transactions with greater than 20 tree height are blocked
- Change blockchain.MAX_TX_BYTES to resource.TX_BYTES_LIMIT

Resolve: #311

### DIFF
--- a/db/index.js
+++ b/db/index.js
@@ -858,10 +858,10 @@ class DB {
     // NOTE(seo): It's not allowed for users to send transactions with auth.fid.
     const executionResult = this.executeOperation(
         txBody.operation, { addr: tx.address }, txBody.timestamp, tx);
-    const executedStateInfo = this.getStateInfo();
-    if (executionResult && executedStateInfo[StateInfoProperties.TREE_HEIGHT] > TREE_HEIGHT_LIMIT) {
+    const stateInfo = this.getStateInfo('/');
+    if (executionResult && stateInfo[StateInfoProperties.TREE_HEIGHT] > TREE_HEIGHT_LIMIT) {
       return ChainUtil.returnError(910, `Out of tree height limit ` +
-          `(${executedStateInfo[StateInfoProperties.TREE_HEIGHT]} > ${TREE_HEIGHT_LIMIT})`);
+          `(${stateInfo[StateInfoProperties.TREE_HEIGHT]} > ${TREE_HEIGHT_LIMIT})`);
     }
     return executionResult;
   }
--- a/db/index.js
+++ b/db/index.js
@@ -859,7 +859,7 @@ class DB {
     const executionResult = this.executeOperation(
         txBody.operation, { addr: tx.address }, txBody.timestamp, tx);
     const stateInfo = this.getStateInfo('/');
-    if (executionResult && stateInfo[StateInfoProperties.TREE_HEIGHT] > TREE_HEIGHT_LIMIT) {
+    if (executionResult === true && stateInfo[StateInfoProperties.TREE_HEIGHT] > TREE_HEIGHT_LIMIT) {
       return ChainUtil.returnError(910, `Out of tree height limit ` +
           `(${stateInfo[StateInfoProperties.TREE_HEIGHT]} > ${TREE_HEIGHT_LIMIT})`);
     }
--- a/db/index.js
+++ b/db/index.js
@@ -859,9 +859,10 @@ class DB {
     const executionResult = this.executeOperation(
         txBody.operation, { addr: tx.address }, txBody.timestamp, tx);
     const stateInfo = this.getStateInfo('/');
-    if (executionResult === true && stateInfo[StateInfoProperties.TREE_HEIGHT] > TREE_HEIGHT_LIMIT) {
+    const treeHeight = stateInfo[StateInfoProperties.TREE_HEIGHT];
+    if (executionResult === true && treeHeight > TREE_HEIGHT_LIMIT) {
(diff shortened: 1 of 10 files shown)


db/index.js
```diff
@@ -855,8 +856,14 @@ class DB {
       return false;
     }
     // NOTE(seo): It's not allowed for users to send transactions with auth.fid.
-    return this.executeOperation(
+    const executedResult = this.executeOperation(
         txBody.operation, { addr: tx.address }, txBody.timestamp, tx);
+    const executedStateInfo = this.getStateInfo();
+    if (executedResult && executedStateInfo[StateInfoProperties.TREE_HEIGHT] > TREE_HEIGHT_LIMIT) {
+      return ChainUtil.returnError(910, `Out of tree height limit ` +
+          `(${executedStateInfo[StateInfoProperties.TREE_HEIGHT]} > ${TREE_HEIGHT_LIMIT})`);
```
platfowner, line 864: What about defining a variable for `executedStateInfo[StateInfoProperties.TREE_HEIGHT]`?

db/index.js
```diff
@@ -855,8 +856,14 @@ class DB {
       return false;
     }
     // NOTE(seo): It's not allowed for users to send transactions with auth.fid.
-    return this.executeOperation(
+    const executedResult = this.executeOperation(
```
platfowner, line 859: minor: executedResult sounds like 'the result was executed'. What about executionResult or execResult in short instead?

db/index.js
```diff
@@ -855,8 +856,14 @@ class DB {
...
         txBody.operation, { addr: tx.address }, txBody.timestamp, tx);
+    const executedStateInfo = this.getStateInfo();
```
platfowner, line 861: s/executedStateInfo/stateInfo/?

And getStateInfo() requires input parameter. Could you call `getStateInfo('/')` explicitly?

db/index.js
```diff
@@ -855,8 +856,14 @@ class DB {
...
+    if (executedResult && executedStateInfo[StateInfoProperties.TREE_HEIGHT] > TREE_HEIGHT_LIMIT) {
+      return ChainUtil.returnError(910, `Out of tree height limit ` +
```
platfowner, line 863: Could you add some test cases to `db.test.js`?

db/index.js
```diff
@@ -72,7 +73,7 @@ class DB {
 
   /**
    * Sets state version with its state root.
-   *
```
platfowner, line 75: I'd recommend to turn off the lint rule for these trailing spaces as it seems it doesn't help much for readability or productivity.

db/index.js
```diff
@@ -855,8 +856,14 @@ class DB {
       return false;
     }
     // NOTE(seo): It's not allowed for users to send transactions with auth.fid.
-    return this.executeOperation(
+    const executedResult = this.executeOperation(
         txBody.operation, { addr: tx.address }, txBody.timestamp, tx);
+    const executedStateInfo = this.getStateInfo();
+    if (executedResult && executedStateInfo[StateInfoProperties.TREE_HEIGHT] > TREE_HEIGHT_LIMIT) {
```
platfowner, line 862: The return value of executeOperation() is non-null even with error code, so it needs to explicitly compared with `true` e.g. `result === true`.

db/index.js
cshcomcom: OK. I'll change it to 'executionResult'!

db/index.js
cshcomcom: Looks good! Done

db/index.js
cshcomcom: I agree with your recommendation. But, It needs to set 'no-trailing-spaces' to off. Is it okay?

cshcomcom: I couldn't find an option to turn off comments only.

db/index.js
cshcomcom: Thanks! In this codes, even if an error is passed, it will pass. I'll change it

db/index.js
platfowner: I see. I think it's just fine to keep it if we don't have any nicer options.

db/index.js
cshcomcom: Done!

db/index.js
cshcomcom: Done!

db/index.js
cshcomcom: Okay!

cshcomcom:
@platfowner All comments are addressed! PTAL

unittest/db.test.js
```diff
@@ -3179,5 +3179,19 @@ describe("Transaction execution", () => {
...
+        operation: {
+          type: 'SET_VALUE',
+          ref: '/test/1/2/3/4/5/6/7/8/9/10/11/12/13/14/15/16/17/18/19/20',
+          value: 'some value',
+        },
+        nonce: -1,
+        timestamp: 1568798344000,
+      };
+      const overHeightTx = Transaction.fromTxBody(overHeightTxBody, node.account.private_key);
+      assert.equal(node.db.executeTransaction(overHeightTx).code, 910);
```
platfowner, line 3194: what about adding a success case with an input of height 19 to check that the limit is applied preciously?

cshcomcom: Looks good! Done

platfowner (approved):
LGTM
Thanks!

Could you merge this quickly so that https://github.com/ainblockchain/ain-blockchain/pull/317 is updated?

cshcomcom:
Thanks for reviews!!

Issue #316, opened 2021-04-12
Need to track critical resource metrics: gas amount


platfowner:

platfowner:
Duplicated by https://github.com/ainblockchain/ain-blockchain/issues/300

platfowner:
Fixed.

PR #317, opened 2021-04-12
Always return code of transaction execution

platfowner:
Summary:
- Always return code of transaction execution
- Rename transactionFailed() -> isFailedTx()
- Define path util as a class with static methods
- Fix a tx result path bug
- Migrate to deepEqual() from equal() as equal() is deprecated: https://nodejs.org/api/assert.html#assert_assert_equal_actual_expected_message 
- Use strict equal in reference comparison test cases
- Allow explicit turn-off by CONSOLE_LOG=false 

### DIFF
--- a/integration/afan_dapp.test.js
+++ b/integration/afan_dapp.test.js
@@ -80,7 +80,7 @@ function setUp() {
       nonce: -1,
     }
   }).body.toString('utf-8')).result;
-  assert.equal(_.get(res, 'result'), true);
+  assert.deepEqual(_.get(res, 'result.code'), 0);
   if (!waitUntilTxFinalized(serverList, res.tx_hash)) {
     console.log(`Failed to check finalization of setUp() tx.`)
   }
@@ -109,7 +109,7 @@ function cleanUp() {
       nonce: -1,
     }
   }).body.toString('utf-8')).result;
-  assert.equal(_.get(res, 'result'), true);
+  assert.deepEqual(_.get(res, 'result.code'), 0);
   if (!waitUntilTxFinalized(serverList, res.tx_hash)) {
     console.log(`Failed to check finalization of cleanUp() tx.`)
   }
--- a/blockchain/block.js
+++ b/blockchain/block.js
@@ -18,11 +18,7 @@ const {
   ProofProperties,
   StateVersions,
 } = require('../common/constants');
-const {
-  getTransferValuePath,
-  getCreateAppRecordPath,
-  getStakingStakeRecordValuePath,
-} = require('../common/path-util');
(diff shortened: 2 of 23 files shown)


integration/afan_dapp.test.js
```diff
@@ -80,7 +80,7 @@ function setUp() {
       nonce: -1,
     }
   }).body.toString('utf-8')).result;
-  assert.equal(_.get(res, 'result'), true);
+  assert.deepEqual(_.get(res, 'result.code'), 0);
```
cshcomcom, line 83: question: If it is not an object, the result will be the same. Is there a reason for changing it to deepEqual?

liayoo, line 83: minor) I think it's better to use .equal() (strict equality) than .deepEqual() for non-objects?

cshcomcom (approved):
Looks nice! I think It would be nice to change codes to constant later.

liayoo (commented):
Can we change [checkForTransactionErrorCode](https://github.com/ainblockchain/ain-blockchain/pull/317/files#diff-5119edef5eda0552aab3a253404274c2aae6f4772c2a1eb8d33f1b79c50b42f4R264)() in chain-util.js logic now?

liayoo (approved):
LGTM with some minor comments. Thanks!!

integration/afan_dapp.test.js
platfowner: Do you have any examples of the cases that strictEqual() catches but deepEqual() doesn't?

platfowner: It seems that equal() is deprecated so we need to choose either of strictEqual(), notStrictEqual(), or deepEqual() (or their analogous in expect() library e.g. to.equal(), to.deep.equal()):
https://nodejs.org/api/assert.html#assert_assert_equal_actual_expected_message
PTAL
I think most of the use cases we have are comparing values so I just tried to migrate them to deepEqual() except some reference comparison cases in db.test.js. 
(ref: https://stackoverflow.com/questions/63249848/in-chai-assertion-library-what-is-the-use-of-assert-equal-when-we-already-ha)

liayoo: Oh, I didn't know .equal() was deprecated! In that case, I think just using deepEqual is fine for our use cases. Thanks!

platfowner:
Thanks for the fast review. Let me merge it now..

integration/afan_dapp.test.js
cshcomcom: I didn't know that `equeal()` was deprecated! Thanks

Issue #327, opened 2021-04-14
Need some limit for batch tx size


platfowner:

platfowner:
Fixed.

PR #328, opened 2021-04-14
Set batch transaction list size limit

platfowner:
Summary:
- Set batch transaction list size limit
- Fix transaction bytes limit application 

Related issue: https://github.com/ainblockchain/ain-blockchain/issues/327

### DIFF
--- a/integration/node.test.js
+++ b/integration/node.test.js
@@ -14,11 +14,12 @@ const APP_SERVER = PROJECT_ROOT + "client/index.js"
 const {
   CURRENT_PROTOCOL_VERSION,
   CHAINS_DIR,
-  FunctionResultCode,
-  TX_BYTES_LIMIT,
-  GenesisAccounts,
   HASH_DELIMITER,
+  FunctionResultCode,
+  GenesisAccounts,
   ProofProperties,
+  TX_BYTES_LIMIT,
+  BATCH_TX_LIST_SIZE_LIMIT,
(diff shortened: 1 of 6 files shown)


integration/node.test.js
```diff
@@ -2412,70 +2412,143 @@ describe('Blockchain Node', () => {
...
+        })
+      })
+
+      it('rejects a batch transaction carrying an over-size transaction.', () => {
+        const account = ainUtil.createAccount();
+        const client = jayson.client.http(server1 + '/json-rpc');
+        let longText = '';
+        for (let i = 0; i < TX_BYTES_LIMIT / 2; i++) {
+          longText += 'a'
+        }
```
cshcomcom, line 2525: minor: How about `const longText = 'a'.repeat(TX_BYTES_LIMIT / 2);`?

cshcomcom (approved):
LGTM with a minor comment! Thanks

liayoo (commented):
Should we perform the same checks at /batch client api?

integration/node.test.js
platfowner: Done. Thanks.

platfowner:
@liayoo For the client APIs, I think it's fine as they're dev / debug APIs and will be disabled when we launch the mainnet.

platfowner:
Thanks for the fast review. Now merging..

PR #334, opened 2021-04-16
Move getAddress func to chain-util

minsulee2:
as the getAddress func is called in both transaction.js and p2p/index.js, the function is moved to chain-util.

### DIFF
--- a/p2p/util.js
+++ b/p2p/util.js
@@ -102,9 +102,9 @@ function signMessage(messageBody, privateKey) {
   return ainUtil.ecSignMessage(JSON.stringify(messageBody), Buffer.from(privateKey, 'hex'));
 }
 
-function getAddressFromSignature(message) {
+function getAddressFromMessage(message) {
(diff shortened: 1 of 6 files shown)


p2p/util.js
```diff
@@ -105,8 +104,7 @@ function signMessage(messageBody, privateKey) {
 
 function getAddressFromSignature(message) {
```
liayoo, line 105: How about we rename this to `getAddressFromMessage` and `getAddress` to `getAddressFromSignature` ?

platfowner: or `getAddressFromHash`?

minsulee2: I took @liayoo 's. It seems intuitive very much. Thanks!

minsulee2:
PTAL

liayoo:
LGTM, thanks!

liayoo (approved):

platfowner (approved):
LGTM
Thanks

minsulee2:
Thanks for the quick reviews! merging it now!

PR #335, opened 2021-04-16
Feature/gas amount gas price

liayoo:
DONE:
- Add `gas_price` field to Transaction
- Add `gas_amount_total` and `gas_cost_total` to Block
- Add `forceGasFeeWorkaround` to FeatureFlags & FORCE_GAS_FEE_WORKAROUND env var
- Allow gas_price = -1 for dev purposes (skip fee collection)
- Update escrow open related script & function names

TODO:
- Add _collectFee, _distributeFee
- Implement tx ordering policy


unittest/blockchain.test.js
```diff
@@ -57,11 +60,14 @@ describe('Blockchain', () => {
   });
 
   it('invalidates corrupt chain', () => {
-    const tx = getTransaction(node1, { operation: { type: 'SET_VALUE', ref: '/afan/test', value: 'foo'} });
+    const tx = getTransaction(node1, {
+        operation: { type: 'SET_VALUE', ref: '/afan/test', value: 'foo'},
```
platfowner, line 64: indentation?

common/chain-util.js
```diff
@@ -288,6 +288,10 @@ class ChainUtil {
     return _.get(result, 'gas.gas_amount', 0);
   }
 
+  static getTotalGasCost(gasPrice, result) {
+    return gasPrice * ChainUtil.getTotalGasAmount(result);
```
platfowner, line 292: As discussed offline, could you consider using a default unit of the gas price for user's convenience?
e.g. micro AIN (10^{-6} AIN) 

Actually, I'm trying to introduce account registration fee and its scale could be around 10^3 (in write operations). So if we apply micro AIN unit for example, `gas_price = 1` means gas cost of 10^{-3} AIN for a single account registration. Of course users can specify 0.1 or 0.01 if they want smaller values.
Does it make sense?

blockchain/block.js
```diff
@@ -22,7 +22,7 @@ const PathUtil = require('../common/path-util');
 
 class Block {
   constructor(lastHash, lastVotes, transactions, number, epoch, timestamp,
-      stateProofHash, proposer, validators) {
+      stateProofHash, proposer, validators, gasAmountTotal, gasCostTotal) {
```
minsulee2, line 25: Seems pipeline updates necessary :)

consensus/index.js
```diff
@@ -318,14 +318,20 @@ class Consensus {
...
     for (const tx of transactions) {
       const res = this.executeOrRollbackTransactionForBlock(
           tempDb, tx, validTransactions, invalidTransactions);
       if (!res) {
         this.node.destroyDb(tempDb);
         return null;
       }
+      resList.push(res);
     }
+    const gasAmountTotal = resList.reduce((acc, cur) => acc + ChainUtil.getTotalGasAmount(cur), 0);
```
minsulee2, line 331: I saw the same logic above. Can it be one of Chain-util functions?

blockchain/block.js
```diff
@@ -190,6 +196,7 @@ class Block {
     const firstTxBody = {
       nonce: -1,
       timestamp,
+      gas_price: 1,
```
minsulee2, line 199: Interesting! Is it fixed as 1?

liayoo: Not really, it would need to be changed according to the network's status for regular txs. For these genesis txs, we definitely have more discussion on what it should be. I just set it to 1 because it's simple.

+ Fyi, gas prices are going to be in a smaller denomination called "microain", which equals 10^-6 ain. On Ethereum, the gas price unit is gwei (= 10^-9 ether).

unittest/blockchain.test.js
liayoo: done!

common/chain-util.js
liayoo: Ah, right. Seems reasonable! 👍 Added microain & milliain units and made microain the default denomination for gas_price.

blockchain/block.js
liayoo: Yeahhh and ain-js as well! Thanks for catching that~

consensus/index.js
liayoo: done!

liayoo:
Addressed all comments, PTAL @platfowner @minsulee2

common/constants.js
```diff
@@ -68,6 +72,8 @@ const CHAINS_H2N_DIR_NAME = 'h2n'; // Note: Block hash to block number
 const HASH_DELIMITER = '#';
 const TX_NONCE_ERROR_CODE = 900;
 const TX_TIMESTAMP_ERROR_CODE = 901;
+const MICRO_AIN = 10**-3; // 1,000 microain = 1 ain
+const MILLI_AIN = 10**-6; // 1,000,000 milliain = 1 ain
```
platfowner, line 76: milli -> 1/1,000, micro -> 1/1,000,000?

common/chain-util.js
```diff
@@ -288,6 +288,25 @@ class ChainUtil {
     return _.get(result, 'gas.gas_amount', 0);
   }
 
+  /**
+   *
```
platfowner, line 292: Description? or redundant line?

common/constants.js
```diff
@@ -29,23 +29,27 @@ const FeatureFlags = {
...
 // Environment variables.
-const DEBUG = process.env.DEBUG ? process.env.DEBUG.toLowerCase().startsWith('t') : false;
+const DEBUG = process.env.DEBUG ? ChainUtil.convertEnvVarInputToBool(process.env.DEBUG) : false;
 const CONSOLE_LOG = FeatureFlags.forceConsoleLogging || (process.env.CONSOLE_LOG ?
-    process.env.CONSOLE_LOG.toLowerCase().startsWith('t') : false);
+    ChainUtil.convertEnvVarInputToBool(process.env.CONSOLE_LOG) : false);
 const ENABLE_DEV_CLIENT_API = FeatureFlags.forceDevClientApi || (process.env.ENABLE_DEV_CLIENT_API ?
-    process.env.ENABLE_DEV_CLIENT_API.toLowerCase().startsWith('t') : false);
+    ChainUtil.convertEnvVarInputToBool(process.env.ENABLE_DEV_CLIENT_API) : false);
+const FORCE_GAS_FEE_WORKAROUND = FeatureFlags.forceGasFeeWorkaround ||
```
platfowner, line 44: What about ENABLE_GAS_FEE_WORKAROUND? The background here could be basically the env variable turns on/off the feature and the feature flag forces (or overrides) the value?

platfowner: If you think it could cause unnecessary confusion, I think we can just use 'enable' everywhere. What do you think?
In the case, could you update forceDevClientApi flag as well?

platfowner (approved):
LGTM with some minor comments
Thanks!

common/constants.js
liayoo: oops....... fixed. thanks for catching my mistake!

common/chain-util.js
liayoo: description added!

common/constants.js
liayoo: aha, got it! changed the env var to ENABLE_GAS_FEE_WORKAROUND

liayoo:
Thanks for the review! 🙌  Now merging..

PR #341, opened 2021-04-19
Remove redundant code

platfowner:


liayoo:
Could you help me understand why this is redundant and similar code in [functions.test.js](https://github.com/ainblockchain/ain-blockchain/blob/develop/unittest/functions.test.js#L106) is not..?

platfowner:
@liayoo Ah, this mocking code catches network requests in unit tests but in integration tests it cannot as the servers run as standalone jobs.

minsulee2 (approved):
LGTM! Good to see tidy up!!

liayoo (approved):
LGTM. Thanks for the explanation!

PR #343, opened 2021-04-20
Version up

minsulee2:


client/protocol_versions.json
```diff
@@ -19,5 +19,8 @@
   },
   "0.6.0": {
     "min": "0.5.0"
+  },
+  "0.7.0": {
+    "min": "0.7.0"
```
liayoo, line 24: should we set the max of 0.5.x and 0.6.0 to 0.6.0 then? cc: @platfowner

minsulee2: Oh, Good point. I am also waiting for the answer :)

platfowner: Yes, I think so. Thanks!

minsulee2:
PTAL. All addressed!

platfowner (approved):
LGTM
Thanks!

minsulee2:
Thanks thanks!

liayoo (approved):
LGTM, thanks!

PR #351, opened 2021-04-21
Add nonce / gas_price validity checking logic

platfowner:
Summary:
- Move default gas_price setting logic from client to node (createSingleTransaction())
- Add nonce / gas_price validity checking logic to isValidTxBody() and test cases
- Always use env variable instead of feature flags (remove feature flags) 
- Simplify gas price handling logic 


tx-pool/transaction.js
```diff
@@ -186,12 +187,31 @@ class Transaction {
...
 
   static hasRequiredFields(txBody) {
-    return txBody && txBody.timestamp !== undefined && txBody.nonce !== undefined &&
-        txBody.operation !== undefined;
+    return txBody && txBody.operation !== undefined && txBody.timestamp !== undefined &&
+        txBody.nonce !== undefined;
+  }
+
+  static isValidNonce(nonce) {
+    return ChainUtil.isInteger(nonce) && (nonce >= 0 || nonce === -1 || nonce === -2);
```
liayoo, line 209: question) Is this chosen over `ChainUtil.isInteger(nonce) && nonce >= -2` for readability?

liayoo (approved):
LGTM with a question. Thanks for the nice refactoring! 👍

tx-pool/transaction.js
platfowner: That looks better. Just applied. Thanks.

platfowner:
Thanks for the fast review!

Issue #348, opened 2021-04-21
[BUG] Node 4 doesn't work


cshcomcom:
**To Reproduce**
1. Run 'sh start_servers.sh'
2. Node 4 does nothing!

**Screenshots**
![image](https://user-images.githubusercontent.com/74223287/115501051-76235100-a2ad-11eb-8176-0e698a5f998c.png)

**Additional context**
There seems to be something wrong with 'start_servers.sh' script being changed!
There was no problem before #333

minsulee2:
@cshcomcom closing it?

platfowner:
@minsulee2 Could you create a quick PR to alter the limit values to MIN_NUM_VALIDATORS - 1?

PR #352, opened 2021-04-21
Set max in and out bound - 1

minsulee2:


common/constants.js
```diff
@@ -467,7 +467,7 @@ function initializeNetworkEnvronments() {
   } else {
     return {
       // Note(minsu): Need a discussion that MIN_NUM_VALIDATORS and MAX_INBOUND_LIMIT should not be
-      // related one another.
+      // related to one another.
       MAX_OUTBOUND_LIMIT: GenesisParams.consensus.MIN_NUM_VALIDATORS - 1,
       MAX_INBOUND_LIMIT: GenesisParams.consensus.MIN_NUM_VALIDATORS - 1,
```
minsulee2, line 472: Sorry for confusing. I directly pushed to develop with -1 parts by mistake.  I made another branch and PR for this issue #348

platfowner: I think that shouldn't be possible under the current branch configuration(i.e., requiring at least 1 approval)?:
https://github.com/ainblockchain/ain-blockchain/settings/branches

minsulee2: I thought so too. But it happened (I still don't get it too). https://github.com/ainblockchain/ain-blockchain/blob/develop/common/constants.js#L464

platfowner: Just confirmed that our branch config is incomplete. I'll take a look. :)

platfowner (approved):
LGTM
Thanks!

common/constants.js
minsulee2: Thanks! I am glad that found a security hole by accident :)

PR #357, opened 2021-04-25
Feature/service app gas amount

liayoo:
Change summary:
- Separate gas amounts, costs for service & app ops
- Separate bandwidth & state gas amounts
- Support multi-app txs

### DIFF
--- a/common/chain-util.js
+++ b/common/chain-util.js
@@ -377,7 +377,7 @@ class ChainUtil {
   }
 
   static getServiceGasCostTotalFromTxList(txList, resList) {
-    const gasAmountTotal = resList.reduce((acc, cur) => acc + _.get(cur, 'service.bandwidth', 0), 0);
+    const gasAmountTotal = resList.reduce((acc, cur) => acc + ChainUtil.getTotalGasAmount(cur), 0);
     const gasCostTotal = resList.reduce((acc, cur, index) => {
       return acc + ChainUtil.getTotalGasCost(txList[index].tx_body.gas_price, cur);
     }, 0);
--- a/common/chain-util.js
+++ b/common/chain-util.js
@@ -297,9 +297,9 @@ class ChainUtil {
     return code !== 0;
   }
 
-  // TODO(lia): fix testing paths (writing at the root) and update isAppTx().
   static isAppTx(parsedPath) {
-    return _.get(parsedPath, 0) === 'apps' || !ChainUtil.isServiceTx(parsedPath);
+    const { PredefinedDbPaths } = require('../common/constants');
+    return _.get(parsedPath, 0) === PredefinedDbPaths.APPS;
   }
 
   // TODO(lia): fix testing paths (writing at the root) and update isServiceTx().
@@ -313,7 +313,7 @@ class ChainUtil {
     if (ChainUtil.isServiceTx(parsedPath)) {
       ChainUtil.setJsObject(gasAmount, ['service', 'bandwidth'], bandwidth);
       ChainUtil.setJsObject(gasAmount, ['service', 'state'], state);
-    } else {
+    } else if (ChainUtil.isAppTx(parsedPath)) {
       const appName = _.get(parsedPath, 1);
       if (!appName) return;
       ChainUtil.setJsObject(gasAmount, ['app', appName, 'bandwidth'], bandwidth);
--- a/common/chain-util.js
+++ b/common/chain-util.js
@@ -331,25 +331,10 @@ class ChainUtil {
   }
 
   static getSingleOpTotalGasAmount(result) {
-    const gasAmount = {
-      service: 0,
-      app: {}
-    };
     if (!result) {
-      return gasAmount;
+      return 0;
     }
-    const serviceBandwidth = _.get(result, 'gas.gas_amount.service.bandwidth', 0);
-    const serviceState = _.get(result, 'gas.gas_amount.service.state', 0);
-    ChainUtil.setJsObject(gasAmount, ['service'], serviceBandwidth + serviceState);
-    const apps = _.get(result, 'gas.gas_amount.app');
-    if (apps) {
-      for (const [appName, val] of Object.entries(apps)) {
-        const appBandwidth = _.get(val, 'bandwidth', 0);
-        const appState = _.get(val, 'state', 0);
-        ChainUtil.setJsObject(gasAmount, ['app', appName], appBandwidth + appState);
-      }
-    }
-    return gasAmount;
+    return _.get(result, 'gas.gas_amount.service.bandwidth', 0);
   }
 
   /**
@@ -357,20 +342,17 @@ class ChainUtil {
    */
   static getTotalGasAmount(result) {
     if (Array.isArray(result)) {
-      const gasAmount = {
-        service: 0,
-        app: {}
-      };
+      let gasAmount = 0;
       for (const elem of result) {
-        ChainUtil.mergeGasAmounts(gasAmount, ChainUtil.getSingleOpTotalGasAmount(elem));
+        gasAmount += ChainUtil.getSingleOpTotalGasAmount(elem);
       }
       return gasAmount;
     }
     return ChainUtil.getSingleOpTotalGasAmount(result);
   }
-
   /**
    * Calculate the gas cost (unit = ain).
+   * Only the service bandwidth gas amount is counted toward gas cost.
    * 
    * @param {Number} gasPrice gas price in microain
    * @param {Object} result transaction execution result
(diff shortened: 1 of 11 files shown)


common/chain-util.js
```diff
@@ -324,13 +377,22 @@ class ChainUtil {
...
+    };
+    for (const [appName, appGasAmount] of Object.entries(gasAmount.app)) {
+      ChainUtil.setJsObject(gasCost, ['app', appName], appGasAmount * gasPriceAIN);
+    }
+    return gasCost;
   }
 
-  static getGasAmountCostTotalFromTxList(txList, resList) {
-    const gasAmountTotal = resList.reduce((acc, cur) => acc + ChainUtil.getTotalGasAmount(cur), 0);
+  static getServiceGasAmountCostTotalFromTxList(txList, resList) {
```
platfowner, line 392: s/GasAmountCost/GasCost/?

common/chain-util.js
```diff
@@ -324,13 +377,22 @@ class ChainUtil {
...
     if (gasPrice === undefined) gasPrice = 0; // Default gas price = 0 microain
-    return gasPrice * MICRO_AIN * ChainUtil.getTotalGasAmount(result);
+    const gasAmount = ChainUtil.getTotalGasAmount(result);
+    const gasPriceAIN = gasPrice * MICRO_AIN;
+    const gasCost = {
+      service: gasAmount.service * gasPriceAIN,
+      app: {}
+    };
+    for (const [appName, appGasAmount] of Object.entries(gasAmount.app)) {
+      ChainUtil.setJsObject(gasCost, ['app', appName], appGasAmount * gasPriceAIN);
```
platfowner, line 387: minor: `ChainUtil.setJsObject(gasCost.app, [appName], appGasAmount * gasPriceAIN);`? or just
`gasCost.app[appName] = appGasAmount * gasPriceAIN;`?

common/chain-util.js
```diff
@@ -297,21 +297,74 @@ class ChainUtil {
...
+    const gasAmount = {
+      service: 0,
+      app: {}
+    };
+    if (!result) {
+      return gasAmount;
+    }
+    const serviceBandwidth = _.get(result, 'gas.gas_amount.service.bandwidth', 0);
+    const serviceState = _.get(result, 'gas.gas_amount.service.state', 0);
+    ChainUtil.setJsObject(gasAmount, ['service'], serviceBandwidth + serviceState);
```
platfowner, line 341: As discussed offline, bandwidth only would be enough for gas cost calculation by our design (the design doc needs to be updated).

common/chain-util.js
```diff
@@ -297,21 +297,74 @@ class ChainUtil {
     return code !== 0;
   }
 
+  static isAppTx(parsedPath) {
+    return _.get(parsedPath, 0) === 'apps' || !ChainUtil.isServiceTx(parsedPath);
+  }
+
+  static isServiceTx(parsedPath) {
+    const { NATIVE_SERVICE_TYPES } = require('../common/constants');
+    return NATIVE_SERVICE_TYPES.includes(_.get(parsedPath, 0));
```
platfowner, line 306: Question: Do we need opt-in based classification? Do we have any corner case that opt-out based classification (i.e. `!isAppTx()`) is not enough?

common/chain-util.js
```diff
@@ -297,21 +297,74 @@ class ChainUtil {
     return code !== 0;
   }
 
+  static isAppTx(parsedPath) {
+    return _.get(parsedPath, 0) === 'apps' || !ChainUtil.isServiceTx(parsedPath);
```
platfowner, line 301: Question: Do we need the latter case? Isn't  only`_.get(parsedPath, 0) === 'apps'` enough?

common/chain-util.js
platfowner: Ah, some genesis txs such as those on `/token` would be the case?

common/chain-util.js
```diff
@@ -297,21 +297,74 @@ class ChainUtil {
...
+  static isServiceTx(parsedPath) {
+    const { NATIVE_SERVICE_TYPES } = require('../common/constants');
+    return NATIVE_SERVICE_TYPES.includes(_.get(parsedPath, 0));
+  }
+
+  static getGasAmountObj(parsedPath, bandwidth, state) {
+    const gasAmount = {};
+    if (ChainUtil.isServiceTx(parsedPath)) {
+      ChainUtil.setJsObject(gasAmount, ['service', 'bandwidth'], bandwidth);
+      ChainUtil.setJsObject(gasAmount, ['service', 'state'], state);
```
platfowner, line 313: minor: `ChainUtil.setJsObject(gasAmount, ['service'], { bandwidth, state });` or
`gasAmount['service'] = { bandwidth, state };`
would be simpler? (it's up to you :) )

platfowner: please ignore the previous comment. the current code just looks good,

common/chain-util.js
platfowner: ditto

db/index.js
```diff
@@ -869,13 +874,14 @@ class DB {
...
       if (gasPrice > 0 && blockNumber > 0) {
         const gasCost = ChainUtil.getTotalGasCost(gasPrice, result);
-        // TODO(lia): update this to billing accounts for app transactions
-        const gasFeeCollectPath = PathUtil.getGasFeeCollectPath(auth.addr, blockNumber, tx.hash);
-        const gasFeeCollectRes = this.setValue(
-            gasFeeCollectPath, { amount: gasCost }, auth, timestamp, tx, false);
-        if (ChainUtil.isFailedTx(gasFeeCollectRes)) {
-          return ChainUtil.returnTxResult(
-              16, `Failed to collect gas fee: ${JSON.stringify(gasFeeCollectRes, null, 2)}`);
+        if (gasCost.service > 0) {
```
platfowner, line 877: What about adding the charged gas cost to the result?

common/chain-util.js
platfowner: So, I think this util function can provide separate values for bandwidth and state.

common/chain-util.js
```diff
@@ -297,21 +297,74 @@ class ChainUtil {
...
+    }
+    const serviceBandwidth = _.get(result, 'gas.gas_amount.service.bandwidth', 0);
+    const serviceState = _.get(result, 'gas.gas_amount.service.state', 0);
+    ChainUtil.setJsObject(gasAmount, ['service'], serviceBandwidth + serviceState);
+    const apps = _.get(result, 'gas.gas_amount.app');
+    if (apps) {
+      for (const [appName, val] of Object.entries(apps)) {
+        const appBandwidth = _.get(val, 'bandwidth', 0);
+        const appState = _.get(val, 'state', 0);
+        ChainUtil.setJsObject(gasAmount, ['app', appName], appBandwidth + appState);
```
platfowner, line 347: ditto

common/chain-util.js
liayoo: Right now, fees are [not collected](https://github.com/ainblockchain/ain-blockchain/blob/develop/db/index.js#L870) from the genesis txs. This is actually for tests (many of our tests allow writing at the root, e.g. /test/...). Can I make another PR for revamping the test paths & isAppTx / isServiceTx logic?

platfowner: Ok. Could you leave a TODO for that?

common/chain-util.js
```diff
@@ -297,21 +297,74 @@ class ChainUtil {
...
+      if (!appName) return;
+      ChainUtil.setJsObject(gasAmount, ['app', appName, 'bandwidth'], bandwidth);
+      ChainUtil.setJsObject(gasAmount, ['app', appName, 'state'], state);
+    }
+    return gasAmount;
+  }
+
+  static mergeGasAmounts(gasAmountObj1, gasAmountObj2) {
+    _.mergeWith(gasAmountObj1, gasAmountObj2, (a, b) => {
+      if (ChainUtil.isNumber(a) && ChainUtil.isNumber(b)) {
```
platfowner, line 327: What about `return ChainUtil.numberOrZero(a) + ChainUtil.numberOrZero(b);` as undefined result is interpreted as overriding? 
https://lodash.com/docs/4.17.15#mergeWith

common/chain-util.js
liayoo: Added a todo (as discussed below)

common/chain-util.js
liayoo: Thanks, added a todo.

common/chain-util.js
liayoo: done!

common/chain-util.js
liayoo: Thanks for catching that! Updated to include only the service bandwidth to gas cost calculation.

db/index.js
liayoo: Great idea, done!

common/chain-util.js
platfowner: As discussed offline, it seems it needs to return `undefined` for objects. So,
```
if (!ChainUtil.isDic(a)) {
  return ChainUtil.numberOrZero(a) + ChainUtil.numberOrZero(b);
}
```
seems to work.

liayoo: Thanks for the suggestion! Changed to check isDict() and use numberOrZero().

common/chain-util.js
liayoo: Is this ditto regarding the ChainUtil.setJsObject()?

liayoo:

common/chain-util.js
```diff
@@ -297,25 +297,62 @@ class ChainUtil {
...
+    const { NATIVE_SERVICE_TYPES } = require('../common/constants');
+    return NATIVE_SERVICE_TYPES.includes(_.get(parsedPath, 0));
+  }
+
+  static getGasAmountObj(parsedPath, bandwidth, state) {
+    const gasAmount = {};
+    if (ChainUtil.isServiceTx(parsedPath)) {
+      ChainUtil.setJsObject(gasAmount, ['service', 'bandwidth'], bandwidth);
+      ChainUtil.setJsObject(gasAmount, ['service', 'state'], state);
+    } else {
```
platfowner, line 316: What about applying to only app txs (and ignoring other txs explicitly)?
```
} else if (ChainUtil.isAppTx(paredPath)) {
...
}
```

common/chain-util.js
platfowner: Ah, it was about separating bandwidth and state values. Looks like it's done already.

common/chain-util.js
platfowner: As discussed offline, please check if tests pass with out `!ChainUtil.isServiceTx()`.

platfowner (approved):
LGTM with some minor comments (isAppTx() and getGasAmountObj()). 
Thanks!

common/chain-util.js
liayoo: 👍 done!

liayoo:
Thanks for the thorough review! Checked that all tests are passing & now merging..

PR #359, opened 2021-04-26
Feature/message timestamp

minsulee2:
Changes:
- Add timestamp in p2p messages
- Check whether timestamp is stale

### DIFF
--- a/p2p/index.js
+++ b/p2p/index.js
@@ -26,6 +26,7 @@ const {
   getAddressFromMessage,
   verifySignedMessage,
   checkProtoVer,
+  checkTimestamp,
   closeSocketSafe,
   encapsulateMessage
 } = require('./util');
@@ -316,6 +317,11 @@ class P2pClient {
         return;
       }
       if (!this.checkDataProtoVer(socket, parsedMessage.dataProtoVer)) {
+        return;
+      }
+      if (!checkTimestamp(_.get(parsedMessage, 'timestamp'))) {
+        logger.error(`The message from the node(${address}) is stale. Discard the message.`);
+        logger.debug(`The detail is as follows: ${parsedMessage}`);
         return;
       }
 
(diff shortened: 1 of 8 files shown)


p2p/util.js
```diff
@@ -148,11 +149,26 @@ function encapsulateMessage(type, dataObj) {
...
   };
   return message;
 }
 
+function checkTimestamp(timestamp) {
+  if (!timestamp) {
+    return false;
+  } else {
+    const now = Date.now();
+    const tenMinutesAsMilliseconds = moment.duration(10, 'minutes').as('milliseconds');
```
platfowner, line 163: What about defining this as a global constant (e.g. in 'network' section of the genesis params)? 
With more general name using upper snake case? (e.g. P2P_MESSAGE_TIMEOUT_MS)

p2p/index.js
```diff
@@ -318,6 +319,11 @@ class P2pClient {
       if (!this.checkDataProtoVer(socket, parsedMessage.dataProtoVer)) {
         return;
       }
+      if (!checkTimestamp(_.get(parsedMessage, 'timestamp'))) {
+        logger.error(`The message from the node(${address}) is stale. Discard the message.`);
```
liayoo, line 323: Could you prefix this log with LOG_HEADER & do the same to other logs as well? They all look the same, so.. I think it would help with debugging :)

liayoo (approved):

p2p/index.js
minsulee2: I've forgot this. Added all for setPeerEventHandler :)

p2p/util.js
minsulee2: I took P2P_MESSAGE_TIMEOUT_MS for the global var! Thanks! Done!

minsulee2:
@platfowner PTAL. All addressed!

platfowner (approved):
LGTM
Thanks!

minsulee2:
Thanks! merging it now!

PR #362, opened 2021-04-27
Do not return state gas amount.

platfowner:
Change summary:
- Do not return state gas amount
- Remove dup elements from NATIVE_SERVICE_TYPES

Background:
- We found some inconsistency in the value esp. in tests e.g. different results from:
  `yarn run test_node` and `yarn run test_node -g 'REST function with external RPC call'`
  due to common database state path between e.g. `/test/test_function_triggering/rest_function_path` and 
  `/test/test_function_triggering/allowed_path/value`
- So we decide to remove this feature as currently it doesn't have any use case except debugging

### DIFF
--- a/common/chain-util.js
+++ b/common/chain-util.js
@@ -316,17 +316,15 @@ class ChainUtil {
     return NATIVE_SERVICE_TYPES.includes(_.get(parsedPath, 0));
   }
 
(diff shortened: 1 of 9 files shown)


liayoo (approved):
LGTM, thanks!

Issue #265, opened 2021-03-09
[BUG] Need to be able to revert changes made in native functions


liayoo:
Currently, native function failures do not bubble up, which means state changes made in native functions are not reverted when function execution fails. 

Related TODO: https://github.com/ainblockchain/ain-blockchain/pull/262/files#diff-c581d39cccd0f924d890ba5d28f2c97e4a5d5385cb1c7d3c0a27187d8b2c94b4R609

platfowner:
Fixed.

PR #374, opened 2021-05-05
Update isFailedTx() to check function results as well

platfowner:
Change summary:
- Update isFailedTx() to check function results as well
- Include REST function id in func_results
- Fix a bug in _updateLatestShardReport() result
- Don't use ExecResultProperties anymore
- Let ENABLE_GAS_FEE_WORKAROUND default to true
- Always set stdioInherit = true, CONSOLE_LOG = false in integration tests for tester's convenience

Related issue: https://github.com/ainblockchain/ain-blockchain/issues/265

### DIFF
--- a/common/chain-util.js
+++ b/common/chain-util.js
@@ -289,24 +289,64 @@ class ChainUtil {
   /**
    * Returns true if the given result is from failed transaction or transaction list.
    */
-  // TODO(seo): Check the function results as well.
   static isFailedTx(result) {
     if (!result) {
       return true;
     }
-    if (result.result_list && ChainUtil.isArray(result.result_list)) {
-      for (const elem of result.result_list) {
-        if (ChainUtil.isFailedTxResultCode(elem.code)) {
+    if (ChainUtil.isArray(result.result_list)) {
+      for (const subResult of result.result_list) {
+        if (ChainUtil.isFailedTxResultCode(subResult.code)) {
           return true;
+        }
+        if (subResult.func_results) {
+          if (ChainUtil.isFailedFuncTrigger(subResult.func_results)) {
+            return true;
+          }
         }
       }
       return false;
     }
-    return ChainUtil.isFailedTxResultCode(result.code);
+    if (ChainUtil.isFailedTxResultCode(result.code)) {
+      return true;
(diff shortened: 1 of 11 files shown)


common/chain-util.js
```diff
@@ -396,13 +436,12 @@ class ChainUtil {
   }
 
   static returnTxResult(code, message = null, gasAmount = 0, funcResults = null) {
-    const { ExecResultProperties } = require('../common/constants');
     const result = {};
     if (message) {
       result.error_message = message;
     }
     if (!ChainUtil.isEmpty(funcResults)) {
-      result[ExecResultProperties.FUNC_RESULTS] = funcResults;
```
minsulee2, line 405: Has ExecResultProperties.FUNC_RESULTS removed?

unittest/chain-util.test.js
```diff
@@ -452,13 +452,99 @@ describe("ChainUtil", () => {
             "gas_amount": 0,
           }
         },
-        "code": 201,
+        "code": 201,  // The root operation failed
```
minsulee2, line 455: Are these codes stuff defined somewhere else?

platfowner: At this stage where code changes fast, we don't have separate definition except the corresponding code (code search with the error code or the error message). But, someday we have stabilized code, yes I think we can have separate definition (e.g. in constants.js).

common/chain-util.js
platfowner: Yes, it's gone for convenience for now as we have many property names.

minsulee2: Oh, I see. Thanks for the exaplain!

unittest/chain-util.test.js
minsulee2: I actually asked this question because of the code: 0 is success. It seems to me a bit bizarre cuz it is usually thought 1 is ok and 0 must be somewhat problematic. Hope Some codes are fixed quick :) Thanks for the exaplain!!

minsulee2 (approved):
LGTM!

liayoo (approved):
LGTM, thanks!

PR #376, opened 2021-05-06
Feature/compatible protocol versions

minsulee2:
To check compatibility just before sending message via a socket, both outbound and inbound now have node infos with socket.

- Update README.md
- Restruct outbound and inbound
- Get back to data version 1.0.0

### DIFF
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -249,7 +249,7 @@ class Consensus {
         // can keep sending messages with higher numbers, making the node's status unsynced, and
         // prevent the node from getting/handling messages properly.
         // this.node.state = BlockchainNodeStates.SYNCING;
-        Object.values(this.server.client.outbound).forEach(node=> {
+        Object.values(this.server.client.outbound).forEach(node => {
           this.server.client.requestChainSegment(node.socket, this.node.bc.lastBlock());
         });
         return;
--- a/consensus/index.js
+++ b/consensus/index.js
@@ -249,8 +249,8 @@ class Consensus {
         // can keep sending messages with higher numbers, making the node's status unsynced, and
         // prevent the node from getting/handling messages properly.
         // this.node.state = BlockchainNodeStates.SYNCING;
-        Object.values(this.server.client.outbound).forEach((socket) => {
-          this.server.client.requestChainSegment(socket, this.node.bc.lastBlock());
+        Object.values(this.server.client.outbound).forEach(node=> {
+          this.server.client.requestChainSegment(node.socket, this.node.bc.lastBlock());
         });
         return;
       }
--- a/p2p/index.js
+++ b/p2p/index.js
@@ -26,7 +26,6 @@ const {
   signMessage,
   getAddressFromMessage,
   verifySignedMessage,
-  checkProtoVer,
   isValidDataProtoVer,
   checkTimestamp,
   closeSocketSafe,
@@ -289,10 +288,6 @@ class P2pClient {
     const LOG_HEADER = 'setPeerEventHandlers';
     socket.on('message', (message) => {
       const parsedMessage = JSON.parse(message);
-      if (!checkProtoVer(this.outbound, socket,
-          this.server.minProtocolVersion, this.server.maxProtocolVersion, parsedMessage.protoVer)) {
-        return;
-      }
       const dataProtoVer = _.get(parsedMessage, 'dataProtoVer');
       if (!isValidDataProtoVer(dataProtoVer)) {
         const address = getAddressFromSocket(socket);
--- a/p2p/index.js
+++ b/p2p/index.js
@@ -448,8 +448,7 @@ class P2pClient {
       const address = getAddressFromSocket(this.outbound, socket);
       if (address) {
         logger.info(`with (${address}).`);
-      }
-      else {
+      } else {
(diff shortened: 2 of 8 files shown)


consensus/index.js
```diff
@@ -249,8 +249,8 @@ class Consensus {
         // can keep sending messages with higher numbers, making the node's status unsynced, and
         // prevent the node from getting/handling messages properly.
         // this.node.state = BlockchainNodeStates.SYNCING;
-        Object.values(this.server.client.outbound).forEach((socket) => {
-          this.server.client.requestChainSegment(socket, this.node.bc.lastBlock());
+        Object.values(this.server.client.outbound).forEach(node=> {
```
platfowner, line 252: minor: space?

p2p/index.js
```diff
@@ -454,6 +441,21 @@ class P2pClient {
...
 
+  waitForAddress = (socket) => {
+    return new Promise(resolve => {
+      setTimeout(resolve, WAIT_FOR_ADDRESS_TIMEOUT_MS);
+    }).then(() => {
+      const address = getAddressFromSocket(this.outbound, socket);
+      if (address) {
+        logger.info(`with (${address}).`);
+      }
+      else {
```
platfowner, line 452: `} else {`?

p2p/index.js
```diff
@@ -454,6 +441,21 @@ class P2pClient {
     });
   }
 
+  waitForAddress = (socket) => {
+    return new Promise(resolve => {
+      setTimeout(resolve, WAIT_FOR_ADDRESS_TIMEOUT_MS);
```
platfowner, line 446: minor: Can ChainUtil.sleep() be used instead?

p2p/index.js
```diff
@@ -252,31 +254,6 @@ class P2pClient {
     socket.send(JSON.stringify(payload));
   }
 
-  checkDataProtoVer(socket, version) {
```
platfowner, line 255: Question: Is the version checking logic here being abandoned?

p2p/index.js
```diff
@@ -355,7 +336,13 @@ class P2pClient {
               return;
             }
             logger.info(`[${LOG_HEADER}] A new websocket(${address}) is established.`);
-            this.outbound[address] = socket;
+            this.outbound[address] = {
+              socket: socket,
+              versions: {
+                API_VERSION: _.get(parsedMessage, 'protoVer'),
```
platfowner, line 342: Question: Do we need to check protocol version in p2p communication? cc: @liayoo

minsulee2: @platfowner @liayoo I actually wonder about it too. It's been the proto version checks in the setPeerEventhandler for a long time though, I don't also see any needs of checking this in p2p network. What about just checking DATA_PROTOCOL_VERSION here?

p2p/index.js
minsulee2: Nah, It is integrated into the function in the ./p2p/util.js

p2p/index.js
minsulee2: Done!

p2p/index.js
minsulee2: Done!

consensus/index.js
minsulee2: Done!

p2p/index.js
minsulee2: Done!!

p2p/index.js
platfowner: I'm asking becuase checkDataProtoVer() in util.js has only version validity checking logic?

p2p/index.js
platfowner: I think isolating data protocol version and consensus protocol version from other versions has some advantages:
- Logical simplicity
- Allow p2p communication between different package versions

and I see no special drawbacks. Let's just check data proto version in checkDataProtoVer() and consensus proto version in checkConsensusProtoVer()?

p2p/index.js
minsulee2: Ahhhh! I get it. As All checks work individually depending on message types in switch cases, the checkDataProtoVer is no longer check versions actually. So that, it now simply does check validity. What about renaming the checkDataProtoVer into isValidVersion or something?

p2p/index.js
minsulee2: Do you mean all three of API_VERSION, DATA_PROTOCOL_VERSION and CONSENSUS_PROTOCOL_VERSION right?

p2p/index.js
platfowner: The renaming idea looks good to me, thanks.

p2p/index.js
platfowner: @minsulee2 I mean we can skip checking api version in p2p message handling, so we can skip saving API_VERSION?

minsulee2: Oh, I got confused bcuz of mentioning consensus above. I got this. I am on the right track now!

minsulee2:
PTAL! All addressed!

p2p/index.js
minsulee2: Renamed it!

platfowner (approved):
LGTM
Thanks!

minsulee2:
Thanks for the review!

PR #391, opened 2021-05-13
Get cpu usages

minsulee2:
Add cpu usage using os package.

### DIFF
--- a/p2p/server.js
+++ b/p2p/server.js
@@ -185,21 +185,22 @@ class P2pServer {
   }
 
   getCpuUsage() {
-    const cpuUsage = {
-      free: 0,
-      usage: 0,
-      total: 0
-    }
     const cores = os.cpus();
+    let free = 0;
+    let total = 0;
     for (const core of cores) {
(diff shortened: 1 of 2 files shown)


cshcomcom:
Thanks! LGTM

cshcomcom (approved):
Thanks! LGTM

liayoo (approved):
LGTM, thanks!

p2p/server.js
```diff
@@ -184,6 +184,24 @@ class P2pServer {
     }
   }
 
+  getCpuUsage() {
+    const cpuUsage = {
+      free: 0,
+      usage: 0,
```
platfowner, line 190: minor: what about not setting this value as it will be overwritten below anyway?

platfowner (approved):
Thanks!

p2p/server.js
minsulee2: Done!!

minsulee2:
Thanks for the fast reviews!! merging now :)

PR #393, opened 2021-05-16
Fix integration test

minsulee2:
Updates here:
- Remove sleep package
- Use Chainutil.sleep() instead
- Have tested in act environment!


.github/workflows/local/github-actions-for-local.yml
```diff
@@ -0,0 +1,63 @@
...
+          registry-url: 'https://registry.npmjs.org'
+      - name: Install Yarn
+        run: sudo npm install -g yarn
+      - name: install pip
+        run: |
+          sudo apt-get install software-properties-common
+          sudo apt-add-repository universe
+          sudo apt-get update
+          sudo apt-get -y install python-pip
+          pip install ast
```
liayoo, line 33: question) why do we need pip and ast?

unittest/test-util.js
```diff
@@ -81,39 +80,39 @@ function waitUntilTxFinalized(servers, txHash) {
...
           .body
           .toString('utf-8')).result;
       if (txStatus && txStatus.is_finalized === true) {
         unchecked.delete(server);
       }
-    });
-    sleep(200);
-    iterCount++;
+      await ChainUtil.sleep(200);
+      iterCount++;
```
platfowner, line 91: Need to be moved out of this block?

unittest/test-util.js
```diff
@@ -81,39 +80,39 @@ function waitUntilTxFinalized(servers, txHash) {
...
       const txStatus = parseOrLog(syncRequest('GET', server + `/get_transaction?hash=${txHash}`)
           .body
           .toString('utf-8')).result;
       if (txStatus && txStatus.is_finalized === true) {
         unchecked.delete(server);
       }
-    });
-    sleep(200);
-    iterCount++;
+      await ChainUtil.sleep(200);
```
platfowner, line 90: Question: Is this change necessary for fixing the test?
Just curious as sleep() is left as-is in the previous refactoring since it makes difference only in performance, which is not that critical in tests?

.github/workflows/local/github-actions-for-local.yml
minsulee2: In the act environment, it is not exactly the same as the github actions environment. Since the setup-node@v2 image in act app haven't got ast module in python pip by default, the yarn install cannot be run. To fix this, pip install ast has been added!

unittest/test-util.js
minsulee2: The sleep package does not work in linux environment at all. So I removed them all in the test codes.

unittest/test-util.js
minsulee2: Can you explain it more?

unittest/test-util.js
platfowner: I see. Thanks for the explanation.

platfowner: As you can see, the two lines used to be out of the forEach() block and for the outer while loop.

unittest/test-util.js
platfowner: The two lines (sleep() + iterCount++) used to be out of the forEach() loop (now just for-loop) sleeping for and counting the iteration of the outer while-loop.

minsulee2: Oh, I removed .forEach() cuz the await does't work in the .forEach(). Now the two lines are in for() loop. Please check the code above :)

platfowner: I mean.. sleep() for example doesn't need to be called for each iteration of the inner loop (the server checks are independent of each other). It was intended to have some waiting time for the outer while-loop (not for inner loop). iterCount was also intended to count the iteration of the while-loop (not inner loop).

minsulee2: My apology! I got confused the lines.. Fixed!!

platfowner (approved):
LGTM

.github/workflows/local/github-actions-for-local.yml
liayoo: 👌 Thanks for the explanation!

liayoo:
LGTM, thanks!

liayoo (approved):

minsulee2:
Thanks for the reviews!

PR #397, opened 2021-05-20
Fix resList including invalid tx results.

liayoo:
Changed to not include a failed transaction's result in the result list.


platfowner (approved):
LGTM
Thanks!

liayoo:

PR #401, opened 2021-05-20
Pose limits on the tx pool size and the per-account tx pool size.

platfowner:
Change summary:
- Pose limits on the tx pool size and the per-account tx pool size


liayoo (approved):
LGTM, thanks! No more bloated tx pool! 🙌

integration/node.test.js
```diff
@@ -2346,6 +2349,77 @@ describe('Blockchain Node', () => {
...
+            signature,
+          });
+        }
+        const res1 = await client.request('ain_sendSignedTransactionBatch', {
+          tx_list: txList1,
+          protoVer: CURRENT_PROTOCOL_VERSION
+        });
+        const resultList1 = _.get(res1, 'result.result', null);
+        // Accepts transactions.
+        expect(Array.isArray(resultList1)).to.equal(true);
```
minsulee2, line 2385: Q) We have got Chainutil.isArray() though, Do we use Array.isArray() for the test case only?

minsulee2 (approved):
LGTM with a question!

integration/node.test.js
platfowner: I think it's totally fine in tests (as they're the same internally) but tried to update them. Thanks!

platfowner:
Thanks for the fast reviews! Now merging..

Issue #398, opened 2021-05-20
[BUG] Bandwidth limit broken


cshcomcom:
## Situation
- Set 'genesis_params.json' to `"BANDWIDTH_BUDGET_PER_BLOCK": 100, "SERVICE_BANDWIDTH_BUDGET_PER_BLOCK": 50`
- Run ain-benchmark with `"duration": 600, "numberOfTransactions": 60000`
- I ran test with 100 tps(1 block: 300 txs), but it passed normally

cshcomcom:
Fixed! Thanks @liayoo

PR #404, opened 2021-05-20
Bugfix/set gas extra field for zero cost tx

liayoo:
- Background: The 'gas' extra field was only set when gas cost > 0, which made transactions with 0 gas price bypass bandwidth limits. 
- Updated to set gas field even when gas cost = 0
- Updated broken tests
- Fixes https://github.com/ainblockchain/ain-blockchain/issues/398

### DIFF
--- a/integration/afan_dapp.test.js
+++ b/integration/afan_dapp.test.js
@@ -53,7 +53,7 @@ function startServer(application, serverName, envVars, stdioInherit = false) {
   });
 }
 
-function setUp() {
+async function setUp() {
   const addr = parseOrLog(syncRequest(
       'GET', server1 + '/get_address').body.toString('utf-8')).result;
   const createAppRes = parseOrLog(syncRequest('POST', server1 + '/set', {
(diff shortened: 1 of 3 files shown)


minsulee2 (approved):
LGTM!

platfowner (approved):
Thanks!

integration/afan_dapp.test.js
```diff
@@ -53,8 +53,32 @@ function startServer(application, serverName, envVars, stdioInherit = false) {
...
+          value: 1
+        }
+      ]
+    }
+  }).body.toString('utf-8')).result;
+  assert.deepEqual(ChainUtil.isFailedTx(_.get(createAppRes, 'result')), false);
+  if (!(await waitUntilTxFinalized(serverList, createAppRes.tx_hash))) {
+    console.log(`setUp(): Failed to check finalization of create app tx.`)
+  }
+  const legacySetupRes = parseOrLog(syncRequest('POST', server1 + '/set', {
```
platfowner, line 81: The name seems to hint some future work?
If so, what about adding a todo?

liayoo: Done, thanks!

liayoo:
Thanks for the reviews! Now merging..

Issue #406, opened 2021-05-21
Need to allow function IDs in owner config.


platfowner:

platfowner:
Fixed.

PR #407, opened 2021-05-21
Allow function IDs (auth.fid) in owner config

platfowner:
Change summary:
- Allow function IDs (auth.fid) in owner config
- Add _setOwnerConfig native function for testing
- Rename execGasAmount -> extraGasAmount

Related issue: https://github.com/ainblockchain/ain-blockchain/issues/406

### DIFF
--- a/common/constants.js
+++ b/common/constants.js
@@ -237,6 +237,7 @@ const AccountProperties = {
 const OwnerProperties = {
   ANYONE: '*',
   BRANCH_OWNER: 'branch_owner',
+  FID_PREFIX: 'fid:',
   OWNER: '.owner',
   OWNERS: 'owners',
   WRITE_FUNCTION: 'write_function',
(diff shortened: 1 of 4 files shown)


liayoo:
This made me realize that we currently do not have a check for owner values (e.g. check if the object's key is either an address or a string starting with 'fid:')! Should we maybe add a todo for that?

liayoo (approved):
LGTM with a comment. Thanks!

platfowner:
> This made me realize that we currently do not have a check for owner values (e.g. check if the object's key is either an address or a string starting with 'fid:')! Should we maybe add a todo for that?

Actually, we have a todo for the owner config value sanitization: https://github.com/ainblockchain/ain-blockchain/blob/master/db/index.js#L749 

I think the issue can be covered when we add the sanitization logic.
Thanks!

platfowner:
Thanks for the review. Now being merged..

PR #416, opened 2021-05-29
Bugfix for p2p test bug

minsulee2:
Fix p2p unittest is not correctly finishing. The problem was not wsServer but intervalConnection for tracker is multiple set and the after() in p2p.test.js is not working in a way of test intention. That is, the after() is called before setInterval() is called that makes interval is running after test finished. To fix this P2P_UNITTEST_FLAG is added.

Updates:
- Add P2P_UNITTEST_FLAG on package.json
- Do not make multiple intervals
- Remove unnecessary process.kill()

Special thanks for @minho-comcom-ai

### DIFF
--- a/p2p/index.js
+++ b/p2p/index.js
@@ -48,7 +48,7 @@ class P2pClient {
 
   run() {
     this.server.listen();
-    if (!process.env.P2P_UNITTEST_FLAG) this.connectToTracker();
+    this.connectToTracker();
   }
 
   // NOTE(minsulee2): The total number of connection is up to more than 5 without limit.
--- a/p2p/index.js
+++ b/p2p/index.js
@@ -48,7 +48,7 @@ class P2pClient {
 
   run() {
     this.server.listen();
-    this.setIntervalForTrackerConnection();
+    if (!process.env.P2P_UNITTEST_FLAG) this.connectToTracker();
   }
 
   // NOTE(minsulee2): The total number of connection is up to more than 5 without limit.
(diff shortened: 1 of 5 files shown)


p2p/index.js
```diff
@@ -48,7 +48,7 @@ class P2pClient {
 
   run() {
     this.server.listen();
-    this.setIntervalForTrackerConnection();
+    if (!process.env.P2P_UNITTEST_FLAG) this.connectToTracker();
```
platfowner, line 51: Could we avoid making changes on the original business logic to be tested for making tests run successfully? It seems that we can consider some alternatives like below?
1. Changing the testing environment to an integration test environment like other test cases under `integration/`.
2. Add some methods for extra access to the original server logic from the test code (e.g. stopTrackerConnectionIntervalForTesting()) and give the control to the test code (ref: setAccountForTesting(), setOwnersForTesting()).

minsulee2: Oh, right. I thought the solution was not that good approach either. With some code reviewing again, I found myself that it doesn't need to be called p2pClient.run() here. The p2pClient.run() calls the tracker stuff(we don't need here for the test) and I solved this problem simply by not calling it but p2pServer.listen() instead.
I probably have a difficulty in this due to tiredness of hard-working recently :( simple problem hard approach... Thanks for the review! I got this right back.

minsulee2: But, with some experiments, I found some bizarre issues such that if I, on the one hand, run the code via running node ./client then it cannot open the socket for tracker. On the other hands, if I run the code in mocha test suite, it IS ABLE TO open the socket. I presume the mocha makes some difference(I need to read some docs for this.)

minsulee2:
PTAL! I simply solve this problem! and now actions unittest passed as well.

platfowner (approved):
LGTM
Thanks!

liayoo:
"gets initial memory usage (it depends on the machine)" test case breaks on my MacBook.. Is this expected as the test name suggests?

minsulee2:
@liayoo I will investigate it before merging.

platfowner:
> @liayoo I will investigate it before merging.

It seems it happens before this PR. fyi

minsulee2:
> > @liayoo I will investigate it before merging.
> 
> It seems it happens before this PR. fyi

@platfowner @liayoo #419 opened! I will update it in the next PR. Thanks!

Issue #241, opened 2021-02-23
Service/app name should be restricted to alphabetic characters


platfowner:
Related issues: #217

platfowner:
'|' in service accounts, ':' in escrow service.
So, restrictions on service name e.g. alphabetic + numeric + '_'?

platfowner:
Fixed.

PR #433, opened 2021-06-07
Add service/app name validation and service type validation logic

platfowner:
Change summary:
- Add service/app name validation logic
- Add service type validation logic

Related topics: https://github.com/ainblockchain/ain-blockchain/issues/241


unittest/state-util.test.js
```diff
@@ -296,6 +297,69 @@ describe("state-util", () => {
     })
   })
 
+  describe("isValidServiceName", () => {
+    it("when non-string input", () => {
+      expect(isValidServiceName(null)).to.equal(false);
```
minsulee2, line 302: I have studied chai a bit recently and found that expect().to.be.false can also be good way to check false :)

minsulee2 (approved):
LGTM! Really good to see neat codesssss!!

common/constants.js
```diff
@@ -325,6 +325,34 @@ const NativeFunctionIds = {
...
+  const fidList = Object.values(NativeFunctionIds);
+  return fidList.find((elem) => elem === fid) !== undefined;
+}
+
+/**
+ * Types of service.
+ *
+ * @enum {string}
+ */
+const ServiceTypes = {
```
liayoo, line 341: I believe this is a subset of service types, so how about renaming it to something like `AccountAssociatedServiceTypes` or `ServiceAccountTypes`?

platfowner: Actually it was intended to include all service types. Please let me know if there are anything missing.

db/state-util.js
```diff
@@ -88,6 +89,11 @@ function isWritablePathWithSharding(fullPath, root) {
   return {isValid, invalidPath: isValid ? '' : ChainUtil.formatPath(path)};
 }
 
+function hasVarNamePattern(name) {
```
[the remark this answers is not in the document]
platfowner: Maybe some prefix can be added for such use cases?
Basically I'm open to relieve the restrictions once we see strong use cases. Until then how about keeping it a little bit conservative as variable-like service/app names are easy to identify and read?

unittest/state-util.test.js
platfowner: Yes, I saw it on your previous PRs. Thanks for the tip. I'll try to use it in new code. :)

common/constants.js
liayoo: I see, maybe I have a different definition of a service type.. I thought that service types could be everything in [NATIVE_SERVICE_TYPES](https://github.com/ainblockchain/ain-blockchain/pull/433/files#L79). (Paths in the root other than the apps).

db/state-util.js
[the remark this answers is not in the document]
liayoo: True, we don't have such a use case as of now. Relaxing the rules once we have a clear use case sounds good to me! 👍

common/constants.js
platfowner: Ah, I missed NATIVE_SERVICE_TYPES. Let me try to combine them.

platfowner: Just defined the list as SERVICE_ACCOUNT_SERVICE_TYPES.
One thing I found in adding test cases is that direct 'transfer' to service accounts is currently allowed for all valid service types (see the test case 'transfer: transfer with valid service account service type' in integration/node.test.js).
In order to avoid direct transfers to those we don't have use cases yet (e.g. /service_accounts/manage_app/...), the list was defined as a separately from SERVICE_TYPES. Let's think about better solutions.

platfowner:
All comments addressed. PTAL @liayoo @minsulee2

common/constants.js
```diff
@@ -1,18 +1,17 @@
-const os = require('os');
 const fs = require('fs');
 const path = require('path');
 const semver = require('semver');
 const ChainUtil = require('./chain-util');
 
-// Genesis configs.
+// Genesis configs **
```
liayoo, line 6: Just curious, what does `**` at the end mean?

common/constants.js
```diff
@@ -455,6 +447,44 @@ const GasFeeConstants = {
   REST_FUNCTION_CALL_GAS_AMOUNT: 10,
 };
 
+// Lists **
+
+/**
+ * Root labels of service paths.
+ */
+const SERVICE_TYPES = [
+  PredefinedDbPaths.ACCOUNTS,
```
liayoo, line 456: Thanks for organizing them! Looking much better 🙌

liayoo (approved):
LGTM, thanks!

common/constants.js
platfowner: Ah, it's a kind of delimiter for constant variable groups in constants.js. Just tried to make it more obvious.

platfowner:

PR #434, opened 2021-06-08
Feature/fast sync mode

liayoo:
- Added `SYNC_MODE= full | fast`
  - `SYNC_MODE=fast` means the node will try to sync from the existing snapshot first, check the block files, and then get any other missing blocks from the peers
  - `SYNC_MODE=full` is the default mode, and it's the way the nodes have been syncing until now. They sync with the block files and the peers
  - Nodes store snapshots at `/<DATA_DIR>/snapshots/n2s `every 1000 blocks, removing older snapshots
- Added proof hash checks in `executeChainOnDb()`
- Changed to request chain segments with a block number rather than a block
- Replaced `blockchain/block-file-util.js` with `common/file-util.js`
- Removed some unused code & packages


p2p/index.js
```diff
@@ -214,9 +213,11 @@ class P2pClient {
     logger.debug(`SENDING: ${JSON.stringify(consensusMessage)}`);
   }
 
-  requestChainSegment(socket, lastBlock) {
-    const payload = encapsulateMessage(MessageTypes.CHAIN_SEGMENT_REQUEST,
-        { lastBlock: lastBlock });
+  requestChainSegment(socket, lastBlockNumber) {
+    if (this.server.node.state === BlockchainNodeStates.STARTING) {
```
platfowner, line 217: minor: what about inclusive expression to help when the logic below this if-block should be applied? e.g.
`if (this.server.node.state !== BlockchainNodeStates.SYNCING && this.server.node.state !== BlockchainNodeStates.SERVING) {`

p2p/index.js
```diff
@@ -342,6 +343,11 @@ class P2pClient {
           }
           break;
         case MessageTypes.CHAIN_SEGMENT_RESPONSE:
+          if (this.server.node.state === BlockchainNodeStates.STARTING) {
```
platfowner, line 346: minor: ditto?

node/index.js
```diff
@@ -51,6 +59,8 @@ class BlockchainNode {
     this.db = this.createDb(StateVersions.EMPTY, initialVersion, this.bc, this.tp, false, true);
     this.nonce = null;  // nonce from current final version
     this.state = BlockchainNodeStates.STARTING;
+    this.snapshotPath = path.resolve(SNAPSHOTS_DIR, `${PORT}`);
```
platfowner, line 62: minor: What about SNAPSHOTS_DIR -> SNAPSHOTS_ROOT_DIR, snapshotPath -> snapshotDir (or snapshotDirPath)?

blockchain/index.js
```diff
@@ -181,66 +204,59 @@ class Blockchain {
...
-    * @param {Block} refBlock - The current highest block tin the querying nodes blockchain
+    * @param {Block} refBlockNumber - The current highest block number in the querying nodes blockchain
     * @return {list} A list of Block instances with refBlock at index 0, up to a maximuim length
     *                CHAIN_SEGMENT_LENGTH
     */
-  requestBlockchainSection(refBlock) {
-    const refBlockNumber = refBlock ? refBlock.number : -1;
-    const nextBlockNumber = refBlockNumber + 1;
-
+  requestBlockchainSection(refBlockNumber) {
```
platfowner, line 220: minor: The function name could mislead a little bit? What about getChainSegment() or getBlockchainSection()?

node/index.js
```diff
@@ -461,15 +482,22 @@ class BlockchainNode {
...
       if (!db.executeTransactionList(block.last_votes)) {
-        logger.error(`[${LOG_HEADER}] Failed to execute last_votes`)
+        logger.error(`[${LOG_HEADER}] Failed to execute last_votes`);
+        return; // TODO(liayoo): remove the invalid block and blocks that follow.
       }
       if (!db.executeTransactionList(block.transactions, block.number)) {
         logger.error(`[${LOG_HEADER}] Failed to execute transactions`)
+        return; // TODO(liayoo): remove the invalid block and blocks that follow.
+      }
+      if (block.state_proof_hash !== db.stateRoot.getProofHash()) {
```
platfowner, line 494: Happy to see this logic added! :)

blockchain/index.js
```diff
@@ -259,17 +275,20 @@ class Blockchain {
     return validBlocks;
   }
 
-  loadChain() {
+  loadChain(latestSnapshotBlockNumber) {
     const chainPath = this.blockchainPath;
     const newChain = [];
-    const blockPaths = BlockFileUtil.getAllBlockPaths(chainPath);
+    const numBlockFiles = fs.readdirSync(path.join(chainPath, CHAINS_N2B_DIR_NAME)).length;
+    const fromBlockNumber = latestSnapshotBlockNumber === undefined ? latestSnapshotBlockNumber :
```
platfowner, line 282: Should `0` be used instead of `latestSnapshotBlockNumber` when it's not a number? 

And would it be safer if isNumber() is used? e.g. `ChainUtil.isNumber(latestSnapshotBlockNumber) ? 0 : latestSnapshotBlockNumber + 1`

blockchain/index.js
```diff
@@ -48,18 +53,23 @@ class Blockchain {
         logger.info('################################################################');
         logger.info('\n');
       }
-      const newChain = this.loadChain();
+      const newChain = this.loadChain(latestSnapshotBlockNumber);
```
platfowner, line 56: minor: It seems to be notable that loadChain() loads only chains after latestSnapshotBlockNumber if a valid number is given. What about leaving a NOTE?

blockchain/index.js
platfowner: And any possibilities to be combined with getChainSegment() below?

common/chain-util.js
```diff
@@ -72,18 +70,6 @@ class ChainUtil {
     return address;
   }
 
-  // TODO(liayoo): Remove this function.
-  static genKeyPair() {
```
platfowner, line 76: Ah, it's not used anymore after the tx executor tool was removed. :)

node/index.js
```diff
@@ -75,23 +85,51 @@ class BlockchainNode {
...
+    // 3. Initialize DB (with the latest snapshot, if it exists)
     const startingDb =
         this.createDb(StateVersions.EMPTY, StateVersions.START, this.bc, this.tp, true);
-    startingDb.initDbStates();
-    this.executeChainOnDb(startingDb);
-    this.nonce = this.getNonceFromChain();
+    startingDb.initDbStates(latestSnapshot);
+
+    // 4. Execute the chain on the DB and finalize it.
+    this.executeChainOnDb(startingDb, latestSnapshotBlockNumber);
```
platfowner, line 116: Is the second param not used anymore?

node/index.js
```diff
@@ -175,44 +213,27 @@ class BlockchainNode {
     }
     const nodeVersion = `${StateVersions.NODE}:${blockNumber}`;
     this.syncDbAndNonce(nodeVersion);
+
+    // Create a snapshot of the new state
```
platfowner, line 217: This seems to be the best place for snapshotting!
minor: What about extracting the snapshotting logic below as a separate function for readability e.g. updateSnapshots()?

common/file-util.js
```diff
@@ -0,0 +1,169 @@
...
+      return -1;
+    }
+  }
+
+  static writeSnapshot(snapshotPath, blockNumber, snapshot) {
+    const filePath = this.getSnapshotPathByBlockNumber(snapshotPath, blockNumber);
+    fs.writeFileSync(filePath, zlib.gzipSync(Buffer.from(JSON.stringify(snapshot))));
+  }
+
+  static deleteSnapshot(snapshotPath, blockNumber) {
```
platfowner, line 157: minor: could this function be combined with writeSnapshot() with snapshot = null? Never mind if you prefer explicit functions. (In the case addSnapshot() would be clearer?)

common/file-util.js
```diff
@@ -0,0 +1,169 @@
...
+const path = require('path');
+const zlib = require('zlib');
+const _ = require('lodash');
+const {
+  CHAINS_N2B_DIR_NAME,
+  CHAINS_H2N_DIR_NAME,
+  SNAPSHOTS_N2S_DIR_NAME,
+} = require('./constants');
+const ChainUtil = require('./chain-util');
+const FILE_NAME_SUFFIX = 'json.gz';
```
cshcomcom, line 11: Looks too general. How about `GZIP_FILE_NAME_SUFFIX`?

cshcomcom: Or `BLOCK_FILE_NAME_SUFFIX`?

common/file-util.js
```diff
@@ -0,0 +1,169 @@
...
+    let latestSnapshotBlockNumber = -1;
+    let files = [];
+    try {
+      files = fs.readdirSync(snapshotPathPrefix);
+    } catch (err) {
+      logger.debug(`Failed to read snapshots: ${err.stack}`);
+      return { latestSnapshotPath, latestSnapshotBlockNumber };
+    }
+    for (const file of files) {
+      const blockNumber = _.get(file.split('.json.gz'), 0);
```
cshcomcom, line 48: How about `???_NAME_SUFFIX`!

p2p/index.js
liayoo: Done! Less ambiguity 👍

p2p/index.js
liayoo: Done

node/index.js

blockchain/index.js
liayoo: Wow, thanks for such a great suggestion! Combined the two functions into getBlockList().

blockchain/index.js
liayoo: This is some leftover from my earlier version of the code... latestSnapshotBlockNumber is now -1 if fast sync mode is not on or there's no snapshot. Thanks for catching this!

blockchain/index.js
liayoo: Right, added a jsDoc for the function. Thanks!

node/index.js
liayoo: Removed!

node/index.js

common/file-util.js
liayoo: There was some code duplication, so combined the two functions. Thanks for the idea!

common/file-util.js
liayoo: You're absolutely right, I renamed it to `JSON_GZIP_FILE_EXTENSION`. (there's a typo in my commit..) Thanks for the idea! 👍

common/file-util.js
liayoo: Done!

common/file-util.js
```diff
@@ -0,0 +1,168 @@
...
+    if (snapshot === null) { // Delete
+      if (fs.existsSync(filePath)) {
+        try {
+          fs.unlinkSync(filePath);
+        } catch (err) {
+          logger.debug(`Failed to delete ${filePath}: ${err.stack}`);
+        }
+      }
+    } else {
+      fs.writeFileSync(filePath, zlib.gzipSync(Buffer.from(JSON.stringify(snapshot))));
```
platfowner, line 163: As the database size gets bigger, the write time on storage, which is much slower than memory write, probably gets bigger. Would it be a good todo item to turn this write operation to async one?

liayoo:
Addressed all comments, PTAL @platfowner @cshcomcom

platfowner (approved):
LGTM with some additional minor comments.
Thanks!

node/index.js
```diff
@@ -461,15 +485,22 @@ class BlockchainNode {
   executeChainOnDb(db) {
     const LOG_HEADER = 'executeChainOnDb';
 
-    this.bc.chain.forEach((block) => {
+    for (const block of this.bc.chain) {
       if (!db.executeTransactionList(block.last_votes)) {
-        logger.error(`[${LOG_HEADER}] Failed to execute last_votes`)
+        logger.error(`[${LOG_HEADER}] Failed to execute last_votes`);
+        return; // TODO(liayoo): remove the invalid block and blocks that follow.
```
platfowner, line 491: (Sorry for this last minute comment) Could the job be just halted as it's still in boot-strapping stage?
Logging block numbers would be helpful for manual recovering?

node/index.js
```diff
@@ -461,15 +485,22 @@ class BlockchainNode {
...
       }
       if (!db.executeTransactionList(block.transactions, block.number)) {
         logger.error(`[${LOG_HEADER}] Failed to execute transactions`)
```
platfowner, line 494: block number logging?

node/index.js
```diff
@@ -461,15 +485,22 @@ class BlockchainNode {
...
+        return; // TODO(liayoo): remove the invalid block and blocks that follow.
+      }
+      if (block.state_proof_hash !== db.stateRoot.getProofHash()) {
+        logger.error(`[${LOG_HEADER}] Invalid state proof hash: ` +
+            `${db.stateRoot.getProofHash()}, ${block.state_proof_hash}`);
```
platfowner, line 499: ditto?

blockchain/index.js
```diff
@@ -72,13 +82,13 @@ class Blockchain {
...
   getBlockByHash(hash) {
     if (!hash) return null;
-    const blockPath = BlockFileUtil.getBlockPath(this.blockchainPath,
-        BlockFileUtil.readHashToNumber(this.blockchainPath, hash));
-    if (blockPath === undefined) {
+    const blockPath = FileUtil.getBlockPath(this.blockchainPath,
+        FileUtil.readHashToNumber(this.blockchainPath, hash));
+    if (!blockPath) {
       const found = this.chain.filter((block) => block.hash === hash);
       return found.length ? found[0] : null;
```
cshcomcom, line 89: How about use `find` instead of `filter`? I missed it in block optimization PR...

cshcomcom (approved):
LGTM with a minor comment!

common/file-util.js
liayoo: That would be really nice to have. I added a todo for it. Thanks!

node/index.js
liayoo: Thanks for another great suggestion! Updated to exit right away, logging the block number. 🙌

node/index.js

node/index.js
liayoo: Done!

blockchain/index.js
liayoo: Updated getBlockByHash() and getBlockByNumber() as well. Thanks!!

liayoo:
Thanks all for the fast & valuable reviews! Now running the integration tests & will merge this once they're passed.

PR #442, opened 2021-06-13
Bugfix/null ip addresses

liayoo:
Background: In GCP, a node is initially connecting to the tracker before setting its IP addresses (reporting them as nulls)

- Updated to wait until the IP addresses are set
- Replaced a [deprecated](https://nodejs.org/docs/latest-v12.x/api/url.html#url_url_format_urlobject) function (`url.format()`) with a [WHATWG URL API](https://nodejs.org/docs/latest-v12.x/api/url.html#url_the_whatwg_url_api)

### DIFF
--- a/p2p/index.js
+++ b/p2p/index.js
@@ -46,8 +46,8 @@ class P2pClient {
     this.startHeartbeat();
   }
 
-  run() {
-    this.server.listen();
+  async run() {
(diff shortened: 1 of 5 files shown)


platfowner (approved):
LGTM
Thanks!

Issue #444, opened 2021-06-15
Tx pool check API


cshcomcom:
I can't get tx pool information from json_rpc. We can add it to json_rpc/index.js with `getPoolSize` and `getPerAccountPoolSize`.

cshcomcom:
Feature is added to develop branch!

PR #448, opened 2021-06-15
Add `ain_getRemainingTransactionPoolSize` to json_rpc

cshcomcom:
## Description
- Now we can check remaining pool size per account and total remaining pool size
- `address` args is optional (undefined: total, address: per account)

Resolve: #444

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -16,8 +16,6 @@ const {
   BlockchainNodeStates,
   WriteDbOperations,
   TransactionStatus,
-  TX_POOL_SIZE_LIMIT,
-  TX_POOL_SIZE_LIMIT_PER_ACCOUNT,
 } = require('../common/constants');
 const { ConsensusStatus } = require('../consensus/constants');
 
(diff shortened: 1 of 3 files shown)


json_rpc/index.js
```diff
@@ -123,6 +125,18 @@ module.exports = function getMethods(node, p2pServer, minProtocolVersion, maxPro
       done(null, addProtocolVersion({result: node.tp.transactions}));
     },
 
+    ain_getRemainingTransactionPoolSize: function(args, done) {
```
platfowner, line 128: idea: can this be renamed ain_getTransactionPoolSizeUtilization and return both capacity and used? e.g. 
```
{
  limit: 100,
  used: 50,
}
  ```

Comment (platfowner) [reply json_rpc/index.js:128]: what about adding a similar client API?

Comment (cshcomcom) [reply json_rpc/index.js:128]: It seems to be more usable! I'll change it

Comment (cshcomcom) [reply json_rpc/index.js:128]: By 'client API' do you mean adding this API to the node?? like `http://node.ainetwork.ai:8080/get_value`?

Comment (cshcomcom) [reply json_rpc/index.js]: `ain_getTransactionPoolSizeUtilization` Done! Please, check the previous comment!

Comment (liayoo) [reply json_rpc/index.js]: How about `/tx_pool_size_util` ?

Comment (cshcomcom) [reply json_rpc/index.js]: @liayoo Looks more comfortable to use!

Comment (cshcomcom) [reply json_rpc/index.js]: I added `http://~~~/tx_ool_size_util! PTAL @liayoo @platfowner

File: client/index.js [hunk 2]
```diff
@@ -378,6 +380,24 @@ app.get('/dump_final_version', (req, res) => {
     .end();
 });
 
+app.get('/tx_pool_size_util', (req, res) => {
```
platfowner, line 383: As this api shares internal logic with the json rpc api, what about extracting the shared logic to a method of node/index.js? e.g. getTxPoolSizeUtilization()

platfowner (approved):
Thanks!

client/index.js
cshcomcom: Looks good! Done

cshcomcom:

PR #451, opened 2021-06-22
Feature/shallow get

cshcomcom:
## Description
- Now, we can use shallow query in json_rpc & client API
- Supports getValue, getOwner, getRule, getFunction methods
- Add related testcases to db.test.js

image

### DIFF
--- a/db/index.js
+++ b/db/index.js
@@ -408,9 +408,8 @@ class DB {
     if (stateNode === null) {
       return null;
     }
-    if (isShallow === true) {
-      const childLabels = stateNode.getChildLabels();
-      return childLabels.length > 0 ? childLabels : null;
+    if (isShallow) {
+      return stateNode.toJsObjectShallow();
     } else {
       return stateNode.toJsObject();
     }
--- a/db/index.js
+++ b/db/index.js
@@ -395,7 +395,7 @@ class DB {
     return DB.removeEmptyNodesRecursive(fullPath, 0, stateRoot);
   }
 
-  static readFromStateRoot(stateRoot, rootLabel, refPath, isGlobal, shardingPath) {
+  static readFromStateRoot(stateRoot, rootLabel, refPath, isGlobal, shardingPath, isShallow) {
     if (!stateRoot) return null;
     const parsedPath = ChainUtil.parsePath(refPath);
     const localPath = isGlobal === true ? DB.toLocalPath(parsedPath, shardingPath) : parsedPath;
@@ -405,28 +405,37 @@ class DB {
     }
     const fullPath = DB.getFullPath(localPath, rootLabel);
     const stateNode = DB.getRefForReadingFromStateRoot(stateRoot, fullPath);
-    return stateNode !== null ? stateNode.toJsObject() : null;
+    if (stateNode === null) {
+      return null;
+    }
+    if (isShallow === true) {
+      const childLabels = stateNode.getChildLabels();
+      return childLabels.length > 0 ? childLabels : null;
+    } else {
+      const stateObj = stateNode.toJsObject();
+      return stateObj;
+    }
   }
 
-  readDatabase(refPath, rootLabel, isGlobal) {
-    return DB.readFromStateRoot(this.stateRoot, rootLabel, refPath, isGlobal, this.shardingPath);
+  readDatabase(refPath, rootLabel, isGlobal, isShallow) {
+    return DB.readFromStateRoot(this.stateRoot, rootLabel, refPath, isGlobal, this.shardingPath, isShallow);
   }
 
   // TODO(platfowner): Support lookups on the final version.
-  getValue(valuePath, isGlobal) {
-    return this.readDatabase(valuePath, PredefinedDbPaths.VALUES_ROOT, isGlobal);
+  getValue(valuePath, isGlobal, isShallow) {
+    return this.readDatabase(valuePath, PredefinedDbPaths.VALUES_ROOT, isGlobal, isShallow);
   }
 
-  getFunction(functionPath, isGlobal) {
-    return this.readDatabase(functionPath, PredefinedDbPaths.FUNCTIONS_ROOT, isGlobal);
+  getFunction(functionPath, isGlobal, isShallow) {
+    return this.readDatabase(functionPath, PredefinedDbPaths.FUNCTIONS_ROOT, isGlobal, isShallow);
   }
 
-  getRule(rulePath, isGlobal) {
-    return this.readDatabase(rulePath, PredefinedDbPaths.RULES_ROOT, isGlobal);
+  getRule(rulePath, isGlobal, isShallow) {
+    return this.readDatabase(rulePath, PredefinedDbPaths.RULES_ROOT, isGlobal, isShallow);
   }
 
-  getOwner(ownerPath, isGlobal) {
-    return this.readDatabase(ownerPath, PredefinedDbPaths.OWNERS_ROOT, isGlobal);
+  getOwner(ownerPath, isGlobal, isShallow) {
+    return this.readDatabase(ownerPath, PredefinedDbPaths.OWNERS_ROOT, isGlobal, isShallow);
   }
 
   /**
@@ -456,7 +465,7 @@ class DB {
 
   static getValueFromStateRoot(stateRoot, statePath) {
     return DB.readFromStateRoot(
-        stateRoot, PredefinedDbPaths.VALUES_ROOT, statePath, false, []);
+        stateRoot, PredefinedDbPaths.VALUES_ROOT, statePath, false, [], false);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


db/index.js
```diff
@@ -680,7 +689,7 @@ class DB {
   }
 
   incValue(valuePath, delta, auth, timestamp, transaction, isGlobal) {
-    const valueBefore = this.getValue(valuePath, isGlobal);
+    const valueBefore = this.getValue(valuePath, isGlobal, false);
```
liayoo, line 692: suggestion) how about setting `isShallow = false` as default for getValue(), getFunction(), getRule() and getOwner() so that we can keep the current signatures? cc: @platfowner

db/index.js
```diff
@@ -405,28 +405,37 @@ class DB {
...
-  readDatabase(refPath, rootLabel, isGlobal) {
-    return DB.readFromStateRoot(this.stateRoot, rootLabel, refPath, isGlobal, this.shardingPath);
+  readDatabase(refPath, rootLabel, isGlobal, isShallow) {
+    return DB.readFromStateRoot(this.stateRoot, rootLabel, refPath, isGlobal, this.shardingPath, isShallow);
   }
 
   // TODO(platfowner): Support lookups on the final version.
-  getValue(valuePath, isGlobal) {
-    return this.readDatabase(valuePath, PredefinedDbPaths.VALUES_ROOT, isGlobal);
+  getValue(valuePath, isGlobal, isShallow) {
```
platfowner, line 425: Could you switch isGlobal and isShallow? It's likely that isShallow will be used more often than isGlobal, which is only used with sharding.

db/index.js
```diff
@@ -405,28 +405,37 @@ class DB {
     }
     const fullPath = DB.getFullPath(localPath, rootLabel);
     const stateNode = DB.getRefForReadingFromStateRoot(stateRoot, fullPath);
-    return stateNode !== null ? stateNode.toJsObject() : null;
+    if (stateNode === null) {
+      return null;
+    }
+    if (isShallow === true) {
+      const childLabels = stateNode.getChildLabels();
+      return childLabels.length > 0 ? childLabels : null;
```
platfowner, line 413: In the case of leaf node, null is returned. What about returning the value instead by adding toJsObjectShallow() method? e.g.
```
if (isShallow) {
  return stateNode.toJsObjectShallow();
} else {
  return stateNode.toJsObject();
}
```

db/index.js
```diff
@@ -405,28 +405,37 @@ class DB {
     }
     const fullPath = DB.getFullPath(localPath, rootLabel);
     const stateNode = DB.getRefForReadingFromStateRoot(stateRoot, fullPath);
-    return stateNode !== null ? stateNode.toJsObject() : null;
+    if (stateNode === null) {
+      return null;
+    }
+    if (isShallow === true) {
+      const childLabels = stateNode.getChildLabels();
```
platfowner, line 412: fyi, getChildLables() returns an array. I think it should be an object.

db/index.js
```diff
@@ -405,28 +405,37 @@ class DB {
     }
     const fullPath = DB.getFullPath(localPath, rootLabel);
     const stateNode = DB.getRefForReadingFromStateRoot(stateRoot, fullPath);
-    return stateNode !== null ? stateNode.toJsObject() : null;
+    if (stateNode === null) {
+      return null;
+    }
+    if (isShallow === true) {
```
platfowner, line 411: minor: I think `if (isShallow) {` would be enough in this case. fyi, there are some cases of explicitly comparing with `null` as `null` means "not existing".

db/index.js
liayoo: Ah, I agree with platfowner's comments that isShallow will be used more often than isGlobal. Changing the order sounds good to me. Scratch my comment!

db/index.js
```diff
@@ -405,28 +405,37 @@ class DB {
...
+      const childLabels = stateNode.getChildLabels();
+      return childLabels.length > 0 ? childLabels : null;
+    } else {
+      const stateObj = stateNode.toJsObject();
```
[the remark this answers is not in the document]
cshcomcom: Looks good! Done

db/index.js
cshcomcom: I agree with this suggestion! Done

db/index.js
cshcomcom: Looks very clean!! I'll change it right now.

db/index.js
cshcomcom: How about this rule? I referenced firebase docs (https://firebase.google.com/docs/database/rest/retrieve-data#shallow)
```
If the data at the location is a JSON primitive (string, number, or boolean) its value will simply be returned. 
If the data snapshot at the location is a JSON object, the values for each key will be truncated to true.
```
It looks in the most convenient form for users to use.

platfowner: Yes, an object of the first-level labels with value `true`.

cshcomcom: I changed it to object format! PTAL

db/index.js
cshcomcom: Done!

db/index.js
cshcomcom: Done! Thanks

cshcomcom:
All comments are addressed! PTAL @liayoo @platfowner

db/state-node.js
```diff
@@ -99,6 +99,16 @@ class StateNode {
     return obj;
   }
 
+  toJsObjectShallow() {
+    if (this.getIsLeaf()) {
+      return this.getValue();
+    }
+    return this.getChildLabels().reduce((shallowTable, label) => {
```
platfowner, line 106: minor: what about shallowTable -> shallowCopy?

db/state-node.js
```diff
@@ -99,6 +99,16 @@ class StateNode {
     return obj;
   }
 
+  toJsObjectShallow() {
```
platfowner, line 102: Could you add test cases to state-node.test.js?

unittest/db.test.js
```diff
@@ -269,6 +269,10 @@ describe("DB operations", () => {
         assert.deepEqual(node.db.getValue("test"), dbValues)
       })
 
+      it("when retrieving high value near top of database with is_shallow", () => {
```
platfowner, line 272: s/high value/value/?

unittest/db.test.js
```diff
@@ -269,6 +269,10 @@ describe("DB operations", () => {
...
+        assert.deepEqual(Object.keys(node.db.getValue("test", true, false)), Object.keys(dbValues))
```
platfowner, line 273: Could you compare without Object.keys()?

db/state-node.js
cshcomcom: Done!

db/state-node.js
cshcomcom: Done! Thanks

unittest/db.test.js
cshcomcom: Done!

unittest/db.test.js
cshcomcom: I implemented it because the shape of the object is different, do you have any good ideas?
A
```
{
  "a": true,
  "b": true,
  ...
}
```
B
```json
{
  "a": { ... },
  "b": { ... }, 
  ...
}
```

cshcomcom:
All comments are addressed! Round2 PTAL @platfowner

unittest/db.test.js
platfowner: What about just enumerating the expected values? e.g.
```
  assert.deepEqual(node.db.getValue("test", true, false), {
    "ai": true,
    "increment": true,
    "decrement": true,
    "nested": true,
    "shards": true,
  });
```

cshcomcom: Looks good!

cshcomcom: Done! Thanks

unittest/db.test.js
```diff
@@ -312,6 +326,10 @@ describe("DB operations", () => {
           }
         });
       })
+
+      it("when retrieving existing function config with is_shallow", () => {
+        assert.deepEqual(Object.keys(node.db.getFunction("test/test_function", true, false)), Object.keys(dbFuncs));
```
platfowner, line 331: could you do without Object.keys() here as well?

unittest/db.test.js
```diff
@@ -330,6 +348,10 @@ describe("DB operations", () => {
           }
         });
       })
+
+      it("when retrieving existing rule config with is_shallow", () => {
+        assert.deepEqual(Object.keys(node.db.getRule("test/test_rule", true, false)), Object.keys(dbRules));
```
platfowner, line 353: ditto

unittest/db.test.js
```diff
@@ -377,6 +399,10 @@ describe("DB operations", () => {
           }
         });
       })
+
+      it("when retrieving existing owner config with is_shallow", () => {
+        assert.deepEqual(Object.keys(node.db.getOwner("test/test_owner", true, false)), Object.keys(dbOwners))
```
platfowner, line 404: ditto

platfowner (approved):
LGTM with some additional comments.
Thanks!

unittest/db.test.js
cshcomcom: Done!

unittest/db.test.js
cshcomcom: Done!

unittest/db.test.js
cshcomcom: Done!

cshcomcom:
Thanks for reviews! Now merging!

PR #457, opened 2021-06-23
Add closed type to pull_request

cshcomcom:
## Background
- Performance test pipeline need to catch 'closed' event

## Description
- When declaring event types, the default value disappears, so I added `closed` to the default values.
- Refer to [docs](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request), default event type are 'opened, synchronize, reopened'.

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -11,6 +11,7 @@ on:
       - 'develop'
       - 'release/*'
       - 'master'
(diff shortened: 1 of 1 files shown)


platfowner (approved):
LGTM
Thanks!

cshcomcom:
Thanks for the reviews! Now merging

PR #459, opened 2021-06-28
Add cpu_status info to grafana.json

cshcomcom:
## Description
- Add cpu_status:usage to grafana.json

### DIFF
--- a/monitoring/grafana.json
+++ b/monitoring/grafana.json
@@ -25,9 +25,7 @@
       "dashes": false,
       "datasource": null,
       "fieldConfig": {
-        "defaults": {
-          "custom": {}
-        },
+        "defaults": {},
         "overrides": []
       },
       "fill": 1,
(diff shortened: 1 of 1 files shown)


platfowner (approved):
LGTM

BTW I think the cpu values can be converted to more human-friendly values (e.g. usage percent) by the blockchain node. (Maybe in the next release)

cshcomcom:
Thanks for the quick review! How can we get usage percent from elapsedTimeMs!?

cshcomcom:
Now merging

platfowner:
> Thanks for the quick review! How can we get usage percent from elapsedTimeMs!?

elapsedTimeMs is just a reference: https://github.com/ainblockchain/ain-blockchain/blob/master/p2p/server.js#L155

We can get whatever we want by manipulating given info?

platfowner:
> > Thanks for the quick review! How can we get usage percent from elapsedTimeMs!?
> 
> elapsedTimeMs is just a reference: https://github.com/ainblockchain/ain-blockchain/blob/master/p2p/server.js#L155
> 
> We can get whatever we want by manipulating given info?

ref: https://stackoverflow.com/questions/9565912/convert-the-output-of-os-cpus-in-node-js-to-percentage

Issue #466, opened 2021-07-07
[BUG] FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory


platfowner:
- How to reproduce:
`sh deploy_blockchain_incremental_gcp.sh spring seo 0 canary --setup`

- Error log message:
```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0xa1a640 node::Abort() [node]
 2: 0xa1aa4c node::OnFatalError(char const*, char const*) [node]
 3: 0xb9a68e v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xb9aa09 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xd57c85  [node]
 6: 0xd58316 v8::internal::Heap::RecomputeLimits(v8::internal::GarbageCollector) [node]
 7: 0xd64bd5 v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags) [nod
e]
 8: 0xd65a85 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReaso
n, v8::GCCallbackFlags) [node]
 9: 0xd6853c v8::internal::Heap::AllocateRawWithRetryOrFail(int, v8::internal::AllocationType, v8::internal::Alloca
tionOrigin, v8::internal::AllocationAlignment) [node]
10: 0xd2ef5b v8::internal::Factory::NewFillerObject(int, bool, v8::internal::AllocationType, v8::internal::Allocati
onOrigin) [node]
11: 0x107158e v8::internal::Runtime_AllocateInYoungGeneration(int, unsigned long*, v8::internal::Isolate*) [node]
12: 0x140dcd9  [node]
```

platfowner:
Fixed.

PR #467, opened 2021-07-07
Tweak deploy scripts and log messages for node starts

platfowner:
Change summary:
- Add 'canary' run mode 
- Set 'fast' sync mode as the default
- Increase nodejs memory limit to 4GB (by --max-old-space-size)
- Defer job-killing and file-removing steps as much as possible (in start_node_incremental_gcp.sh)
- Use [[ ... ]] expression instead of [ ... ] expression in if statement
- Leave more log messages when node starts 

Related issue: https://github.com/ainblockchain/ain-blockchain/issues/466

### DIFF
--- a/p2p/index.js
+++ b/p2p/index.js
@@ -148,7 +148,6 @@ class P2pClient {
   }
 
   async setTrackerEventHandlers() {
-    const node = this.server.node;
     this.trackerWebSocket.on('message', async (message) => {
       const parsedMsg = JSON.parse(message);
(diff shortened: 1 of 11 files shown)


p2p/index.js
```diff
@@ -175,6 +166,32 @@ class P2pClient {
     });
   }
 
+  async startNode(numLivePeers) {
```
liayoo, line 169: 🙌

liayoo (approved):
LGTM, thanks!!

platfowner:

PR #472, opened 2021-07-12
Set up & start tracker server incrementally with start_tracker_incremental_gcp.sh

platfowner:
Change summary:
- Set up & start tracker server incrementally with start_tracker_incremental_gcp.sh
- Merge setup stages to start stages in genesis deploy
- Added _genesis suffix and _local suffix to deploy scripts


start_node_incremental_gcp.sh
```diff
@@ -172,7 +172,7 @@ do
     printf "\nconsensusStatus = ${consensusStatus}\n"
     lastBlockNumber=$(curl -m 20 -X POST -H "Content-Type: application/json" --data "$(generate_post_data 'ain_getRecentBlockNumber')" "http://localhost:8080/json-rpc" | jq -r '.result.result')
     printf "\nlastBlockNumber = ${lastBlockNumber}\n"
-    if [[ "$consensusStatus" = "RUNNING" ]]; then
+    if [[ "$consensusStatus" = "RUNNING" ]] && [[ "$lastBlockNumber" -gt 0 ]]; then
```
liayoo, line 175: If we're just starting a new chain, the first few nodes may not be able to finalize blocks until more nodes are online. Is this assuming that in that case, we don't use these incremental scripts?

platfowner: The main purpose of this change was to stop the incremental deploy process unless the deployed node is checked synced and healthy, and I was thinking that initial start can be done with batch deploy. But just realized that /health_check API is a much better option as it's checking both consensus state and node state.
https://github.com/ainblockchain/ain-blockchain/blob/master/client/index.js#L58

Updated. PTAL

liayoo: I see! By batch deploy, do you mean `deploy_blockchain_gcp.sh`? Maybe then we should make clear that the incremental script should only be used for non-initial deployments? Or we could rename `deploy_blockchain_gcp.sh` to something like `deploy_blockchain_genesis_gcp.sh`. What do you think?
Using the health_check API sounds good to me! Thanks for the explanation!

liayoo (approved):
LGTM with a comment. Thanks!

start_node_incremental_gcp.sh
platfowner: Just renamed so. Thanks for the good idea.

PR #475, opened 2021-07-14
Feature/improve app experience

liayoo:
- Now users can create "public" apps, which allow **anyone** to set values, rules, functions by default (no write_owner permission), in addition to the existing admin config.
  - Will be adding a util function for this feature in ain-js as well.
- Organized & added missing scripts in package.json.


db/functions.js
```diff
@@ -543,31 +543,38 @@ class Functions {
...
-      const owner = {};
-      let rule = '';
-      const adminAddrList = Object.keys(adminConfig);
-      for (let i = 0; i < adminAddrList.length; i++) {
-        const addr = adminAddrList[i];
-        CommonUtil.setJsObject(owner, [OwnerProperties.OWNER, OwnerProperties.OWNERS, addr],
-            buildOwnerPermissions(true, true, true, true));
-        rule += `auth.addr === '${addr}'` + (i < adminAddrList.length - 1 ? ' || ' : '');
-      }
+    sanitizedVal[PredefinedDbPaths.MANAGE_APP_CONFIG_ADMIN] = adminConfig;
```
platfowner, line 557: Could you consider extracting these sanitization parts into a separate function?

platfowner, line 557: Could you consider rejecting the transactions with non-standard input for safety. We already have some reference code:
- https://github.com/ainblockchain/ain-blockchain/blob/develop/tx-pool/transaction.js#L239
- https://github.com/ainblockchain/ain-blockchain/blob/develop/db/index.js#L752

db/functions.js
```diff
@@ -543,31 +543,38 @@ class Functions {
...
+    const appPath = PathUtil.getAppPath(appName);
+    const owner = {};
+    let rule = '';
+    const adminAddrList = Object.keys(adminConfig);
+    for (let i = 0; i < adminAddrList.length; i++) {
+      const addr = adminAddrList[i];
+      CommonUtil.setJsObject(owner, [OwnerProperties.OWNER, OwnerProperties.OWNERS, addr],
+          buildOwnerPermissions(true, true, true, true));
+      rule += `auth.addr === '${addr}'` + (i < adminAddrList.length - 1 ? ' || ' : '');
```
platfowner, line 566: minor: you can consider using <array>.join() or _.join().
(ref: https://github.com/ainblockchain/ain-blockchain/blob/develop/db/state-node.js#L269)

db/functions.js
```diff
@@ -543,31 +543,38 @@ class Functions {
...
-            buildOwnerPermissions(true, true, true, true));
-        rule += `auth.addr === '${addr}'` + (i < adminAddrList.length - 1 ? ' || ' : '');
-      }
+    sanitizedVal[PredefinedDbPaths.MANAGE_APP_CONFIG_ADMIN] = adminConfig;
+    const appPath = PathUtil.getAppPath(appName);
+    const owner = {};
+    let rule = '';
+    const adminAddrList = Object.keys(adminConfig);
+    for (let i = 0; i < adminAddrList.length; i++) {
+      const addr = adminAddrList[i];
```
platfowner, line 563: applying CommonUtil.isCksumAddr()?
(ref: https://github.com/ainblockchain/ain-blockchain/blob/develop/db/state-util.js#L301)

db/functions.js
```diff
@@ -543,31 +543,38 @@ class Functions {
     const recordId = context.params.record_id;
     const resultPath = PathUtil.getCreateAppResultPath(appName, recordId);
     const sanitizedVal = {};
-    const adminConfig = value[PredefinedDbPaths.MANAGE_APP_CONFIG_ADMIN];
+    const adminConfig = _.get(value, PredefinedDbPaths.MANAGE_APP_CONFIG_ADMIN);
     const billingConfig = _.get(value, PredefinedDbPaths.MANAGE_APP_CONFIG_BILLING);
     const serviceConfig = _.get(value, PredefinedDbPaths.MANAGE_APP_CONFIG_SERVICE);
+    const isPublic = _.get(value, PredefinedDbPaths.MANAGE_APP_CONFIG_IS_PUBLIC);
```
platfowner, line 549: what about applying CommonUtil.isBool() to isPublic for sanity checking?

db/functions.js
liayoo: Ahhh that made _createApp much simpler..! Thanks for the suggestion!

liayoo: Done!

db/functions.js
liayoo: Done!

db/functions.js
liayoo: The check is now included in `sanitizeCreateAppConfig()`. Thanks!!

db/functions.js
liayoo: Done, thanks!

db/functions.js
```diff
@@ -536,44 +536,71 @@ class Functions {
...
+      return this.saveAndReturnFuncResult(context, resultPath, FunctionResultCode.INVALID_SERVICE_NAME);
+    }
+    const { sanitizedVal, errorCode } = this.sanitizeCreateAppConfig(value);
+    if (errorCode) {
+      return this.saveAndReturnFuncResult(context, resultPath, errorCode);
+    }
+    let rule;
+    const owner = {};
+    const adminAddrList = Object.keys(sanitizedVal[PredefinedDbPaths.MANAGE_APP_CONFIG_ADMIN]);
+    for (let i = 0; i < adminAddrList.length; i++) {
```
platfowner, line 587: Do we need this for-loop?

platfowner: Ah sorry, please ignore the previous comment.

db/functions.js
```diff
@@ -536,44 +536,71 @@ class Functions {
...
+      const addr = adminAddrList[i];
+      rule = adminAddrList.map((addr) => `auth.addr === '${addr}'`).join(' || ');
```
platfowner, line 589: can this line be moved to:
```
if (sanitizedVal[PredefinedDbPaths.MANAGE_APP_CONFIG_IS_PUBLIC]) {
  rule = true;
  ...
} else {
  // here?
}
```

platfowner (approved):
Thanks!

db/functions.js
liayoo: Oops... thanks for catching it!

db/functions.js
liayoo: Done, thanks!

db/functions.js
liayoo: I was resetting rules over and over in the loop, so placed it outside (in the 'else' as you suggested) 🙂

liayoo:
Thanks for the thorough review! Now merging~

Issue #473, opened 2021-07-13
Need to generalize .write to .rule: { .write: ... } for extensibility


platfowner:
Future extension:
- states optimization
- state listener 
- etc

platfowner:
Fixed.

PR #476, opened 2021-07-14
Move .write rule under .rule layer

platfowner:
Change summary:
- Add .rule layer over .write rule:
  - `.write: <write rule>` -> `.rule: { .write: <write rule> }`
- Rename .write rule: `.write` -> `write`

Related issue: https://github.com/ainblockchain/ain-blockchain/issues/473

### DIFF
--- a/db/state-util.js
+++ b/db/state-util.js
@@ -44,11 +44,11 @@ function getFunctionConfig(funcNode) {
 }
 
 function hasRuleConfig(ruleNode) {
-  return hasConfig(ruleNode, RuleProperties.WRITE);
+  return hasConfig(ruleNode, RuleProperties.RULE);
 }
 
 function getRuleConfig(ruleNode) {
-  return getConfig(ruleNode, RuleProperties.WRITE);
+  return getConfig(ruleNode, RuleProperties.RULE);
 }
 
 function hasOwnerConfig(ownerNode) {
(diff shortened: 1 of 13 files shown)


db/state-util.js
```diff
@@ -169,9 +169,16 @@ function isValidJsObjectForStates(obj) {
...
  */
  function isValidRuleConfig(ruleConfigObj) {
-  if (!CommonUtil.isBool(ruleConfigObj) && !CommonUtil.isString(ruleConfigObj)) {
+  if (!CommonUtil.isDict(ruleConfigObj)) {
     return { isValid: false, invalidPath: CommonUtil.formatPath([]) };
   }
+  if (!ruleConfigObj.hasOwnProperty(RuleProperties.WRITE)) {
+    return { isValid: false, invalidPath: CommonUtil.formatPath([]) };
+  }
+  const writeConfigObj = ruleConfigObj[RuleProperties.WRITE];
```
liayoo, line 178: minor) how about just `writeConfig` since it's either a boolean or a string?

liayoo: Or writeProp, like ownersProp?

liayoo (approved):

db/state-util.js
platfowner: Just done. Thanks for the idea!

platfowner:
Thanks for the fast review. Now being merged..

PR #480, opened 2021-07-15
Move entry point labels to PredefinedDbPaths from *Properties

platfowner:
Change summary:
- Move entry point labels to PredefinedDbPaths from *Properties

Rationale:
- They're actually not part of properties but predefined db paths

Related issue: https://github.com/ainblockchain/ain-blockchain/issues/473


liayoo (approved):
LGTM, thanks!

PR #482, opened 2021-07-18
State gas budget check & gas budget allowance for free tier apps

liayoo:
- Add state gas checks for service & apps
- Set aside 10% state & bandwidth budgets for free tier apps


liayoo:
@platfowner @minsulee2 @pengin7384 Please hold any further review until https://github.com/ainblockchain/ain-blockchain/pull/483 is merged & I incorporate the changes into my PR. Thanks!

db/index.js
```diff
@@ -929,6 +944,82 @@ class DB {
...
+    if (serviceStateSize > SERVICE_STATE_BUDGET) {
+      return CommonUtil.returnTxResult(
+          25, `Exceeded state budget limit for services (${serviceStateSize} > ${SERVICE_STATE_BUDGET})`);
+    }
+    if (appStateSize > APPS_STATE_BUDGET) {
+      return CommonUtil.returnTxResult(
+          26, `Exceeded state budget limit for apps (${appStateSize} > ${APPS_STATE_BUDGET})`);
+    }
+    const stateFreeTierInUse = this.getStateFreeTierInUse();
+    let freeTierLimitReached = false;
```
platfowner, line 988: minor: `const freeTierLimitReached = stateFreeTierInUse >= FREE_STATE_BUDGET;`?

db/index.js
```diff
@@ -929,6 +944,82 @@ class DB {
...
+    if (stateFreeTierInUse >= FREE_STATE_BUDGET) {
+      freeTierLimitReached = true;
+    }
+    const appStakesTotal = this.getAppStakesTotal();
+    const appNameList = CommonUtil.getAppNameList(op);
+    for (const appName of appNameList) {
```
platfowner, line 994: minor: `for (const appName of CommonUtil.getAppNameList(op)) {`?

db/index.js
```diff
@@ -929,6 +944,82 @@ class DB {
     return result;
   }
 
+  getStateFreeTierInUse() {
```
platfowner, line 947: This function recomputes the target value by iterating the app list for every single transaction, so the computation is linear to the app name list size. Could this time complexity be reduced by reusing the computed value some how? Or just leave a todo item for such cases of having many apps although it won't come in very near future? :)

db/index.js
```diff
@@ -929,6 +944,82 @@ class DB {
...
+  getStateSizes() {
+    const rootStateSize = this.getStateSizeAtPath('/');
+    const appStateSize = this.getStateSizeAtPath('apps');
+    const serviceStateSize = rootStateSize - appStateSize;
+    return { rootStateSize, appStateSize, serviceStateSize };
+  }
+
+  checkStateBudgetsAndUpdateGasAmount(op, tx, result, serviceStateSizeBefore) {
+    const LOG_HEADER = 'checkStateBudgetsAndUpdateTxGasAmount';
+    const { appStateSize, serviceStateSize } = this.getStateSizes();
```
platfowner, line 978: Can this value be added as a parameter of checkStateBudgetsAndUpdateGasAmount() to optimize the computation?

platfowner: Ah, the two calls are for 'before' and 'after' respectively. Please ignore the previous comment.

db/index.js
```diff
@@ -929,6 +944,82 @@ class DB {
...
+        // else, we allow apps without stakes
+      } else {
+        const appStateBudget = APPS_STATE_BUDGET * appStake / appStakesTotal;
+        if (appTreeSize > appStateBudget) {
+          return CommonUtil.returnTxResult(
+              28, `Exceeded state budget limit for app ${appName} (${appTreeSize} > ${appStateBudget})`);
+        }
+      }
+    }
+    const serviceStateSizeDelta = Math.max(serviceStateSize - serviceStateSizeBefore, 0);
```
platfowner, line 1011: It looks like the 'update' part of this function depends only on serviceStateSizeBefore and serviceStateSize. Could this be extracted into a separate function and called from out side of this function?

db/index.js

db/index.js

db/index.js
liayoo: Right, thank you for pointing that out. I was actually thinking about optimizing this as well.. It seems that in order to minimize computation, we would need to know, for each tx,
1) if staking amounts have been changed and 
2) which apps have been changed,
and update & record the state budget in use by free tier apps, say, as a variable of DB. I'm not sure where would be best to locate this code yet, so I'm just leaving a to-do for now. Will be thinking about the optimization though! If you have any ideas, please let me know.

db/index.js
liayoo: Great idea! Separated the updating part out of the checking function. Thanks!

db/index.js
```diff
@@ -930,6 +947,81 @@ class DB {
...
 
+  static updateStateGasAmounts(tx, result, serviceStateSizeBefore, serviceStateSizeAfter) {
+    const LOG_HEADER = 'updateStateGasAmounts';
+    const serviceStateSizeDelta = Math.max(serviceStateSizeAfter - serviceStateSizeBefore, 0);
+    logger.debug(`[${LOG_HEADER}] serviceStateSizeDelta: ${serviceStateSizeDelta}`);
+    const txGas = _.get(tx, 'extra.gas', {
+      bandwidth: { service: 0, app: {} },
+      state: { service: 0 }
+    });
+    CommonUtil.setJsObject(txGas, ['state', 'service'], serviceStateSizeDelta * STATE_GAS_CONSTANT);
```
platfowner, line 958: Would it be helpful for users if per-app state gas amount is added to tx gas as well?

liayoo: Thanks for the suggestion! I added an 'app' field, but it's just a total sum number instead of a mapping of each app name and its state gas amount. I think if we need the per-app gas amount, it could be changed to something like bandwidth, once we have a way to efficiently get the list of apps that the tx touched (which would be related to this comment as well: https://github.com/ainblockchain/ain-blockchain/pull/482#discussion_r672090595).

liayoo: Actually, come to think of it, I think it would be ok to use the getAppNameList() for this. I'll update the code and let you know!

liayoo: Updated!

liayoo:

db/index.js
platfowner: I think managing such information in state database is a good idea and is a good future work item, and a todo comment at this moment is enough. Thanks for sharing your thought.

db/index.js
```diff
@@ -930,6 +950,99 @@ class DB {
     return result;
   }
 
+  static updateStateGasAmounts(tx, result, stateTreeBytesBefore, stateTreeBytesAfter, stateGasAmountPerAppBefore, stateGasAmountPerAppAfter) {
+    const LOG_HEADER = 'updateStateGasAmounts';
+    const serviceStateBytesDelta = Math.max(stateTreeBytesAfter.service - stateTreeBytesBefore.service, 0);
+    const appStateBytesDelta = Object.keys(stateGasAmountPerAppAfter).reduce((acc, appName) => {
```
platfowner, line 956: minor: what about 'TreeBytes' instead of 'StateBytes' for consistency?

db/index.js
```diff
@@ -901,23 +905,39 @@ class DB {
       }
     }
     let result;
+    const stateTreeBytesBefore = this.getStateTreeBytes();
+    const stateGasAmountPerAppBefore = this.getStateTreeBytesPerApp(op);
```
platfowner, line 909: minor: what about 'stateTreeBytesPerAppBefore' for consistency?

platfowner: and it looks like the values are those before multiplying STATE_GAS_CONSTANT yet.

db/index.js
```diff
@@ -901,23 +905,39 @@ class DB {
...
     if (op.type === WriteDbOperations.SET) {
       result = this.executeMultiSetOperation(op.op_list, auth, timestamp, tx);
     } else {
       result = this.executeSingleSetOperation(op, auth, timestamp, tx);
     }
     const gasPrice = tx.tx_body.gas_price;
-    const gasAmountTotal = CommonUtil.getTotalGasAmount(tx.tx_body.operation, result);
+    const stateGasAmountPerAppAfter = this.getStateTreeBytesPerApp(op);
```
platfowner, line 916: ditto

db/index.js
```diff
@@ -930,6 +950,99 @@ class DB {
     return result;
   }
 
+  static updateStateGasAmounts(tx, result, stateTreeBytesBefore, stateTreeBytesAfter, stateGasAmountPerAppBefore, stateGasAmountPerAppAfter) {
+    const LOG_HEADER = 'updateStateGasAmounts';
+    const serviceStateBytesDelta = Math.max(stateTreeBytesAfter.service - stateTreeBytesBefore.service, 0);
+    const appStateBytesDelta = Object.keys(stateGasAmountPerAppAfter).reduce((acc, appName) => {
+      const delta = stateGasAmountPerAppAfter[appName] - stateGasAmountPerAppBefore[appName];
+      if (delta > 0) {
+        acc[appName] = delta * STATE_GAS_CONSTANT;
```
platfowner, line 959: What about just not multiplying STATE_GAS_CONSTANT as the values are actually state bytes delta (not gas amount)?

platfowner: Ah, I think it would be enough to rename 'appStateBytesDelta' -> 'appStateGasAmountDelta'.

db/index.js

db/index.js
liayoo: Thanks for pointing this out. Renamed it to treeBytesPerAppBefore, and reduced ambiguity between state gas and tree bytes. 
Here, it's tree bytes, so I left it as a raw value, without multiplying STATE_GAS_CONSTANT.

db/index.js
liayoo: Renamed it to `treeBytesPerAppAfter`. Thanks!

db/index.js
liayoo: Thanks, renamed it to `appStateGasAmountDelta`.

platfowner (approved):
LGTM
Could you update the PR title to more intuitive one for future reference?
Thanks for the nice implementation.

liayoo:
Thanks for the helpful reviews! Now merging..

Issue #477, opened 2021-07-15
Need state subtree bytes to be tracked


platfowner:
We need state subtree bytes to be managed for further applications (e.g. gas fee charging).

platfowner:
Implemented.

PR #483, opened 2021-07-19
Track tree size in bytes

platfowner:
Change summary:
- Track tree size in bytes (in addition to tree size in state nodes)

Related issues: https://github.com/ainblockchain/ain-blockchain/issues/477


db/state-node.js
```diff
@@ -13,23 +18,25 @@ class StateNode {
     // Used for leaf nodes only.
     this.value = null;
     this.proofHash = null;
-    this.treeHeight = null;
-    this.treeSize = null;
+    this.treeHeight = -1;
+    this.treeSize = -1;
+    this.treeBytes = -1;
```
liayoo, line 23: How about we initialize these values to 0?

db/state-node.js
```diff
@@ -52,7 +59,19 @@ class StateNode {
         that.proofHash === this.proofHash &&
         that.version === this.version &&
         that.treeHeight === this.treeHeight &&
-        that.treeSize === this.treeSize);
+        that.treeSize === this.treeSize &&
+        that.treeBytes === this.treeBytes);
+  }
+
+  computeNodeBytes() {
```
liayoo, line 66: Could this be simplified to just `return sizeof(this);`, or is that not an accurate way of getting the bytes of a StateNode object?

platfowner: Thanks for the good question. Actually it turned out that sizeof() doesn't cover Set nor Map. Also I thought it would be nice if we keep this value implementation-independent for such cases of multiple language cases. Please let me know if you have any other ideas.

liayoo: Ah, I see. Now it makes sense! I think the current implementation is quite reasonable. Thanks for the explanation 👍

db/state-node.js
platfowner: Done. I was trying to distinguish 0 values from pre-set values, but 0 values looks better. Thanks.

platfowner:
Addressed all comments. PTAL @liayoo

liayoo (approved):
LGTM, thanks!

platfowner:
Thanks for the fast review!

platfowner:
Now merging after full-round test run.

Issue #484, opened 2021-07-20
Need PENDING | EXECUTED tx states


platfowner:
We need PENDING | EXECUTED states for transactions.

platfowner:
Implementation added.

PR #492, opened 2021-07-21
Divide IN_POOL tx state into EXECUTED and PENDING states

platfowner:
Change summary:
- Divide IN_POOL tx state into EXECUTED and PENDING states

Related issues: https://github.com/ainblockchain/ain-blockchain/issues/484

### DIFF
--- a/tx-pool/index.js
+++ b/tx-pool/index.js
@@ -39,7 +39,7 @@ class TransactionPool {
     this.txCountTotal = 0;
   }
 
-  addTransaction(tx) {
+  addTransaction(tx, isExecutedTx = false) {
     // NOTE(platfowner): A transaction needs to be converted to an executable form
     //                   before being added.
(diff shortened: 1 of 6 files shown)


tx-pool/index.js
```diff
@@ -39,7 +39,7 @@ class TransactionPool {
     this.txCountTotal = 0;
   }
 
-  addTransaction(tx) {
+  addTransaction(tx, isExecutedTx = false) {
```
liayoo, line 42: question) I noticed that we aren't actually setting isExecutedTx = true anywhere we call addTransaction() yet. Is this intended?

platfowner: Right. Some critical part was mistakenly missing. Just fixed and added some test cases. Thanks.

platfowner:

liayoo (approved):
LGTM, thanks!

Issue #493, opened 2021-07-22
Need to include tx receipts in the state tree


liayoo:
We need to include the transaction receipts in the state tree, especially since we're going to update the protocol to add failed transactions in the block in the near future.

liayoo:
https://github.com/ainblockchain/ain-blockchain/pull/495

PR #495, opened 2021-07-22
Feature/receipts in state tree

liayoo:
- Record tx receipts in the state tree
  - `/receipts/<tx_hash>`: { address, block_number, gas_amount, gas_cost, result }
- Remove old tx receipts from the state tree (i.e. receipts from 1000 blocks ago)

#493


db/index.js
```diff
@@ -1083,6 +1088,48 @@ class DB {
...
+     } else {
+       sanitized.code = result.code;
+      if (result.error_message) {
+        sanitized.error_message = result.error_message;
+      }
+    }
+     return sanitized;
+  }
+
+  recordReceipt(auth, timestamp, tx, blockNumber, result) {
```
platfowner, line 1111: minor: what about more specific param name e.g. 'txExecResult'?

db/index.js
```diff
@@ -1083,6 +1088,48 @@ class DB {
     return true;
   }
 
+  static sanitizeResult(result) {
```
platfowner, line 1091: What about more specific name e.g. sanitizeTxExecResult()?
If this method does more than just normalizing the result, what about other names like simplifyTxExecResult() or something?

db/index.js
```diff
@@ -1083,6 +1088,48 @@ class DB {
...
+  recordReceipt(auth, timestamp, tx, blockNumber, result) {
+    const receiptPath = PathUtil.getReceiptsPath(tx.hash);
+    const gasAmountSum = _.get(result, 'gas_amount_total.bandwidth.service', 0) +
+        _.get(result, 'gas_amount_total.state.service', 0);
+    const receipt = {
+      address: auth.addr,
+      block_number: blockNumber,
+      gas_amount: gasAmountSum,
+      gas_cost: result.gas_cost_total,
+      result: DB.sanitizeResult(result)
```
platfowner, line 1120: what about 'tx_exec_result' or just 'exec_result'?

db/index.js
```diff
@@ -1083,6 +1088,48 @@ class DB {
...
+  }
+
+  recordReceipt(auth, timestamp, tx, blockNumber, result) {
+    const receiptPath = PathUtil.getReceiptsPath(tx.hash);
+    const gasAmountSum = _.get(result, 'gas_amount_total.bandwidth.service', 0) +
+        _.get(result, 'gas_amount_total.state.service', 0);
+    const receipt = {
+      address: auth.addr,
+      block_number: blockNumber,
+      gas_amount: gasAmountSum,
```
platfowner, line 1118: Would it be helpful for users if it logs the detailed (or raw) information?
e.g. `gas_amount: _.get(result, 'gas_amount_total', 0),`?

platfowner: As we're not allowing empty object setting now, I guess the empty value setting part (e.g. apps: {}) needs to be updated (see https://github.com/ainblockchain/ain-blockchain/blob/develop/db/state-util.js#L140 ,
https://github.com/ainblockchain/ain-blockchain/blob/develop/db/index.js#L917 , and others).

platfowner: Just leaving the detailed part of the gas amount in txExecResult and renaming 'gas_amount' -> 'gas_amount_by_transfer' here would be a better option?

db/index.js
```diff
@@ -1083,6 +1088,48 @@ class DB {
...
+    }
+     return sanitized;
+  }
+
+  recordReceipt(auth, timestamp, tx, blockNumber, result) {
+    const receiptPath = PathUtil.getReceiptsPath(tx.hash);
+    const gasAmountSum = _.get(result, 'gas_amount_total.bandwidth.service', 0) +
+        _.get(result, 'gas_amount_total.state.service', 0);
+    const receipt = {
+      address: auth.addr,
```
platfowner, line 1116: Could you add these properties to PredefinedDbPaths as they're state paths (including 'billing' below)?

db/index.js
platfowner: Would 'bandwidth_gas_amount' be a better name?

platfowner: Ah, sorry for now 'gas_amount' looks just fine to me.

db/index.js

db/index.js
liayoo: Removed the sanitization logic as discussed offline :)

db/index.js
liayoo: Updated to exec_result. Thanks for the suggestion!

db/index.js
liayoo: Added the raw exec_result field which includes all the result code, error messages, gas amounts & costs of the operations.

db/index.js

liayoo:

db/functions.js
```diff
@@ -427,10 +423,10 @@ class Functions {
   }
 
   returnFuncResult(context, code, extraGasAmount = 0) {
-    const opResultList = Functions.getOpResultList(context);
+    const opResultListObj = CommonUtil.convertListToObj(context.opResultList);
     const funcResultToReturn = {};
-    if (!CommonUtil.isEmpty(opResultList)) {
-      funcResultToReturn.op_results = opResultList;
+    if (CommonUtil.isDict(opResultListObj)) {
```
platfowner, line 428: question: didn't isEmpty() work? I think isEmpty() case includes null.

client/index.js
```diff
@@ -398,6 +398,22 @@ app.get('/get_transaction', (req, res, next) => {
     .end();
 });
 
+app.get('/get_block_by_hash', (req, res, next) => {
```
platfowner, line 401: suggestion: what about adding json-rpc APIs as well?

common/constants.js
```diff
@@ -44,6 +44,8 @@ const PORT = process.env.PORT || getPortNumber(8080, 8080);
 const P2P_PORT = process.env.P2P_PORT || getPortNumber(5000, 5000);
 const LIGHTWEIGHT = CommonUtil.convertEnvVarInputToBool(process.env.LIGHTWEIGHT);
 const SYNC_MODE = process.env.SYNC_MODE || 'full';
+const MAX_NUM_BLOCKS_RECEIPTS = process.env.MAX_NUM_BLOCKS_RECEIPTS ?
```
platfowner, line 47: Could you consider simpler name MAX_NUM_RECEIPTS as neither the receipts nor this limit is a per-block value?

platfowner (approved):
Thanks!

common/constants.js
liayoo: It's actually not the max number of receipts we can keep, but the max number of blocks that we can keep the receipts in the state for. Maybe `MAX_NUM_BLOCKS_FOR_RECEIPTS_IN_STATE` makes more sense? Please let me know if you have a better idea!

platfowner: I see. That makes sense.
Then what about `MAX_BLOCK_NUMBERS_FOR_RECEIPTS`?

client/index.js
liayoo: Thanks for the suggestion! Checked that we already have ain_getBlockByHash and ain_getBlockByNumber :)

db/functions.js
liayoo: Right, thanks for pointing that out. Reverted!

common/constants.js
liayoo: Sounds good to me! Thanks for the idea!

liayoo:
Thanks for the quick review & for sharing ideas with me! Now merging..

PR #496, opened 2021-07-22
Bugfix/increase state budget

liayoo:
- Increase STATE_TREE_BYTES_LIMIT and --max-old-space-size to 5GB


start_node_incremental_gcp.sh
```diff
@@ -147,7 +147,7 @@ eval $KILL_CMD
 printf "\n#### [Step 6] Start new node server ####\n\n"
 
 sleep 10
-MAX_OLD_SPACE_SIZE_MB=4000
+MAX_OLD_SPACE_SIZE_MB=5000
```
platfowner, line 150: Taking overhead into account, could you please use 6000 instead?

start_node_genesis_gcp.sh
```diff
@@ -109,7 +109,7 @@ export STAKE=100000
 export BLOCKCHAIN_DATA_DIR="/home/ain_blockchain_data"
 
 printf "\nStarting up Blockchain Node server..\n\n"
-START_CMD='nohup node --async-stack-traces --max-old-space-size=4000 client/index.js >/dev/null 2>error_logs.txt &'
+START_CMD='nohup node --async-stack-traces --max-old-space-size=5000 client/index.js >/dev/null 2>error_logs.txt &'
```
platfowner, line 112: ditto
and could you define variable MAX_OLD_SPACE_SIZE_MB here too?

platfowner (approved):
LGTM with some comments.

start_node_incremental_gcp.sh
liayoo: Done!

start_node_genesis_gcp.sh
liayoo: Done!

Issue #486, opened 2021-07-20
Migrate nonce to getAccountNonceAndTimestamp() and updateAccountNonceAndTimestamp()


platfowner:
Now we have nonce and timestamp in state db, so we can migrate blockchain node's nonce to them.

platfowner:
Done.

PR #499, opened 2021-07-23
Use node.db.getAccountNonceAndTimestamp() instead of node.nonce

platfowner:
Change summary:
- Use node.db.getAccountNonceAndTimestamp() instead of node.nonce
- Add timestamp APIs (ain_getTimestamp(), /get_timestamp)

Related issues: https://github.com/ainblockchain/ain-blockchain/issues/486

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -262,8 +262,7 @@ class BlockchainNode {
   }
 
   getNonce() {
-      const { nonce } = this.db.getAccountNonceAndTimestamp(this.account.address);
-      return nonce;
+    return this.getNonceForAddr(this.account.address, true);
   }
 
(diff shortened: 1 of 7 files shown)


node/index.js
```diff
@@ -267,21 +261,34 @@ class BlockchainNode {
     return transactionInfo;
   }
 
-  getNonceForAddr(address, fromPending, fromDb = false) {
+  getNonce() {
```
liayoo, line 264: How about we have fromPending arg here as well?

liayoo (approved):
LGTM with a comment. Thanks!

node/index.js
platfowner: Just updated. Thanks!

PR #501, opened 2021-07-23
Update grafana config

platfowner:
Change summary:
- Add state tree bytes
- Replace cpu usage with cpu usage percent


liayoo (approved):
LGTM, thanks!

PR #504, opened 2021-07-24
Feature/state size limits

liayoo:
- Use `MAX_STATE_TREE_SIZE_PER_BYTE` and `STATE_TREE_BYTES_LIMIT` to get `TREE_SIZE_BUDGET`
- Add tree size limits for service & apps


db/index.js
```diff
@@ -952,12 +955,16 @@ class DB {
...
+  static updateStateGasAmount(tx, result, allStateUsageBefore, allStateUsageAfter, stateUsagePerAppBefore, stateUsagePerAppAfter) {
     const LOG_HEADER = 'updateStateGasAmounts';
+    const serviceTreeBytesDelta =
+        _.get(allStateUsageAfter, `service.${StateInfoProperties.TREE_BYTES}`, 0) -
+        _.get(allStateUsageBefore, `service.${StateInfoProperties.TREE_BYTES}`, 0);
     const stateGasAmount = {
-      service: Math.max(treeBytesAfter.service - treeBytesBefore.service, 0) * STATE_GAS_CONSTANT,
-      app: Object.keys(treeBytesPerAppAfter).reduce((acc, appName) => {
-          const delta = treeBytesPerAppAfter[appName] - treeBytesPerAppBefore[appName];
+      service: Math.max(serviceTreeBytesDelta, 0) * STATE_GAS_CONSTANT,
```
platfowner, line 964: minor: what about _CONSTANT -> _COEFFICIENT?

db/index.js
```diff
@@ -952,12 +955,16 @@ class DB {
...
+      app: Object.keys(stateUsagePerAppAfter).reduce((acc, appName) => {
```
platfowner, line 965: For safety, could you make sure that this prop is not set when empty?

platfowner (approved):
Thanks!

db/index.js
liayoo: Done! Thanks!

db/index.js
liayoo: Thanks for the suggestion! Changed the deleting logic to not setting if the value is empty.

liayoo:
Thanks for the quick review! 🙌 Now merging~

PR #506, opened 2021-07-28
Gas fee charging & receipts for failed txs

liayoo:
- Include failed transactions in blocks
- Charge fees for failed transactions & record receipts
- Add pre-checks for executeTransaction()
- Introduce MIN_STAKING_FOR_APP_TX and MIN_BALANCE_FOR_SERVICE_TX for pre-checking


client/index.js
```diff
@@ -407,7 +407,7 @@ app.get('/get_block_by_hash', (req, res, next) => {
 });
 
 app.get('/get_block_by_number', (req, res, next) => {
-  const block = node.bc.getBlockByNumber(req.query.number);
+  const block = node.bc.getBlockByNumber(Number(req.query.number));
```
platfowner, line 410: Could you add a util function (e.g. toNumber()) for this to rule util and common util (ref: toBool())?

consensus/index.js
```diff
@@ -353,9 +315,7 @@ class Consensus {
     this.node.removeOldReceipts(blockNumber, tempDb);
 
     for (const voteTx of lastVotes) {
-      const res = this.executeLastVoteOrAbort(tempDb, voteTx);
-      if (!res) {
-        this.node.destroyDb(tempDb);
+      if (CommonUtil.isFailedTx(tempDb.executeTransaction(Transaction.toExecutable(voteTx), false))) {
```
platfowner, line 318: question: is it correct that vote txs are protected by rules from attackers?

db/index.js
```diff
@@ -1060,140 +1075,239 @@ class DB {
...
+      const appStake = this.getAppStake(appName);
+      if (appStake < MIN_STAKING_FOR_APP_TX) {
+        return CommonUtil.logAndReturnTxResult(
+          logger, 35, `[${LOG_HEADER}] App stake too low (${appStake} < ${MIN_STAKING_FOR_APP_TX})`);
+      }
+    });
+    return true;
+  }
+
+  executeTransactionPrecheck(tx, restoreIfFails, blockNumber) {
```
platfowner, line 1237: minor: what about 'precheckTransaction'?

db/index.js
```diff
@@ -229,6 +231,13 @@ class DB {
     return true;
   }
 
+  destroyDb() {
```
platfowner, line 234: Could you also move createDb() here?

genesis-configs/base/genesis_rules.json
```diff
@@ -162,13 +162,6 @@
       }
     }
   },
-  "receipts": {
```
platfowner, line 165: question: is it safe to remove this rule?

db/index.js
```diff
@@ -1060,140 +1075,239 @@ class DB {
...
+  isBillingUser(billingAppName, billingId, userAddr) {
+    return this.getValue(
+        PathUtil.getManageAppBillingUsersPath(billingAppName, billingId) + '/' + userAddr) === true;
+  }
+
+  precheckTxBillingParam(op, addr, billing, blockNumber) {
+    if (!billing || blockNumber === 0) {
+      return true;
+    }
+    const LOG_HEADER = 'precheckTxBillingParam';
```
platfowner, line 1183: Could you move this line up?

db/index.js
```diff
@@ -1060,140 +1075,239 @@ class DB {
...
-    const LOG_HEADER = 'executeTransaction';
+  getBalance(addrOrServAcnt) {
+    return this.getValue(CommonUtil.getBalancePath(addrOrServAcnt)) || 0;
+  }
+
+  precheckBalanceAndStakes(op, addr, billing, blockNumber) {
+    if (blockNumber === 0) {
+      return true;
+    }
+    const LOG_HEADER = 'precheckBalanceAndStakes';
```
platfowner, line 1217: ditto

db/index.js
```diff
@@ -1060,140 +1075,239 @@ class DB {
...
+    const checkBillingResult = this.precheckTxBillingParam(op, addr, billing, blockNumber);
+    if (checkBillingResult !== true) {
+      return checkBillingResult;
+    }
+    const checkBalanceResult = this.precheckBalanceAndStakes(op, addr, billing, blockNumber);
+    if (checkBalanceResult !== true) {
+      return checkBalanceResult;
+    }
+    if (restoreIfFails) {
+      if (!this.backupDb()) {
```
platfowner, line 1262: Could this backup part be moved to executeTransaction() where restoreDb lives for readability?

db/index.js
```diff
@@ -1060,140 +1075,239 @@ class DB {
...
+    tx.setExtraField('executed_at', Date.now());
+    // NOTE(platfowner): It's not allowed for users to send transactions with auth.fid.
+    const auth = { addr: tx.address };
+    const timestamp = txBody.timestamp;
+    const executionResult = this.executeOperation(txBody.operation, auth, timestamp, tx, blockNumber);
+    if (CommonUtil.isFailedTx(executionResult)) {
+      if (restoreIfFails) {
+        this.restoreDb();
+      } else {
+        this.destroyDb();
```
platfowner, line 1288: Could we keep the caller makes explicit decision whether to destroy it or not? I think implicit destroy could be bug prone.

db/index.js
```diff
@@ -1060,140 +1075,239 @@ class DB {
...
       }
-      const treeSize = stateInfo[StateInfoProperties.TREE_SIZE];
-      if (treeSize > TREE_SIZE_LIMIT) {
-        return CommonUtil.returnTxResult(24, `Out of tree size limit ` +
-            `(${treeSize} > ${TREE_SIZE_LIMIT})`);
+    }
+    return true;
+  }
+
+  executeTransaction(tx, restoreIfFails, blockNumber = 0) {
```
platfowner, line 1270: suggestion: what about providing a default value: restoreIfFails = false for temp db cases?

common/common-util.js
```diff
@@ -295,6 +295,11 @@ class CommonUtil {
     return newObj;
   }
 
+  static execTxPrecheckFailed(result) {
```
platfowner, line 298: minor: what about 'txPrecheckFailed'?

db/index.js
```diff
@@ -696,7 +705,7 @@ class DB {
     const valueBefore = this.getValue(valuePath, false, isGlobal);
     logger.debug(`VALUE BEFORE:  ${JSON.stringify(valueBefore)}`);
     if ((valueBefore && typeof valueBefore !== 'number') || typeof delta !== 'number') {
```
platfowner, line 707: CommonUtil.isNumber() can be used here and below.

db/index.js
```diff
@@ -913,41 +922,39 @@ class DB {
...
     const gasAmountTotal = {
       bandwidth: CommonUtil.getTotalBandwidthGasAmount(tx.tx_body.operation, result),
       state: { service: 0 }
     };
+    tx.setExtraField('gas', gasAmountTotal);
     result.gas_amount_total = gasAmountTotal;
     result.gas_cost_total = 0;
-    // TODO(platfowner): Consider charging gas fee for the failure cases.
     if (!CommonUtil.isFailedTx(result)) {
+      const stateInfo = this.getStateInfo('/');
```
platfowner, line 934: minor suggestion: `{ [StateInfoProperties.TREE_HEIGHT]: treeHeight, [StateInfoProperties.TREE_SIZE]: treeSize } = this.getStateInfo('/');`?

db/index.js
```diff
@@ -1060,140 +1075,239 @@ class DB {
...
+    let balance = this.getBalance(billedTo);
+    if (balance < gasAmountChargedByTransfer) {
+      Object.assign(executionResult, {
+        code: 36,
+        error_message: `Failed to collect gas fee: balance too low (${balance} / ${gasAmountChargedByTransfer})`
+      });
+      this.restoreDb(); // Revert changes made by the tx operations
+      balance = this.getBalance(billedTo);
+      gasAmountChargedByTransfer = Math.min(balance, serviceBandwidthGasAmount);
+      executionResult.gas_amount_charged = gasAmountChargedByTransfer;
```
platfowner, line 1144: minor: this line can be moved out of this block?

db/index.js
```diff
@@ -1060,140 +1075,239 @@ class DB {
...
+    return this.getValue(
+        PathUtil.getManageAppBillingUsersPath(billingAppName, billingId) + '/' + userAddr) === true;
+  }
+
+  precheckTxBillingParam(op, addr, billing, blockNumber) {
+    if (!billing || blockNumber === 0) {
+      return true;
+    }
+    const LOG_HEADER = 'precheckTxBillingParam';
+    const billingParsed = billing.split('|');
```
platfowner, line 1184: Could you add a util function like parseServAcntName()?

db/index.js
```diff
@@ -1060,140 +1075,239 @@ class DB {
...
-          29, `Failed to record receipt ${JSON.stringify(recordRes, null, 2)}`, 0);
+    this.writeDatabase([PredefinedDbPaths.VALUES_ROOT, ...CommonUtil.parsePath(receiptPath)], receipt);
+  }
+
+  isBillingUser(billingAppName, billingId, userAddr) {
+    return this.getValue(
+        PathUtil.getManageAppBillingUsersPath(billingAppName, billingId) + '/' + userAddr) === true;
+  }
+
+  precheckTxBillingParam(op, addr, billing, blockNumber) {
```
platfowner, line 1179: minor: what about 'Params'?

consensus/index.js
liayoo: Yes, we are performing basic checks in the rules (e.g. whether the vote/proposal is signed by the correct address, validator / proposer has staked enough AIN, and the format of the value, ...). We can always add more checks in the rule and/or checkVoteTx() & checkProposal().

db/index.js
liayoo: It seems a little tricky to achieve this, since createDb() needs access to stateManger in BlockchainNode.. and destroying self makes sense but creating self / new instance in a db seems odd to me.. What do you think?
To minimize confusion, we could also try renaming the functions:
- backupDb() -> createBackup()
- restoreDb() -> restoreFromBackup()
- destroyDb() -> destroySelf() or destroy()

genesis-configs/base/genesis_rules.json
liayoo: The checks previously done in the rules are now done in code (precheckTransaction(), collectFee()). Plus, the operation was not supposed to trigger any other functions, so I safely changed from setValue() to writeDatabase() and removed the rule (now no one can write with a tx at this path).

node/index.js
```diff
@@ -420,15 +390,14 @@ class BlockchainNode {
...
+    const result = this.db.executeTransaction(executableTx, true, this.bc.lastBlockNumber() + 1);
     if (CommonUtil.isFailedTx(result)) {
       if (FeatureFlags.enableRichTransactionLogging) {
         logger.error(
             `[${LOG_HEADER}] FAILED TRANSACTION: ${JSON.stringify(executableTx, null, 2)}\n ` +
             `WITH RESULT:${JSON.stringify(result)}`);
       }
-      const errorCode = _.get(result, 'code');
-      if (errorCode === TX_NONCE_ERROR_CODE || errorCode === TX_TIMESTAMP_ERROR_CODE) {
+      if (!CommonUtil.execTxPrecheckFailed(result)) {
```
platfowner, line 400: Looks nice. A NOTE would be helpful for readers.

consensus/index.js
platfowner: Thanks for the explanation. It was just a double-check.

db/index.js
platfowner: static creator method seems to be a pattern quite often used and we have an example:
https://github.com/ainblockchain/ain-blockchain/blob/develop/db/state-node.js#L27

As we already passing a ref to the state manager to the constructor of the class, I think it would be just okay to pass it to the creator e.g.
`static create(baseVersion, newVersion, bc, tp, finalizeVersion, isNodeDb, blockNumberSnapshot, stateManager) {`

I have an action item "Block number / epoch key for state version" for Q3, I expect this structure could be simplified further.
What do you think?

genesis-configs/base/genesis_rules.json
platfowner: I see. A NOTE on the use of writeDatabase() would be helpful for readers.

db/index.js
liayoo: Ah, I see! A static function sounds good to me! Thanks for the idea 👍

client/index.js
liayoo: I noticed there was CommonUtil.numberOrZero() & used that instead. Thanks!

db/index.js
liayoo: Done. Thanks!

genesis-configs/base/genesis_rules.json
liayoo: A NOTE has been added~

db/index.js
liayoo: Done

db/index.js
liayoo: Done

db/index.js
liayoo: Done

db/index.js
liayoo: You're absolutely right. Changed from implicit -> explicit destroy.

db/index.js
liayoo: Done

common/common-util.js
liayoo: Done

db/index.js
liayoo: Done, and the logic was outdated (checking valueBefore is undefined), so updated that as well :)

db/index.js
liayoo: Done

db/index.js
liayoo: Oops! Thanks for pointing that out. Done.

db/index.js
liayoo: Changed to use parseServAcntName(). Thanks!

db/index.js
liayoo: Done

node/index.js
liayoo: Thanks! Added a NOTE at the function definition.

liayoo:

db/index.js
```diff
@@ -1074,140 +1107,238 @@ class DB {
...
+    tx.setExtraField('executed_at', Date.now());
+    // NOTE(platfowner): It's not allowed for users to send transactions with auth.fid.
+    const auth = { addr: tx.address };
+    const timestamp = txBody.timestamp;
+    const executionResult = this.executeOperation(txBody.operation, auth, timestamp, tx, blockNumber);
+    if (CommonUtil.isFailedTx(executionResult)) {
+      if (restoreIfFails) {
+        this.restoreDb();
+      } else {
+        return executionResult;
```
platfowner, line 1320: Is it okay to skip collecting fee in this case?

blockchain/block.js
```diff
@@ -323,7 +323,7 @@ class Block {
     tempGenesisDb.initDbStates();
     const resList = [];
     for (const tx of genesisTxs) {
-      const res = tempGenesisDb.executeTransaction(Transaction.toExecutable(tx));
+      const res = tempGenesisDb.executeTransaction(Transaction.toExecutable(tx), false);
```
platfowner, line 326: minor: false param can be skipped by the default value here and below?

node/index.js
platfowner: Ah, I mean a note why tx pool addition is skipped in the pre-check failure cases. :)

platfowner (approved):
LGTM with some comments/questions.
Thanks for the quick progress!

minor: Could you change the PR title to a more relevant one for future reference?

platfowner:
> LGTM with some comments/questions.
> Thanks for the quick progress!
> 
> minor: Could you change the PR title to a more relevant one for future reference?

Ah, I mean more descriptive one. :)

db/index.js
liayoo: Thanks for the great question! 
We're setting restoreIfFails = true in two cases: 1) genesis txs and 2) consensus txs. I thought we could skip the backup, restore & collecting gas for these cases, for different reasons. 
For the genesis txs, if the txs are corrupted, the node cannot begin anyway. 
For consensus txs, we're currently not charging gas fees. I was debating whether to charge fees for votes & proposals, but decided not to since we're already making them stake AIN & use their resources for the network.. But I'm happy to discuss this further if you feel differently! 
+ I'll add some notes about this if we decide to keep this logic for now.

platfowner: It seems the gas fee collection is skipped by returning exectionResult when `isFailedTx() && !restoreIfFails`??

For such cases of skipping gas fee collection, what about introducing a separate param explicitly?

liayoo: Added an optional `skipFees` parameter. Thanks for the idea!!

blockchain/block.js
liayoo: Done!

node/index.js
liayoo: Done! :)

liayoo:
Thanks for the thorough review & all the helpful suggestions. Now merging..!

Issue #513, opened 2021-07-30
[BUG] broken test


platfowner:
@liayoo could you take a look at the broken test with message:
```
$ yarn run test_integration_node -g 'write rule: auth.fid: with function permission'
yarn run v1.22.10
$ ./node_modules/mocha/bin/mocha --timeout 160000 integration/node.test.js -g 'write rule: auth.fid: with function permission'


  Blockchain Node
    Function triggering
      Function permission
        Write rule: auth.fid
          1) write rule: auth.fid: with function permission


  0 passing (25s)
  1 failing

  1) Blockchain Node
       Function triggering
         Function permission
           Write rule: auth.fid
             write rule: auth.fid: with function permission:

      AssertionError: expected { Object (func_results, code, ...) } to deeply equal { Object (code, func_results, ...) }
      + expected - actual

             "service": 0
           }
           "state": {
             "app": {
      -        "test": 1516
      +        "test": 1348
             }
      -      "service": 16
      +      "service": 0
           }
         }
         "gas_cost_total": 0
       }
      
      at Context.<anonymous> (integration/node.test.js:3537:18)
      at processImmediate (internal/timers.js:461:21)
```

liayoo:
Fixed in https://github.com/ainblockchain/ain-blockchain/pull/514

PR #514, opened 2021-08-01
Fix broken tests & receipt, gas related bugs

liayoo:
- Problem 1. Flaky node integration test cases (related: https://github.com/ainblockchain/ain-blockchain/issues/513)
  - Added `waitUntilTxFinalized` calls for transactions that fail but get included in blocks
- Problem 2. Changing receipts values & state proof hashes
  - Removed the stringified transaction from tx results & receipts for stable results (was getting state proof hash errors)
- Problem 3. Unstable tree byte calculation
  -  Removed num parents & children from tree bytes calculation for stable results
- Problem 4. Bug in balance checking in collectFee()
  - Changed to compare balance with gas cost total, not the gas amount

### DIFF
--- a/db/state-node.js
+++ b/db/state-node.js
@@ -63,6 +63,8 @@ class StateNode {
         that.treeBytes === this.treeBytes);
   }
 
+  // NOTE(liayoo): Bytes for some data (e.g. parents & children references, version) are excluded
+  //               from this calculation, since their sizes can vary and affect the gas costs and state proof hashes.
   computeNodeBytes() {
     return sizeof(this.isLeaf) +
         sizeof(this.value) +
--- a/db/state-node.js
+++ b/db/state-node.js
@@ -70,8 +70,7 @@ class StateNode {
         sizeof(this.proofHash) +
         sizeof(this.treeHeight) +
         sizeof(this.treeSize) +
-        sizeof(this.treeBytes) +
-        (this.numParents() + this.numChildren()) * JS_REF_SIZE_IN_BYTES;
+        sizeof(this.treeBytes);
   }
 
   static fromJsObject(obj, version) {
--- a/db/state-node.js
+++ b/db/state-node.js
@@ -4,7 +4,6 @@ const sizeof = require('object-sizeof');
 const CommonUtil = require('../common/common-util');
 const {
   HASH_DELIMITER,
-  JS_REF_SIZE_IN_BYTES,
   StateInfoProperties,
 } = require('../common/constants');
 
--- a/integration/node.test.js
+++ b/integration/node.test.js
@@ -883,7 +883,7 @@ describe('Blockchain Node', () => {
         assert.deepEqual(resultAfter, "some value with numbered nonce");
       })
 
-      it('set_value with failing operation', () => {
+      it('set_value with failing operation', async () => {
         // Check the original value.
         const resultBefore = parseOrLog(syncRequest(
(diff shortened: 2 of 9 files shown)


integration/node.test.js
```diff
@@ -3632,17 +3662,14 @@ describe('Blockchain Node', () => {
               },
               "state": {
                 "app": {
-                  "test": 1348
+                  "test": 1252 // 1044
```
platfowner, line 3665: 1044?

integration/node.test.js
```diff
@@ -4164,7 +4200,7 @@ describe('Blockchain Node', () => {
           valueBefore = parseOrLog(syncRequest('GET',
               server2 + `/get_value?ref=${valuePath}`).body.toString('utf-8')).result;
 
-          const body = parseOrLog(syncRequest('POST', server2 + '/set_value', {json: {
+          const body = parseOrLog(syncRequest('POST', server2 + '/set', {json: {
```
platfowner, line 4203: nice catch!

db/state-node.js
```diff
@@ -70,8 +70,7 @@ class StateNode {
         sizeof(this.proofHash) +
         sizeof(this.treeHeight) +
         sizeof(this.treeSize) +
-        sizeof(this.treeBytes) +
-        (this.numParents() + this.numChildren()) * JS_REF_SIZE_IN_BYTES;
```
platfowner, line 74: If this part affect consensus (i.e. different tree bytes value in state db with the same block number), this seems to be the only solution. If not e.g. with just different get timeing (i.e, different tree bytes values in state db with different block numbers), would it be a solution to erase that non-stable part from the value comparison in such specific test cases. (We already have some examples: https://github.com/ainblockchain/ain-blockchain/blob/develop/integration/node.test.js#L532  and  https://github.com/ainblockchain/ain-blockchain/blob/develop/integration/node.test.js#L718)
What do you think?

liayoo: Thanks for the idea! Unfortunately, it seems that the values (especially `numParents()`) vary every time I run the test. Even for the same node, the values differ when it creates the block & checks the block, making the state_proof_hash of the block invalid. I thought that it'd be better to increase the STATE_GAS_COEFFICIENT (if necessary) than to try to include the changing bytes.

platfowner: Ah, if it's more complicated issue (e.g. unstable tree_bytes -> unstable gas_cost), please let me know.

integration/node.test.js
liayoo: Ah, this was from testing.. 1252 is right. Removed the comment. Thanks!

liayoo:

db/state-node.js
liayoo: Yes, it would also affect gas costs.. Maybe we could use the (this.numParents() + this.numChildren()) * JS_REF_SIZE_IN_BYTES calculation only for information's sake? For example, for state usage APIs?

platfowner: I see. I think the consistency between the API and the gas budget calculation have merits.
Then could you add a NOTE why they're not included in tree bytes calculation?
And now JS_REF_SIZE_IN_BYTES can be removed from constants.js without the only use case?

platfowner (approved):

db/state-node.js
liayoo: Sounds good to me! Removed the const and added a note.

liayoo:

Issue #498, opened 2021-07-23
Need APIs for blockchain state utilization


platfowner:
We need blockchain APIs for state usage for users e.g.
- ain_getStateUsage(), /get_state_usage()
- params: app_name = ‘collaborative_ai’ 

and ain-js APIs as well.

liayoo:
APIs added: https://github.com/ainblockchain/ain-blockchain/pull/515

platfowner:
ain-js update is needed. @cshcomcom

platfowner:
ain-js updated: https://github.com/ainblockchain/ain-js/pull/38

PR #515, opened 2021-08-02
Add state utilization apis

liayoo:
- Added `ain_getStateUsage` json-rpc API
- Added `get_state_usage` client API

Related: https://github.com/ainblockchain/ain-blockchain/issues/498

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -134,8 +134,7 @@ app.get('/get_state_info', (req, res, next) => {
  * Returns the state usage of the given app.
  */
  app.get('/get_state_usage', (req, res, next) => {
-  const result = req.query.app_name
-      ? node.db.getStateUsageAtPath(`${PredefinedDbPaths.APPS}/${req.query.app_name}`) : null;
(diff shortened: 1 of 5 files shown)


client/index.js
```diff
@@ -129,6 +130,18 @@ app.get('/get_state_info', (req, res, next) => {
     .end();
 });
 
+/**
+ * Returns the state usage of the given app.
+ */
+ app.get('/get_state_usage', (req, res, next) => {
+  const result = req.query.app_name
+      ? node.db.getStateUsageAtPath(`${PredefinedDbPaths.APPS}/${req.query.app_name}`) : null;
```
platfowner, line 138: what about introducing a function in node/index.js to use both here and in ain_getStateUsage() for future consistency?

minor: could you move '?' operator to above line?

platfowner (approved):
Thanks!

client/index.js
liayoo: Great idea! Done!

liayoo:

PR #522, opened 2021-08-06
Bugfix/receipt bugs

liayoo:
- Fixed variable name typos in db
- Increased TREE_HEIGHT_LIMIT from 20 to 30
- Fixed removeOldReceipts() and node integration test by using getReceiptPath()


liayoo:
Some background about the TREE_HEIGHT_LIMIT increase - I was getting out of tree height errors since the receipt prefix layers have been added. If we don't have a function trigger call depth limit, we should consider adding one since we're now adding the function results to the state!

platfowner:
> Some background about the TREE_HEIGHT_LIMIT increase - I was getting out of tree height errors since the receipt prefix layers have been added. If we don't have a function trigger call depth limit, we should consider adding one since we're now adding the function results to the state!

Thanks for handling this. I was also looking into this issue, and realized that we're adding failed txs to blocks and recording receipts for them. What's happening is the block proposal tx also fails due to the tree height increase from the receipts written to the state db.
I think this could be a security hole as the attacker can send txs designed to fail by the tree height limit no matter what limit we pose.
One of the simplest solutions could be just making the receipts constant-length objects by removing other parts than the gas fee and gas cost. What do you think?

platfowner (approved):
LGTM
Thanks!

platfowner:
Let me merge this PR here to make things easier. :)
Let's continue the discussion though.

PR #528, opened 2021-08-11
Reduce flakiness of integration tests due to tree bytes interference

platfowner:
Change summary:
- Reduce flakiness of integration tests due to tree bytes interference

Background:
- The state db path interference between transactions can affect the gas fee calculation, which causes the flakiness of the test runs. By cleaning up the states used in the previous tests can help make the test cases independent each other.


liayoo (approved):
LGTM, thanks!

PR #533, opened 2021-08-17
Always deploy both genesis and incremental scripts

platfowner:
Change summary:
- Always deploy both genesis and incremental scripts


liayoo (approved):
LGTM, thanks!

PR #534, opened 2021-08-17
Set testnet chain id as 0

minsulee2:
As discussed, chain id for testnet is set to 0.


platfowner (approved):
LGTM
Thanks!

minsulee2:
@platfowner Thanks thanks!!!!!

Issue #519, opened 2021-08-05
Need internal radix tree implementation


platfowner:
It turned out that the recent increase of the cpu utilization is due to proof hash computation for state nodes of high branch factors: https://github.com/ainblockchain/ain-blockchain/issues/517

A general solution for this issue is to introduce an internal radix tree implementation: https://github.com/ainblockchain/ain-docs/blob/master/AIP_docs/AIP-007_Provable_Blockchain_States-20210512.pdf

platfowner:
Done.

Issue #517, opened 2021-08-03
Need to handle cpu usage increase issue (esp. receipt writing) 


platfowner:
We need to handle the cpu usage increase issue recently caused by receipt state writing.

platfowner:
Fixed.

PR #535, opened 2021-08-17
Implement radix tree

platfowner:
Change summary:
- Implement radix tree (w/ radix node)
- Add isHexString(), toHexString() util functions

Todos:
- Implement proof hash methods
- Integrate with StateNode (state-node.js) 

Related issues: 

### DIFF
--- a/db/radix-node.js
+++ b/db/radix-node.js
@@ -1,0 +1,134 @@
+const logger = require('../logger')('RADIX_NODE');
+
+class RadixNode {
+  constructor() {
+    this.stateNode = null;
+    this.labelRadix = '';
+    this.labelSuffix = '';
+    this.parent = null;
+    this.childMap = new Map();
+  }
+
+  getStateNode() {
+    return this.stateNode;
+  }
+
+  setStateNode(stateNode) {
+    this.stateNode = stateNode;
+  }
+
+  hasStateNode() {
+    return this.getStateNode() !== null;
+  }
+
+  resetStateNode() {
+    this.setStateNode(null);
+  }
+
+  getLabelRadix() {
+    return this.labelRadix;
+  }
+
+  setLabelRadix(labelRadix) {
+    this.labelRadix = labelRadix;
+  }
+
+  hasLabelRadix() {
+    return this.getLabelRadix() !== '';
+  }
+
+  resetLabelRadix() {
+    this.setLabelRadix('');
+  }
+
+  getLabelSuffix() {
+    return this.labelSuffix;
+  }
+
+  setLabelSuffix(labelSuffix) {
+    this.labelSuffix = labelSuffix;
+  }
+
+  hasLabelSuffix() {
+    return this.getLabelSuffix() !== '';
+  }
+
+  resetLabelSuffix() {
+    this.setLabelSuffix('');
+  }
+
+  getParent() {
+    return this.parent;
+  }
+
+  setParent(parent) {
+    this.parent = parent;
+  }
+
+  hasParent() {
+    return this.getParent() !== null;
+  }
+
+  resetParent() {
+    this.setParent(null);
+  }
+
+  getChild(labelRadix) {
+    const child = this.childMap.get(labelRadix);
+    if (child === undefined) {
+      return null;
+    }
+    return child;
+  }
+
+  setChild(labelRadix, labelSuffix, child) {
+    const LOG_HEADER = 'setChild';
+    if (this.hasChild(labelRadix)) {
+      logger.error(
+          `[${LOG_HEADER}] Overwriting a child with radix label ${labelRadix}.`);
+      // Does nothing.
+      return false;
+    }
+    this.childMap.set(labelRadix, child);
+    child.setLabelRadix(labelRadix);
+    child.setLabelSuffix(labelSuffix);
+    child.setParent(this);
+    return true;
+  }
+
+  hasChild(labelRadix = null) {
+    if (!labelRadix) {
+      return this.numChildren() > 0;
+    }
+    return this.childMap.has(labelRadix);
+  }
+
+  deleteChild(labelRadix) {
+    const LOG_HEADER = 'deleteChild';
+    if (!this.hasChild(labelRadix)) {
+      logger.error(`[${LOG_HEADER}] Deleting a non-existing child with label: ${labelRadix}.`);
+      // Does nothing.
+      return false;
+    }
+    const child = this.getChild(labelRadix);
+    this.childMap.delete(labelRadix);
+    child.resetLabelRadix();
+    child.resetLabelSuffix();
+    child.resetParent();
+    return true;
+  }
+
+  getChildLabelRadices() {
+    return [...this.childMap.keys()];
+  }
+
+  getChildNodes() {
+    return [...this.childMap.values()];
+  }
+
+  numChildren() {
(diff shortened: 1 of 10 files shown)


db/radix-tree.js
```diff
@@ -0,0 +1,241 @@
...
+
+      // Case 3: Has a child with matching label suffix.
+      curNode = child;
+      labelIndex += 1 + child.getLabelSuffix().length;
+    }
+
+    return curNode;
+  }
+
+  getFromTree(label) {
```
liayoo, line 101: suggestion) how about a note that says it's getting the StateNode? or maybe include it in the function name, like getStateNodeFromTree?

db/radix-node.js
```diff
@@ -0,0 +1,142 @@
+const logger = require('../logger')('RADIX_NODE');
+
+class RadixNode {
+  constructor() {
+    this.stateNode = null;
+    this.labelRadix = '';
+    this.labelSuffix = '';
+    this.parent = null;
+    this.childMap = new Map();
```
liayoo, line 9: Could you make it a little clearer that this is a childMap composed of RadixNode?

db/radix-tree.js
```diff
@@ -0,0 +1,241 @@
+const logger = require('../logger')('RADIX_TREE');
+
+const CommonUtil = require('../common/common-util');
+const RadixNode = require('./radix-node');
+
+class RadixTree {
+  constructor() {
+    this.root = new RadixNode();
+    this.childMap = new Map();
```
liayoo, line 9: Could you make it a little clearer that this is a childMap composed of StateNode?

db/radix-tree.js
```diff
@@ -0,0 +1,241 @@
...
+  getFromTree(label) {
+    const hexLabel = RadixTree.toHexLabel(label);
+    const node = this.getNodeForReading(hexLabel);
+    if (node === null) {
+      return null;
+    }
+    return node.getStateNode();
+  }
+
+  getFromMap(label) {
```
liayoo, line 110: ditto

liayoo (commented):
Thanks for the nice implementation! It seems a little ambiguous at times whether a node / child is a StateNode or a RadixNode, so I left a couple of notes on that.. Let me know if I misunderstood anything!

db/radix-node.js
platfowner: Renamed it: -> 'radixChildMap'. PTAL~

db/radix-tree.js
platfowner: Renamed: -> 'stateNodeMap'. PTAL~

db/radix-tree.js
platfowner: All external methods of RadixTree is about state nodes. Addressed it with:
- Added underscore (_) prefix to all internal methods
- Rename methods: '_getRadixNodeForReading', '_getRadixNodeForWriting'
- Add jsdoc comment to RadixTree class

PTAL~

db/radix-tree.js
platfowner: ditto~

platfowner:
All comments addressed, PTAL~ @liayoo

liayoo:
Thanks for the updates! Looks good! 👍 I just have one more question about the design - how do the radix trees fit in with StateManager? Will it be changed to manage radix trees instead of state node trees?

platfowner:
> Thanks for the updates! Looks good! 👍 I just have one more question about the design - how do the radix trees fit in with StateManager? Will it be changed to manage radix trees instead of state node trees?

My idea is to replace StateNode.childMap with RadixTree (this is why RadixTree has 1-to-1 mapped APIs with Map()) as described in AIP-025, so each state node will have one instance of RadixTree, which means they won't be managed in a center (and they have nothing to do with state versions, which is managed by StateManager - I have a plan to give a new name StateVersionManager).

liayoo:
> > Thanks for the updates! Looks good! 👍 I just have one more question about the design - how do the radix trees fit in with StateManager? Will it be changed to manage radix trees instead of state node trees?
> 
> My idea is to replace StateNode.childMap with RadixTree (this is why RadixTree has 1-to-1 mapped APIs with Map()) as described in AIP-025, so each state node will have one instance of RadixTree, which means they won't be managed in a center (and they have nothing to do with state versions, which is managed by StateManager - I have a plan to give a new name StateVersionManager).

Ah, now I get it! That's brilliant. Thanks for the explanation!

db/radix-tree.js
```diff
@@ -0,0 +1,246 @@
...
+    // 1. Set in the radix tree.
+    this._setInTree(label, stateNode);
+    // 2. Set in the hash map.
+    this._setInMap(label, stateNode);
+  }
+
+  _hasInTree(label) {
+    const hexLabel = RadixTree._toHexLabel(label);
+    const node = this._getRadixNodeForReading(hexLabel);
+    if (node === null) {
```
liayoo, line 148: minor) how about `return node !== null` ?

db/radix-tree.js
```diff
@@ -0,0 +1,246 @@
...
+    const hexLabel = RadixTree._toHexLabel(label);
+    const node = this._getRadixNodeForReading(hexLabel);
+    if (node === null || !node.hasStateNode()) {
+      logger.error(`[${LOG_HEADER}] Deleting a non-existing child with label: ${label}.`);
+      // Does nothing.
+      return false;
+    }
+    node.resetStateNode();
+    if (node.numChildren() === 1 && node.hasParent()) {
+      this._mergeToChild(node);
```
liayoo, line 203: How about return this._mergeToChild(node) ?

db/radix-tree.js
```diff
@@ -0,0 +1,246 @@
...
+    } else if (node.numChildren() === 0) {
+      if (!node.hasParent()) {
+        logger.error(`[${LOG_HEADER}] Deleting a child without parent with label: ${label}.`);
+        // Does nothing.
+        return false;
+      }
+      const parent = node.getParent();
+      parent.deleteChild(node.getLabelRadix());
+      if (parent.numChildren() === 1 && !parent.hasStateNode() && parent.hasParent()) {
+        this._mergeToChild(parent);
```
liayoo, line 213: ditto

unittest/rule-util.test.js
```diff
@@ -171,14 +171,44 @@ describe("RuleUtil", () => {
...
+
   describe("isValidHash", () => {
     it("when invalid input", () => {
       expect(util.isValidHash(true)).to.equal(false);
       expect(util.isValidHash(false)).to.equal(false);
       expect(util.isValidHash(0)).to.equal(false);
       expect(util.isValidHash(10)).to.equal(false);
-      expect(util.isDict(null)).to.equal(false);
-      expect(util.isDict(undefined)).to.equal(false);
+      expect(util.isValidHash(null)).to.equal(false);
```
liayoo, line 210: Oops! Thanks for catching this!

liayoo (approved):

db/radix-tree.js
platfowner: Done, thanks!

db/radix-tree.js
platfowner: Done, thanks!

db/radix-tree.js
platfowner: ditto~

PR #539, opened 2021-08-18
Add shell scripts for restarting servers in GCP

platfowner:
Change summary:
- Use bash shell as the default (some scripts have dependency on bash-only syntax)
- Add shell scripts for restarting servers in GCP
- Fix broken integration tests by increasing finalization wait time


liayoo (approved):
LGTM, thanks!

PR #540, opened 2021-08-19
Update consensus reward distribution logic.

liayoo:
- Now tx fees are split across the proposer and the validators
- New consensus path has been introduced to keep track of unclaimed & cumulative rewards (`/consensus/rewards/${address}: { unclaimed, cumulative }`)
- Added _claimReward native function
  - Triggered by setting a value at `/gas_fee/claim/${address}/${key}: { amount }`
- DB structure updated: `/gas_fee/collect/${address}/${blockNumber}/${txHash}` => `/gas_fee/collect/${blockNumber}/${address}/${txHash}`
- Fixed trimExecutionResult() for multi-set txs

### DIFF
--- a/common/constants.js
+++ b/common/constants.js
@@ -154,19 +154,19 @@ const PredefinedDbPaths = {
   DOT_SHARD: '.shard',
   // Consensus
   CONSENSUS: 'consensus',
-  WHITELIST: 'whitelist',
-  NUMBER: 'number',
-  PROPOSE: 'propose',
-  PROPOSER: 'proposer',
-  VALIDATORS: 'validators',
-  TOTAL_AT_STAKE: 'total_at_stake',
-  VOTE: 'vote',
-  BLOCK_HASH: 'block_hash',
-  STAKE: 'stake',
-  PROPOSAL_RIGHT: 'proposal_right',
-  REWARDS: 'rewards',
-  REWARDS_UNCLAIMED: 'unclaimed',
-  REWARDS_CUMULATIVE: 'cumulative',
+  CONSENSUS_BLOCK_HASH: 'block_hash',
+  CONSENSUS_NUMBER: 'number',
+  CONSENSUS_PROPOSAL_RIGHT: 'proposal_right',
+  CONSENSUS_PROPOSE: 'propose',
+  CONSENSUS_PROPOSER: 'proposer',
+  CONSENSUS_REWARDS: 'rewards',
+  CONSENSUS_REWARDS_UNCLAIMED: 'unclaimed',
+  CONSENSUS_REWARDS_CUMULATIVE: 'cumulative',
+  CONSENSUS_STAKE: 'stake',
+  CONSENSUS_TOTAL_AT_STAKE: 'total_at_stake',
+  CONSENSUS_VALIDATORS: 'validators',
+  CONSENSUS_VOTE: 'vote',
+  CONSENSUS_WHITELIST: 'whitelist',
   // Receipts
   RECEIPTS: 'receipts',
   RECEIPTS_ADDRESS: 'address',
@@ -180,10 +180,10 @@ const PredefinedDbPaths = {
   RECEIPTS_EXEC_RESULT_RESULT_LIST: 'result_list',
   // Gas fee
   GAS_FEE: 'gas_fee',
-  CLAIM: 'claim',
-  COLLECT: 'collect',
-  BILLING: 'billing',
   GAS_FEE_AMOUNT: 'amount',
+  GAS_FEE_BILLING: 'billing',
+  GAS_FEE_CLAIM: 'claim',
+  GAS_FEE_COLLECT: 'collect',
   GAS_FEE_UNCLAIMED: 'unclaimed',
   // Token
   TOKEN: 'token',
@@ -570,7 +570,7 @@ function isServiceType(type) {
  * Service types allowed to create service accounts.
  */
 const SERVICE_ACCOUNT_SERVICE_TYPES = [
-  PredefinedDbPaths.BILLING,
+  PredefinedDbPaths.GAS_FEE_BILLING,
   PredefinedDbPaths.ESCROW,
   PredefinedDbPaths.GAS_FEE,
   PredefinedDbPaths.PAYMENTS,
@@ -627,8 +627,8 @@ function overwriteGenesisParams(overwritingParams, type) {
       const addr = GenesisAccounts[AccountProperties.OTHERS][i][AccountProperties.ADDRESS];
       CommonUtil.setJsObject(whitelist, [addr], true);
       CommonUtil.setJsObject(validators, [addr], {
-          [PredefinedDbPaths.STAKE]: GenesisParams.consensus.MIN_STAKE_PER_VALIDATOR,
-          [PredefinedDbPaths.PROPOSAL_RIGHT]: true
+          [PredefinedDbPaths.CONSENSUS_STAKE]: GenesisParams.consensus.MIN_STAKE_PER_VALIDATOR,
+          [PredefinedDbPaths.CONSENSUS_PROPOSAL_RIGHT]: true
         });
     }
     GenesisParams.consensus.GENESIS_WHITELIST = whitelist;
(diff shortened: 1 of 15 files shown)


common/path-util.js
```diff
@@ -256,6 +256,22 @@ class PathUtil {
...
+  static getConsensusRewardsUnclaimedPath(address) {
+    return CommonUtil.formatPath([PredefinedDbPaths.CONSENSUS, PredefinedDbPaths.REWARDS, address, PredefinedDbPaths.REWARDS_UNCLAIMED]);
+  }
+
+  static getConsensusRewardsCumulativePath(address) {
+    return CommonUtil.formatPath([PredefinedDbPaths.CONSENSUS, PredefinedDbPaths.REWARDS, address, PredefinedDbPaths.REWARDS_CUMULATIVE]);
+  }
+
+  static getConsensusNumberPath(blockNumber) {
+    return CommonUtil.formatPath([PredefinedDbPaths.CONSENSUS, PredefinedDbPaths.NUMBER, blockNumber]);
```
platfowner, line 272: minor: what about adding prefix 'CONSENSUS_' to the names of the consensus-related label constants? e.g. CONSENSUS_NUMBER, CONSENSUS_WHITELIST, ...

common/constants.js
```diff
@@ -174,11 +177,14 @@ const PredefinedDbPaths = {
   RECEIPTS_EXEC_RESULT_ERROR_MESSAGE: 'error_message',
   RECEIPTS_EXEC_RESULT_GAS_AMOUNT_CHARGED: 'gas_amount_charged',
   RECEIPTS_EXEC_RESULT_GAS_COST_TOTAL: 'gas_cost_total',
+  RECEIPTS_EXEC_RESULT_RESULT_LIST: 'result_list',
   // Gas fee
   GAS_FEE: 'gas_fee',
+  CLAIM: 'claim',
```
platfowner, line 183: minor: what about adding prefix 'GAS_FEE_' to the gas-fee-related const names? e.g. GAS_FEE_CLAIM, GAS_FEE_COLLECT, ...

platfowner: And it might help reduce ambiguity between claim reward and claim (payment)?

common/path-util.js
```diff
@@ -256,6 +256,22 @@ class PathUtil {
     return CommonUtil.appendPath(accountPath, PredefinedDbPaths.BALANCE)
   }
 
+  static getConsensusRewardsPath(address) {
+    return CommonUtil.formatPath([PredefinedDbPaths.CONSENSUS, PredefinedDbPaths.REWARDS, address]);
+  }
+
+  static getConsensusRewardsUnclaimedPath(address) {
+    return CommonUtil.formatPath([PredefinedDbPaths.CONSENSUS, PredefinedDbPaths.REWARDS, address, PredefinedDbPaths.REWARDS_UNCLAIMED]);
```
platfowner, line 264: question: Is this path handled as a service account path, i.e., all the permission are managed via service account framework?
If so, what about the rewards cumulative path below? Is this managed as a service account transferable?

db/functions.js
```diff
@@ -625,23 +626,59 @@ class Functions {
...
+      [PredefinedDbPaths.REWARDS_UNCLAIMED]: (prevRewards[PredefinedDbPaths.REWARDS_UNCLAIMED] || 0) + amount,
+      [PredefinedDbPaths.REWARDS_CUMULATIVE]: (prevRewards[PredefinedDbPaths.REWARDS_CUMULATIVE] || 0) + amount
+    }, context);
+  }
+
   _distributeFee(value, context) {
     const blockNumber = context.params.number;
-    const gasCostTotal = value.gas_cost_total;
-    const proposer = value.proposer;
+    if (blockNumber <= 1) {
```
platfowner, line 640: Question: Does this mean the genesis block + 1?

db/functions.js
```diff
@@ -625,23 +626,59 @@ class Functions {
...
+    const lastConsensusRound = this.db.getValue(PathUtil.getConsensusNumberPath(blockNumber - 1));
+    const gasCostTotal = lastConsensusRound.propose.gas_cost_total;
     if (gasCostTotal <= 0) {
       return this.returnFuncResult(context, FunctionResultCode.SUCCESS);
     }
+    const proposer = lastConsensusRound.propose.proposer;
+    const totalAtStake = Object.values(lastConsensusRound.vote).reduce((acc, cur) => acc + cur.stake, 0);
+    const validators = Object.keys(lastConsensusRound.vote);
+    const proposerReward = gasCostTotal / 2;
+    const validatorRewardTotal = gasCostTotal - proposerReward;
```
platfowner, line 652: I think using this equation instead of 'gasCostTotal / 2' might be helpful to minimize the floating-point error accumulation.

db/functions.js
```diff
@@ -625,23 +626,59 @@ class Functions {
...
+    this.incrementConsensusRewards(proposer, proposerReward, context);
+    for (const validator of validators) {
+      const validatorStake = lastConsensusRound.vote[validator].stake;
+      this.incrementConsensusRewards(
+          validator, validatorRewardTotal * (validatorStake / totalAtStake), context);
```
platfowner, line 657: Could a similar equation to above be applied to the last validator, e.g. `const validatorReward = validatorRewardTotal - sum`?

db/rule-util.js
```diff
@@ -274,6 +274,21 @@ class RuleUtil {
         (data.response.status === FunctionResultCode.SUCCESS ||
         data.response.status === FunctionResultCode.FAILURE);
   }
+
+  validateClaimRewardData(userAddr, data, getValue) {
+    const { PredefinedDbPaths } = require('../common/constants');
+    if (!this.isDict(data) || !this.isNumber(data.amount) || data.amount <= 0) {
+      return false;
+    }
+    const unclaimed = getValue(`/${PredefinedDbPaths.CONSENSUS}/${PredefinedDbPaths.REWARDS}/` +
```
platfowner, line 283: minor: Could getConsensusRewardsUnclaimedPath() be reused? e.g.
```
  const PathUtil = require('../common/path-util');
  ... = PathUtil.getConsensusRewardsUnclaimedPath(userAddr);
```

db/functions.js
```diff
@@ -625,23 +626,59 @@ class Functions {
     }
   }
 
+  incrementConsensusRewards(address, amount, context) {
+    const rewardsPath = PathUtil.getConsensusRewardsPath(address);
+    const prevRewards = this.db.getValue(rewardsPath) || {};
+    return this.setValueOrLog(rewardsPath, {
```
platfowner, line 632: _distributeFee() used to be using setServiceAccountTransferOrLog() as it's a kind of our principle to apply 'transfer' to all token moves between individual or service accounts? Doesn't it also apply to the unclaimed service account?
For the cumulative value case, however I think we can change this to a non-service-account value (by applying a non-service-account path pattern?).
The core criteria between a service account and a non-service-account value I think is whether it's a circulated amount or not?

common/path-util.js
platfowner: I guess the unclaimed value is a service account while the cumulative value is not. Let me continue this discussion in the specific code snippet below.

db/functions.js
platfowner: Ah, just realized that both the unclaimed and the cumulative are not account balances as values are still in the gas fee service account. (sorry for my confusion).

platfowner: Looks totally fine, thanks!

db/functions.js
liayoo: Right, because we need to have the votes in order to determine which validators to give the rewards to, we're distributing the rewards from the (N-1)th block when a proposal for the Nth block is written. Genesis block doesn't have rewards, so we can start from block number 2 (= processing block number 1) and so on.

common/path-util.js
liayoo: Good idea! Added CONSENSUS_ and GAS_FEE_ prefixes. Thanks!

common/constants.js
liayoo: Agreed! Done, thanks!

db/functions.js
liayoo: Done!

db/rule-util.js
liayoo: Done! Updated other functions as well~

liayoo:

db/functions.js
platfowner: I see. Thanks for the explanation, which would be worth leaving as a note?

platfowner (approved):
LGTM

db/functions.js
liayoo: Done!

liayoo:
Thanks for the review~ Now merging..

PR #541, opened 2021-08-19
Add set / update / verify proof hash APIs

platfowner:
Change summary:
- Add set / update / verify proof hash APIs for radix tree
- Handle invalid state node input

Related issues: 

### DIFF
--- a/db/radix-node.js
+++ b/db/radix-node.js
@@ -1,5 +1,7 @@
 const logger = require('../logger')('RADIX_NODE');
 
+const StateNode = require('./state-node');
+
 /**
  * Implements a radix node, which is used as a component of RadixTree.
  */
@@ -10,6 +12,7 @@ class RadixNode {
     this.labelSuffix = '';
     this.parent = null;
     this.radixChildMap = new Map();
+    this.proofHash = null;
   }
 
   getStateNode() {
@@ -17,7 +20,15 @@ class RadixNode {
   }
 
   setStateNode(stateNode) {
+    const LOG_HEADER = 'setStateNode';
+    if (!(stateNode instanceof StateNode)) {
+      logger.error(
+          `[${LOG_HEADER}] Setting with a non-StateNode instance.`);
+      // Does nothing.
+      return false;
+    }
     this.stateNode = stateNode;
+    return true;
   }
 
   hasStateNode() {
@@ -25,7 +36,7 @@ class RadixNode {
   }
 
   resetStateNode() {
-    this.setStateNode(null);
+    this.stateNode = null;
   }
 
   getLabelRadix() {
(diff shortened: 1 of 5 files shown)


db/radix-node.js
```diff
@@ -133,6 +148,79 @@ class RadixNode {
...
+  resetProofHash() {
+    this.setProofHash(null);
+  }
+
+  _buildProofHash() {
+    let preimage = '';
+    if (this.hasStateNode()) {
+      preimage = this.getStateNode().getProofHash();
+    }
+    preimage += `${HASH_DELIMITER}${HASH_DELIMITER}`;
```
liayoo, line 172: question) do we have 2 hash delimiters concatenated so that it's distinguished from the StateNode's proof hash?

db/radix-node.js
```diff
@@ -133,6 +148,79 @@ class RadixNode {
...
+    preimage += this.getChildNodes().map((child) => {
+      const childLabel = `${child.getLabelRadix()}${child.getLabelSuffix()}`;
```
liayoo, line 174: suggestion) how about we have a function for this? (e.g. getLabel())

db/radix-node.js
platfowner: Exactly, in the contrast to the state tree, radix tree could have ambiguity between a proof hash from the state node and one of the same value from the first radix child node.

db/radix-node.js
platfowner: Just done, thanks!

platfowner:
@liayoo  PTAL at the addressed comments~

db/radix-node.js
```diff
@@ -133,6 +152,78 @@ class RadixNode {
...
+
+  _buildProofHash() {
+    let preimage = '';
+    if (this.hasStateNode()) {
+      preimage = this.getStateNode().getProofHash();
+    }
+    preimage += `${HASH_DELIMITER}${HASH_DELIMITER}`;
+    preimage += this.getChildNodes().map((child) => {
+      return `${child.getLabel()}${HASH_DELIMITER}${child.getProofHash()}`;
+    }, '').join(HASH_DELIMITER);
```
liayoo, line 179: the second argument '' can be removed?

liayoo (approved):

db/radix-node.js
liayoo: Nice! Thanks for the explanation :)

db/radix-node.js
platfowner: Right, I guess it's redundant code mistakenly added by some confusion with 'reduce' function.
Just fixed. Thanks!

platfowner:
Thanks for the quick review. Let me move on to the final state: integration with state node.

PR #542, opened 2021-08-24
Use radix tree proof hashes in blockchain by default

platfowner:
Change summary:
- Use radix tree proof hashes in blockchain by default
- Turn off enableReceiptPathPrefixLayers flag by default (will be removed later) 

Related issues: https://github.com/ainblockchain/ain-blockchain/pull/535


db/state-node.js
```diff
@@ -290,20 +364,51 @@ class StateNode {
...
+  buildProofHash(updatedChildLabel = null) {
     let preimage;
     if (this.getIsLeaf()) {
       preimage = this.getValue();
     } else {
-      preimage = this.getChildLabels().map((label) => {
-        return `${label}${HASH_DELIMITER}${this.getChild(label).getProofHash()}`;
-      }).join(HASH_DELIMITER);
+      if (FeatureFlags.enableRadixTreeLayers) {
+        if (updatedChildLabel === null) {
```
liayoo, line 374: could you add some notes on what `updatedChildLabel === null` signifies?

PR #545, opened 2021-08-29
Feature/pos penalties

liayoo:
- Added penalties to invalid block proposers
- Votes
  - Regular votes = voting for a valid block
  - Evidence votes = voting against an invalid block
  - /conensus/number/${blockNumber}/${blockHash}/vote/${address} { block_hash, stake, is_against, offense_type }
- Proposals
  - `offenses` field introduced that counts how many offenses the address has made per type ({ [address]: { [offense_type]: number } })
- Blocks
  - `evidence` field introduced ({ [address]: [{ offense_type, transactions, block, votes }] })
- `_blacklist` native function
  - updates `/consensus/blacklist` and stake lockup duration upon a proposal with the `offenses` field


client/index.js
```diff
@@ -281,6 +281,21 @@ if (ENABLE_DEV_SET_CLIENT_API) {
...
+
+  app.post('/broadcast_consensus_msg', (req, res) => {
+    p2pClient.broadcastConsensusMessage(req.body);
+    res.status(200)
+      .set('Content-Type', 'application/json')
+      .send({ code: 0, result: true })
+      .end();
+  });
+
+  app.post('/sign_transaction', (req, res) => {
```
platfowner, line 293: nice utility API!

db/functions.js
```diff
@@ -701,21 +723,22 @@ class Functions {
     const executedAt = context.executedAt;
     const resultPath = PathUtil.getStakingStakeResultPath(serviceName, user, stakingKey, recordId);
     const expirationPath = PathUtil.getStakingExpirationPath(serviceName, user, stakingKey);
+    const currentExpiration = Math.max(Number(this.db.getValue(expirationPath)), timestamp);
```
platfowner, line 726: It seems there is some logical change here. Could you share some background? 
In the case getValue(expirationPath) > timestamp, the new expiration will be getValue(expirationPath) + Number(lockup), while it was timestamp + Number(lockup).

platfowner: Ah, it's from the penalty logic.
What about the following?
- if getValue(expirationPath) > timestamp + Number(lockup), then just keep using getValue(expirationPath)
- otherwise, use timestamp + Number(lockup)

This can make sure the new lockup period always longer than Number(lockup) avoiding to increase the period all the time for new staking events?

db/functions.js
```diff
@@ -692,6 +697,23 @@ class Functions {
...
+  _blacklist(value, context) {
+    if (CommonUtil.isEmpty(value.offenses)) {
+      return this.returnFuncResult(context, FunctionResultCode.SUCCESS);
+    }
+    for (const [offender, offenseList] of Object.entries(value.offenses)) {
+      let numOffenses = Object.values(offenseList).reduce((acc, cur) => acc + cur, 0);
+      const blacklistPath = PathUtil.getConsensusBlacklistAddrPath(offender);
+      this.incValueOrLog(blacklistPath, numOffenses, context);
+      numOffenses = this.db.getValue(blacklistPath); // new # of offenses
+      const lockupExtension = ConsensusConsts.STAKE_LOCKUP_EXTENSION * Math.pow(2, numOffenses - 1);
```
platfowner, line 709: `(Math.pow(2, numOffenses) - 1)` for the case of numOffenses === 0?

platfowner: Ah, it looks like 2^(n - 1) comes from the summation of a geometric series (S_n - S_{n - 1)). If we want this code to cover multiple new offenses from `offenseList`, the lockupExtension computation logic can be made stable by tweaking a little bit:
- For 1th, 2nd, 3rd, ..., nth offenses, we want to give extra lockup period of EXT * 1, EXT * 2, EXT * 2^2, ..., EXT * 2^{n - 1}, respectively.
- So we can run a loop for the new offenses, e.g. nth, (n + 1)th, ..., (n + k - 1)th, summing EXT * 2^{n - 1} + EXT * 2^n + ... + EXT * 2^{n + k - 2}
?

db/functions.js
liayoo: Great idea! Updated the logic. Thanks!!

consensus/index.js
```diff
@@ -303,6 +315,40 @@ class Consensus {
...
     const lastBlockInfo = this.blockPool.hashToBlockInfo[lastBlock.hash];
     logger.debug(`[${LOG_HEADER}] lastBlockInfo: ${JSON.stringify(lastBlockInfo, null, 2)}`);
+
+    const myAddr = this.node.account.address;
+    // Need the block #1 to be finalized to have the stakes reflected in the state
+    let validators = {};
+    if (this.node.bc.lastBlockNumber() < 1) {
+      const whitelist = GENESIS_WHITELIST;
+      for (const address in whitelist) {
+        if (whitelist[address] === true) {
```
platfowner, line 325: question: are we considering whitelist[address] = false case as well?

db/functions.js
liayoo: Oops, you're absolutely right! Thanks for catching this. Fixed the logic, PTAL.

consensus/index.js
```diff
@@ -303,6 +315,40 @@ class Consensus {
...
+          const stake = tempDb.getValue(PathUtil.getConsensusStakingAccountBalancePath(address));
+          if (stake && MIN_STAKE_PER_VALIDATOR <= stake && stake <= MAX_STAKE_PER_VALIDATOR) {
```
platfowner, line 327: minor: what about `stake >= MIN_STAKE_PER_VALIDATOR`?

consensus/block-pool.js
```diff
@@ -387,37 +414,63 @@ class BlockPool {
...
+    const blockNumber = baseDb.blockNumberSnapshot;
+    let backupDb = this.node.createTempDb(
+        baseDb.stateVersion, `${StateVersions.SNAP}:${blockNumber}`, blockNumber);
+    const majority = totalAtStake * ConsensusConsts.MAJORITY;
+    const evidence = {};
+    const offenses = {};
+    Object.values(this.hashToBlockInfo).forEach((blockInfo) => {
+      if (!blockInfo.against_votes || !blockInfo.against_votes.length || !blockInfo.block ||
+          !blockInfo.proposal) {
+        return;
```
platfowner, line 431: what

platfowner, line 431: continue?

consensus/block-pool.js
```diff
@@ -387,37 +414,63 @@ class BlockPool {
...
+    const totalAtStake = Object.values(validators).reduce((acc, cur) => {
+      return acc + _get(cur, 'stake', 0);
+    }, 0);
+    const blockNumber = baseDb.blockNumberSnapshot;
+    let backupDb = this.node.createTempDb(
+        baseDb.stateVersion, `${StateVersions.SNAP}:${blockNumber}`, blockNumber);
+    const majority = totalAtStake * ConsensusConsts.MAJORITY;
+    const evidence = {};
+    const offenses = {};
+    Object.values(this.hashToBlockInfo).forEach((blockInfo) => {
```
platfowner, line 428: question: in each block proposal, do we need to run this loop for all blocks in hashToBlockInfo?

consensus/block-pool.js
```diff
@@ -387,37 +414,63 @@ class BlockPool {
...
 
-  static filterProposal(votes) {
-    if (!votes) return null;
-    const proposalSuffix = 'propose';
-    const proposal = votes.filter((tx) => {
-      if (tx.tx_body.operation.type === WriteDbOperations.SET_VALUE) {
-        return tx.tx_body.operation.ref.endsWith(proposalSuffix);
-      } else if (tx.tx_body.operation.type === WriteDbOperations.SET) {
-        return tx.tx_body.operation.op_list[0].ref.endsWith(proposalSuffix);
+  getOffensesAndEvidence(validators, baseDb) {
```
platfowner, line 417: could you leave some js doc comment about what this function does?

consensus/index.js
liayoo: Currently, we're not. It either exists in the whitelist or it doesn't.

consensus/index.js
liayoo: Do you mean that we should remove the upper limit on validators' stakes?

consensus/block-pool.js
liayoo: Here we're using `forEach`, so the return will only stop the execution of the particular callback function, as a continue would in `for...of` or regular for loop.

consensus/block-pool.js
liayoo: Hmm.. yes it would. Come to think of it, having a separate mapping just for invalid blocks seems to be a better approach..!

consensus/block-pool.js
platfowner: I see. thanks for the explanation. what about using 'for...of' instead?

consensus/block-pool.js
```diff
@@ -387,37 +414,63 @@ class BlockPool {
...
+    const LOG_HEADER = 'getOffensesAndEvidence';
+    const totalAtStake = Object.values(validators).reduce((acc, cur) => {
+      return acc + _get(cur, 'stake', 0);
+    }, 0);
+    const blockNumber = baseDb.blockNumberSnapshot;
+    let backupDb = this.node.createTempDb(
```
platfowner, line 423: question: aren't backupDb() and restoreDb() applicable in this case?

consensus/consensus-util.js
```diff
@@ -0,0 +1,73 @@
...
+
+class ConsensusUtil {
+  static isValidConsensusTx(tx) {
+    const op = _get(tx, 'tx_body.operation');
+    if (!op) return false;
+    const consensusTxPrefix = CommonUtil.formatPath(
+        [PredefinedDbPaths.CONSENSUS, PredefinedDbPaths.CONSENSUS_NUMBER]);
+    if (op.type === WriteDbOperations.SET_VALUE) {
+      return op.ref.startsWith(consensusTxPrefix);
+    } else if (op.type === WriteDbOperations.SET) {
```
platfowner, line 13: question: If we don't have actual cases of consensus txs with SET operation, can this part be just dropped leaving some notes?

consensus/block-pool.js
liayoo: Right. backupDb() and restoreDb() are used for the executeTransaction() call, since we're setting restoreIfFails = true. However, we also need to restore when the votes have been successfully executed but the sum of their stakes haven't met the majority..

consensus/index.js
platfowner: Oh, no. I mean just the order of the expression components '(this) >= (that)'.

consensus/consensus-util.js
platfowner: it's basically up to you even if it's the case.

consensus/index.js
platfowner: I see, then is the if-block redundant as every 'address' comes from 'whitelist'?

db/functions.js
```diff
@@ -692,6 +697,34 @@ class Functions {
     return this.returnFuncResult(context, FunctionResultCode.SUCCESS);
   }
 
+  static getBlacklistLockupExtension(numNewOffenses, updatedNumOffenses) {
+    let n = updatedNumOffenses - numNewOffenses + 1;
```
platfowner, line 701: minor: can be just added into the for-loop?

db/functions.js
```diff
@@ -701,21 +734,28 @@ class Functions {
     const executedAt = context.executedAt;
     const resultPath = PathUtil.getStakingStakeResultPath(serviceName, user, stakingKey, recordId);
     const expirationPath = PathUtil.getStakingExpirationPath(serviceName, user, stakingKey);
-    const lockup = this.db.getValue(PathUtil.getStakingLockupDurationPath(serviceName));
+    const currentExpiration = Number(this.db.getValue(expirationPath));
+    const lockup = Number(this.db.getValue(PathUtil.getStakingLockupDurationPath(serviceName)));
+    const newExpiration = timestamp + lockup;
+    const updateExpiration = currentExpiration <= newExpiration;
```
platfowner, line 740: minor: `newExpiration > currentExpiration ` (without '=')?

consensus/index.js
```diff
@@ -303,6 +315,40 @@ class Consensus {
     }
     const lastBlockInfo = this.blockPool.hashToBlockInfo[lastBlock.hash];
     logger.debug(`[${LOG_HEADER}] lastBlockInfo: ${JSON.stringify(lastBlockInfo, null, 2)}`);
+
```
platfowner, line 318: Could you consider breaking this big function (createProposal()) into pieces somehow for readability?

consensus/index.js
```diff
@@ -431,30 +446,30 @@ class Consensus {
     const block = Block.parse(proposalBlock);
     if (!block) {
       logger.error(`[${LOG_HEADER}] Invalid block: ${JSON.stringify(proposalBlock)}`);
-      return false;
+      return { voteAgainst: false };
```
platfowner, line 449: Could you consider breaking this huge function (checkProposal()) into pieces somehow for readability?

platfowner, line 449: Could you leave some js doc comments about the return value structure?

consensus/index.js
```diff
@@ -475,7 +490,7 @@ class Consensus {
     if (number !== 1 && (!prevBlockInfo || !prevBlockInfo.block)) {
       logger.debug(`[${LOG_HEADER}] No notarized block at number ${number - 1} with ` +
           `hash ${last_hash}`);
-      return;
+      return { voteAgainst: false };
     }
     const prevBlock = number > 1 ? prevBlockInfo.block : prevBlockInfo;
```
platfowner, line 495: What about having prefBlockInfo always have similar structure? e.g. 
prevBlockInfo = number === 1 ? { block: this.node.bc.getBockByNumber(0), ... } : this.blockPool.hashToBlockInfo[last_hash];

consensus/index.js
```diff
@@ -475,7 +490,7 @@ class Consensus {
     if (number !== 1 && (!prevBlockInfo || !prevBlockInfo.block)) {
       logger.debug(`[${LOG_HEADER}] No notarized block at number ${number - 1} with ` +
           `hash ${last_hash}`);
-      return;
+      return { voteAgainst: false };
```
platfowner, line 493: question: is this (not having a proper prev block) not a case for voting against?

consensus/index.js
```diff
@@ -516,7 +531,7 @@ class Consensus {
         isSnapDb = true;
         if (!prevDb) {
           logger.error(`[${LOG_HEADER}] Previous db state doesn't exist`);
-          return false;
+          return { voteAgainst: false };
```
platfowner, line 534: question: in which condition this can happen? e.g. in a newly started blockchain node? any possibilities that the block proposal is trying to manipulate something?

consensus/index.js
```diff
@@ -576,7 +590,7 @@ class Consensus {
       prevDb = this.getSnapDb(prevBlock);
       if (!prevDb) {
         logger.error(`[${LOG_HEADER}] Previous db state doesn't exist`);
-        return false;
+        return { voteAgainst: false };
```
platfowner, line 593: question: in which condition can this happen? (sorry for asking about old code)

integration/consensus.test.js
```diff
@@ -445,4 +462,105 @@ describe('Consensus', () => {
...
     });
   });
+
+  describe('Penalties', () => {
+    function sendInvalidBlockProposal() {
+      const lastBlock = getLastBlock(server1);
+      const proposalBlock = Block.create(lastBlock.hash, [], {}, [], lastBlock.number + 1,
+          lastBlock.epoch + 1, '', server2Addr, {}, 0, 0);
+      proposalBlock.hash += '0'; // Invalid block hash
+      const proposalTxUnsigned = {
```
platfowner, line 472: minor: proposalTxBody?

consensus/index.js
platfowner: Or just leaving a todo for now?

consensus/index.js
platfowner: Or just leaving a todo for now?

consensus/index.js
liayoo: Ah, gotcha. Updated the order!

consensus/block-pool.js
liayoo: Done!

consensus/block-pool.js
liayoo: Done!

consensus/consensus-util.js
liayoo: Actually, the proposal txs can be SET txs, after blockNumber becomes greater than 1000 to remove old data.

db/functions.js
liayoo: Done

db/functions.js
liayoo: Done

consensus/index.js
liayoo: This was seriously overdue, so I went ahead and did a bit of refactoring. Sorry for making this PR even bigger than it was.... 😭

liayoo: I'm planning to refactor/simplify a little more in the near future (especially validatePrevBlock())

consensus/index.js
liayoo: ditto

liayoo: Done. Plus, the return value structure was changed to throwing ConsensusErrors.

consensus/index.js
liayoo: Done~

consensus/index.js
liayoo: I thought that the node might have just not received the prev block, but it won't be clear if that's the case or the proposer is malicious. Changed this case to vote against to be on the safer side.

consensus/index.js
liayoo: This would be similar to not having prevBlockInfo. Maybe network failure / isolation / something else could cause this? Changed this to be voted against as well for now.

consensus/index.js
liayoo: ditto

integration/consensus.test.js
liayoo: Done

consensus/block-pool.js
```diff
@@ -215,113 +220,152 @@ class BlockPool {
...
-        this.hashToBlockInfo[blockHash].tallied = 0;
-        blockInfo.votes.forEach((vote) => {
-          if (block.validators[vote.address]) {
-            this.hashToBlockInfo[blockHash].tallied += get(vote, 'tx_body.operation.value.stake');
-          }
-        });
-        this.tryUpdateNotarized(blockHash);
+    if (isValid) {
+      // Check that there's no other block proposed at the same epoch
+      if (this.epochToBlock[block.epoch] && this.epochToBlock[block.epoch] !== block.hash) {
```
platfowner, line 242: minor: can this block be extracted as a sub-function? e.g. addToEpochBlockMap()

consensus/block-pool.js
```diff
@@ -215,113 +220,152 @@ class BlockPool {
...
+              `${block.epoch} (${block.hash}, ${this.epochToBlock[block.epoch]})`);
+          return false;
+        }
+        logger.info(`[${LOG_HEADER}] Multiple blocks proposed for epoch ` +
+            `${block.epoch} (${block.hash}, ${this.epochToBlock[block.epoch]}) BUT is not notarized`);
+        // FIXME: remove info about the block that's currently this.epochToBlock[block.epoch] ?
       }
-      logger.info(
-          `[${LOG_HEADER}] Block added to the block pool: ${block.number} / ${block.epoch}`);
+      if (!this.hashToBlockInfo[blockHash]) {
```
platfowner, line 253: minor: can this block be extracted as a sub-function? e.g. addToHashBlockInfoMap()

consensus/block-pool.js
```diff
@@ -215,113 +220,152 @@ class BlockPool {
...
-    }
-    this.hashToNextBlockSet[lastHash].add(blockHash);
-
-    // Try updating notarized info for block and next block (if applicable)
-    this.tryUpdateNotarized(blockHash);
-    // FIXME: update all descendants, not just the immediate ones
-    if (this.hashToNextBlockSet[blockHash]) {
-      for (const val of this.hashToNextBlockSet[blockHash]) {
-        this.tryUpdateNotarized(val);
+      if (!this.hashToInvalidBlockInfo[blockHash]) {
```
platfowner, line 296: minor: can this block be extracted as a sub-function? e.g. addToInvalidBlockInfoMap()

consensus/block-pool.js
liayoo: Done. Thanks!

consensus/block-pool.js
liayoo: Done!

consensus/block-pool.js
liayoo: Done!

consensus/block-pool.js
```diff
@@ -387,37 +468,115 @@ class BlockPool {
...
+      const proposal = blockInfo.proposal || _get(validBlockCandidate, 'proposal');
+      if (!block || !proposal) {
+        continue;
+      }
+      const talliedVotes = [];
+      let talliedAgainst = 0;
+      for (const vote of blockInfo.votes) {
+        const stake = _get(validators, `${vote.address}.stake`, 0);
+        if (stake > 0) {
+          const res = baseDb.executeTransaction(Transaction.toExecutable(vote), true, true);
```
platfowner, line 543: As discussed offline, can it be avoided to execute vote txs of the invalid blocks already covered by the ancestor blocks in the longest notarized chain?

consensus/block-pool.js
```diff
@@ -387,37 +468,115 @@ class BlockPool {
...
-    } else if (tx.tx_body.operation.type === WriteDbOperations.SET) {
-      return get(tx.tx_body.operation.op_list[0], 'value.block_hash');
-    } else {
-      return null;
+  /**
+   * Checks if there's any evidence in hashToInvalidBlockInfo that can sufficiently
+   * (i.e. sum of evidence votes stakes > 2/3 * total stakes) support offenses.
+   * Executes the valid evidence votes on baseDb, making changes to the state.
+   * Resulting `offenses` has the following structure: { [<address>]: { [<offenseType>]: <number> } },
+   * and the `evidence` has the following structure: { [<address>]: [{ transactions, block, votes, offense_type }, ...] }
```
platfowner, line 514: minor: would it be helpful for readers to put 'offense_type' at the front of {}?

consensus/index.js
```diff
@@ -701,64 +879,90 @@ class Consensus {
...
+    if (this.votedForBlock(proposalBlock.hash)) {
+      logger.info(`[${LOG_HEADER}] Already voted against block ${proposalBlock.hash}`);
+      return;
+    }
     const myAddr = this.node.account.address;
-    const isValidator = block.validators[myAddr];
+    // NOTE(liayoo): Try voting if I was one of the validators in the last notarized block. However,
+    //               only the votes from the validators in next block's previous block will be counted.
+    const lastNotarizedBlock = this.getLastNotarizedBlock();
+    const isValidator = lastNotarizedBlock.validators[myAddr];
```
platfowner, line 954: minor: what about 'validatorInfo'. isValidator sounds like it's a boolean value.

consensus/index.js
```diff
@@ -1229,34 +1475,27 @@ class Consensus {
...
-      }
-      opList.forEach((op) => {
-        if (!op.ref.startsWith(consensusTxPrefix)) return false;
-      })
-      return true;
-    } else {
-      return false;
-    }
+}
+class ConsensusError extends Error {
```
platfowner, line 1479: suggestion: what about moving this definition before the definition of Consensus class?

db/functions.js
```diff
@@ -692,6 +697,33 @@ class Functions {
...
 
+  static getBlacklistLockupExtension(numNewOffenses, updatedNumOffenses) {
+    let extension = 0;
+    for (let n = updatedNumOffenses - numNewOffenses + 1; n <= updatedNumOffenses; n++) {
+      extension += ConsensusConsts.STAKE_LOCKUP_EXTENSION * Math.pow(2, n - 1);
+    }
+    return extension;
+  }
+
+  _blacklist(value, context) {
```
platfowner, line 708: minor: 'blacklist' seems to have wide meanings e.g. listing up bad users etc. could you consider narrowing the coverage a little bit? e.g. _punishOffenses, _handleOffenses, etc.

unittest/test-util.js
```diff
@@ -110,13 +110,46 @@ async function waitForNewBlocks(server, waitFor = 1) {
   }
 }
 
+async function waitUntilNetworkIsReady(serverList) {
```
platfowner, line 113: nice to see it! thanks for adding it~

consensus/block-pool.js
liayoo: Right, thanks for the suggestion! Added recordedInvalidBlockHashSet to the return value of getLongestNotarizedChain() & excluded them from checks in getOffensesAndEvidence()

consensus/block-pool.js
liayoo: Done!

consensus/index.js
liayoo: Done!

consensus/index.js
liayoo: Done!

db/functions.js
liayoo: Changed to _handleOffenses. Thanks!

genesis-configs/base/genesis_rules.json
```diff
@@ -62,6 +62,13 @@
     }
   },
   "consensus": {
+    "blacklist": {
```
platfowner, line 65: minor: what about softer terms like 'offense_history' and 'offense_records'?

consensus/block-pool.js
platfowner: So you mean backupDb() and restoreDb() are used inside of executeTransaction() so that they're not available here?

consensus/block-pool.js
platfowner: Now I see that it runs over hashToInvalidBlockInfo only. :)

db/functions.js
```diff
@@ -692,6 +697,33 @@ class Functions {
     return this.returnFuncResult(context, FunctionResultCode.SUCCESS);
   }
 
+  static getBlacklistLockupExtension(numNewOffenses, updatedNumOffenses) {
```
platfowner, line 700: suggestion: what about getPenaltyLockupExtension()?

db/functions.js
platfowner: lgtm, thanks!

db/functions.js
```diff
@@ -692,6 +697,33 @@ class Functions {
...
+    }
+    for (const [offender, offenseList] of Object.entries(value.offenses)) {
+      const numNewOffenses = Object.values(offenseList).reduce((acc, cur) => acc + cur, 0);
+      const blacklistPath = PathUtil.getConsensusBlacklistAddrPath(offender);
+      this.incValueOrLog(blacklistPath, numNewOffenses, context);
+      const updatedNumOffenses = this.db.getValue(blacklistPath); // new # of offenses
+      const lockupExtension = Functions.getBlacklistLockupExtension(numNewOffenses, updatedNumOffenses);
+      if (lockupExtension > 0) {
+        const expirationPath = PathUtil.getStakingExpirationPath(PredefinedDbPaths.CONSENSUS, offender, 0);
+        const currentExpiration = Math.max(Number(this.db.getValue(expirationPath)), context.timestamp);
```
platfowner, line 720: Could you consider using non-user-controllable time such as block time instead of context.timestamp? Adding a util function (maybe near by lastBlockNumber()?) could be helpful for further use cases.
We used to have some restrictions on the acceptable tx timestamp, but as you know it's turned off now for some reasons.

platfowner (approved):
LGTM with some comments and suggestions.
Thanks for the hard work!

genesis-configs/base/genesis_rules.json
liayoo: Updated to offense_records. Thanks for the suggestion!

db/functions.js
liayoo: Changed it to getLockupExtensionForNewOffenses(). Thanks!

db/functions.js
liayoo: Great idea! Added blockTime parameter when executing transactions & triggering functions. Replaced context.timestamp with context.blockTime as well, except for executing on node.db (when blockTime is null).

liayoo:
Took me a while to fix all the bugs & pass the integration test (still couldn't get the restart test case to work, so it's now commented out..) Extra thanks to @platfowner for taking the time to review this PR! 🙌 Really appreciate all your comments & thoughts.

platfowner:
> Since I'm using the timestamp of the block that the transaction is included in as blockTime, it's null when the tx is executed on node's db.

I see. Would it make sense to use the timestamp of the previous block, which is available most cases (in db, just use lastBlockTime() like lastBlockNumber())? A value not too far from now and deterministic in the network would suffice? Wondering your thoughts.

liayoo:
> > Since I'm using the timestamp of the block that the transaction is included in as blockTime, it's null when the tx is executed on node's db.
> 
> I see. Would it make sense to use the timestamp of the previous block, which is available most cases (in db, just use lastBlockTime() like lastBlockNumber())? A value not too far from now and deterministic in the network would suffice? Wondering your thoughts.

Hmm, I think we should still try to use the timestamp of the block that tx is in if possible, to reduce confusion. But using lastBlockTime() for executing on node.db sounds good to me! I'll try to add the changes to my future PR! Please let me know if I'm not understanding correctly.

platfowner:
> > > Since I'm using the timestamp of the block that the transaction is included in as blockTime, it's null when the tx is executed on node's db.
> > 
> > 
> > I see. Would it make sense to use the timestamp of the previous block, which is available most cases (in db, just use lastBlockTime() like lastBlockNumber())? A value not too far from now and deterministic in the network would suffice? Wondering your thoughts.
> 
> Hmm, I think we should still try to use the timestamp of the block that tx is in if possible, to reduce confusion. But using lastBlockTime() for executing on node.db sounds good to me! I'll try to add the changes to my future PR! Please let me know if I'm not understanding correctly.

Using the current block time looks good if it's available in all block tx evaluation cases. Using the lastBlockTime() in node.db seems to be the best option at this moment. Thanks!

PR #546, opened 2021-08-29
Use array-based radix child map in radix node

platfowner:
Change summary:
- Replace native-map-based radix child map with array-based radix child map

Related issues: https://github.com/ainblockchain/ain-blockchain/issues/519

### DIFF
--- a/db/radix-child-map.js
+++ b/db/radix-child-map.js
@@ -1,0 +1,116 @@
+const CommonUtil = require('../common/common-util');
+
+/**
+ * Implements Radix Child Map, which is used as a component of RadixNode.
+ */
+class RadixChildMap {
+  constructor() {
+    this.size = 0;
+    this.childArray = RadixChildMap.initChildArray();
+  }
+
+  static initChildArray() {
+    const arr = [];
+    for (let i = 0; i < 16; i++) {
+      arr.push(null);
+    }
+    return arr;
+  }
+
+  static _labelRadixToIndex(labelRadix) {
+    if (!labelRadix || !CommonUtil.isString(labelRadix)) {
+      return -1;
+    }
+    const code = labelRadix.charCodeAt(0);
+    if (code >= 48 && code <= 57) {
+      return code - 48;
+    }
+    if (code >= 97 && code <= 102) {
+      return code - 97 + 10;
+    }
+    return -1;
+  }
+
+  static _indexToLabelRadix(index) {
+    if (!CommonUtil.isInteger(index)) {
+      return '';
+    }
+    let code;
+    if (index >= 0 && index <= 9) {
+      code = index + 48;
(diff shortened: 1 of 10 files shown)


db/radix-child-map.js
```diff
@@ -0,0 +1,116 @@
...
+  static initChildArray() {
+    const arr = [];
+    for (let i = 0; i < 16; i++) {
+      arr.push(null);
+    }
+    return arr;
+  }
+
+  static _labelRadixToIndex(labelRadix) {
+    if (!labelRadix || !CommonUtil.isString(labelRadix)) {
```
liayoo, line 21: redundant check? (`!labelRadix`)

db/radix-child-map.js
```diff
@@ -0,0 +1,116 @@
...
+    if (index >= 0 && index < 16) {
+      return this.childArray[index];
+    }
+    return null;
+  }
+
+  set(labelRadix, child) {
+    const index = RadixChildMap._labelRadixToIndex(labelRadix);
+    if (index >= 0 && index < 16) {
+      const isOverwiring = this._has(index);
```
liayoo, line 60: typo (isOverwiring -> isOverwriting)?

db/radix-child-map.js
```diff
@@ -0,0 +1,116 @@
...
+      if (this._has(index)) {
+        this.childArray[index] = null;
+        this.size--;
+        return true;
+      }
+    }
+    return false;
+  }
+
+  keys() {
```
liayoo, line 95: For keys() and values(), do you think we can keep the insertion order? Or maybe leave as a todo?

db/radix-child-map.js
platfowner: Removed. Thanks!

db/radix-child-map.js
platfowner: Fixed. Thanks!

db/radix-child-map.js
platfowner: Good question! This order affects only the internal order of radix children, and the external list is generated by the StateNode.getChildLabels() and StateNode.getChildNodes() which are controlled by RadixTree.labels() and RadixTree.stateNodes(), respectively. So far, an extra hash map (RadixTree.stateNodeMap) is used internally so the order is kept to be the insertion order. I have a plan to disable the use of the extra hash map. In the case, as you mentioned we need to think of how to handle the child order issue. I'll take a look. Thanks.

platfowner:
Comments addressed. PTAL~ @liayoo

db/radix-child-map.js
liayoo: Gotcha! Thanks for the explanation!

liayoo (approved):
LGTM, thanks!

PR #548, opened 2021-08-30
Correctly get candidate peers

minsulee2:
Updates:
- Get new candidate peers depending on the number of peers
- minor fixes

### DIFF
--- a/tracker-server/index.js
+++ b/tracker-server/index.js
@@ -120,11 +120,7 @@ server.on('connection', (ws) => {
     logger.info(`\n<< Update from node [${abbrAddr(nodeInfo.address)}]`);
     logger.debug(`: ${JSON.stringify(nodeInfo, null, 2)}`);
 
-    let newManagedPeerInfoList = [];
-    if (nodeInfo.networkStatus.connectionStatus.outgoingPeers.length <
-        nodeInfo.networkStatus.connectionStatus.maxOutbound) {
-      newManagedPeerInfoList = assignRandomPeers(getPeerCandidates(nodeInfo.address));
-    }
+    const newManagedPeerInfoList = assignRandomPeers(nodeInfo);
     const msgToNode = {
       newManagedPeerInfoList,
       numLivePeers: getNumAliveNodes() - 1   // except for me.
@@ -164,34 +160,33 @@ function getNumNodes() {
   return Object.keys(peerNodes).length;
 }
 
-function assignRandomPeers(candidates) {
-  if (_.isEmpty(candidates)) {
-    return [];
-  }
-
-  const shuffled = _.shuffle(candidates);
-  if (shuffled.length > 1) {
-    return [shuffled.pop(), shuffled.pop()];
+function getNecessaryCandidateNumber(nodeInfo) {
+  const numOfNecessaryCandidates = nodeInfo.networkStatus.connectionStatus.maxOutbound -
+      nodeInfo.networkStatus.connectionStatus.outgoingPeers.length;
+  if (numOfNecessaryCandidates >= 2) {
+    return 2;
   } else {
-    return shuffled;
+    return numOfNecessaryCandidates;
   }
 }
 
-function getPeerCandidates(myself) {
-  const candidates = [];
-  Object.values(peerNodes).forEach(nodeInfo => {
-    if (nodeInfo.address !== myself &&
-        nodeInfo.isAlive === true &&
-        !nodeInfo.networkStatus.connectionStatus.incomingPeers.includes(myself) &&
-        nodeInfo.networkStatus.connectionStatus.incomingPeers.length <
-            nodeInfo.networkStatus.connectionStatus.maxInbound) {
-      candidates.push({
-        address: nodeInfo.address,
-        url: nodeInfo.networkStatus.p2p.url
-      });
-    }
-  });
-  return candidates;
+function assignRandomPeers(nodeInfo) {
+  const numOfNecessaryCandidates = getNecessaryCandidateNumber(nodeInfo);
+  if (numOfNecessaryCandidates) {
+    const candidates = Object.values(peerNodes)
+    .filter(peer =>
+      peer.address !== nodeInfo.address &&
+      peer.isAlive === true &&
+      !peer.networkStatus.connectionStatus.incomingPeers.includes(nodeInfo.address) &&
+      peer.networkStatus.connectionStatus.incomingPeers.length <
+          peer.networkStatus.connectionStatus.maxInbound)
+    .map(peer => ({ address: peer.address, url: peer.networkStatus.p2p.url }));
+
+    const shuffled = _.shuffle(candidates);
+    return shuffled.slice(0, numOfNecessaryCandidates);
+  } else {
+    return [];
+  }
 }
 
 function printNodesInfo() {


tracker-server/index.js
```diff
@@ -164,34 +160,33 @@ function getNumNodes() {
...
-function assignRandomPeers(candidates) {
-  if (_.isEmpty(candidates)) {
-    return [];
-  }
-
-  const shuffled = _.shuffle(candidates);
-  if (shuffled.length > 1) {
-    return [shuffled.pop(), shuffled.pop()];
+function getNecessaryCandidateNumber(nodeInfo) {
+  const numOfNecessaryCandidates = nodeInfo.networkStatus.connectionStatus.maxOutbound -
```
liayoo, line 164: Would it make sense to have separate min parameters for this? (e.g. MIN_OUT_BOUND, MIN_IN_BOUND)

tracker-server/index.js
```diff
@@ -164,34 +160,33 @@ function getNumNodes() {
...
-      candidates.push({
-        address: nodeInfo.address,
-        url: nodeInfo.networkStatus.p2p.url
-      });
-    }
-  });
-  return candidates;
+function assignRandomPeers(nodeInfo) {
+  const numOfNecessaryCandidates = getNecessaryCandidateNumber(nodeInfo);
+  if (numOfNecessaryCandidates) {
```
liayoo, line 175: How about `numOfNecessaryCandidates > 0`?

tracker-server/index.js
liayoo: Or if this is for limiting the max, how about renaming the function to `getMaxNumberOfNewPeers()` or something?

tracker-server/index.js
```diff
@@ -164,34 +160,33 @@ function getNumNodes() {
...
+    const candidates = Object.values(peerNodes)
+    .filter(peer =>
+      peer.address !== nodeInfo.address &&
+      peer.isAlive === true &&
+      !peer.networkStatus.connectionStatus.incomingPeers.includes(nodeInfo.address) &&
+      peer.networkStatus.connectionStatus.incomingPeers.length <
```
platfowner, line 181: As we're strictly controlling the number of outgoing peers, what about apply a kind of best-efforts approach to the number of incoming peers, i.e., choosing peers of smallest incoming peer numbers?

tracker-server/index.js
```diff
@@ -164,34 +160,33 @@ function getNumNodes() {
...
-    return [];
-  }
-
-  const shuffled = _.shuffle(candidates);
-  if (shuffled.length > 1) {
-    return [shuffled.pop(), shuffled.pop()];
+function getNecessaryCandidateNumber(nodeInfo) {
+  const numOfNecessaryCandidates = nodeInfo.networkStatus.connectionStatus.maxOutbound -
+      nodeInfo.networkStatus.connectionStatus.outgoingPeers.length;
+  if (numOfNecessaryCandidates >= 2) {
```
platfowner, line 166: It's hard to catch the underlying ideas. Could you prepare a AIM doc for this change?

tracker-server/index.js
```diff
@@ -120,11 +120,7 @@ server.on('connection', (ws) => {
     logger.info(`\n<< Update from node [${abbrAddr(nodeInfo.address)}]`);
     logger.debug(`: ${JSON.stringify(nodeInfo, null, 2)}`);
 
-    let newManagedPeerInfoList = [];
-    if (nodeInfo.networkStatus.connectionStatus.outgoingPeers.length <
-        nodeInfo.networkStatus.connectionStatus.maxOutbound) {
-      newManagedPeerInfoList = assignRandomPeers(getPeerCandidates(nodeInfo.address));
-    }
+    const newManagedPeerInfoList = assignRandomPeers(nodeInfo);
```
platfowner, line 123: What about checking whether the whole network is one strongly connected component after applying this peer connections?
It looks like there are some good npm packages available: 
- https://www.npmjs.com/package/strongly-connected-components (theoretic ref: https://en.wikipedia.org/wiki/Strongly_connected_component)

tracker-server/index.js
platfowner: As discussed offline, let's put the new design to the next stage job.
Instead could you define '2' as a genesis param?
Also could you simplify params like:
    "MAX_OUTBOUND_LIMIT": 2,
    "MAX_INBOUND_LIMIT": 3,
    "DEFAULT_MAX_OUTBOUND": 2,
    "DEFAULT_MAX_INBOUND": 3

I think we can just discard MAX_..._LIMIT params even before the new design. What do you think?

tracker-server/index.js
minsulee2: Oh, right. this is limiting the max candidate peers at once(here is 2). I take the `getMaxNumberOfNewPeers` seems more intuitive. Thanks!

tracker-server/index.js
minsulee2: Yeah right. removed both the MAX_OUTBOUND_LIMIT and MAX_INBOUND_LIMIT which have not used. and added the `2` as constant. Thanks!

tracker-server/index.js
minsulee2: As we discussed offline, this will be updated in the next job thanks!!!

tracker-server/index.js
minsulee2: This is also discussed offline! I will handle this in the next job!

minsulee2:
All addressed. PTAL!

common/constants.js
```diff
@@ -650,11 +650,8 @@ function initializeNetworkEnvronments() {
     return GenesisParams.network;
   } else {
     return {
-      // NOTE(minsulee2): Need a discussion that MAX_NUM_VALIDATORS and MAX_INBOUND_LIMIT
-      // should not be related to one another.
       P2P_MESSAGE_TIMEOUT_MS: 600000,
-      MAX_OUTBOUND_LIMIT: GenesisParams.consensus.MAX_NUM_VALIDATORS - 1,
-      MAX_INBOUND_LIMIT: GenesisParams.consensus.MAX_NUM_VALIDATORS - 1,
+      LIMIT_NUMBER_OF_CANDIDATES_AT_ONCE: 2,
```
platfowner, line 654: minor: as there are two limits (lower limit and upper limit), what about just MAX_NUMBER_OF_...? and adding PEERS to the name, e.g. MAX_NUM_PEER_CANDIDATES_AT_ONCE?

tracker-server/index.js
```diff
@@ -164,34 +163,33 @@ function getNumNodes() {
...
-            nodeInfo.networkStatus.connectionStatus.maxInbound) {
-      candidates.push({
-        address: nodeInfo.address,
-        url: nodeInfo.networkStatus.p2p.url
-      });
-    }
-  });
-  return candidates;
+function assignRandomPeers(nodeInfo) {
+  const numOfNecessaryCandidates = getMaxNumberOfNewPeers(nodeInfo);
```
platfowner, line 177: minor: what about just maxNumberOfNewPeers? or shortly maxNumNewPeers or maxNewPeers?

platfowner (approved):
Thanks!

common/constants.js
minsulee2: I was struggling to name this constant! I love it!! Thanks!

tracker-server/index.js
minsulee2: I took the long one! Thanks!

minsulee2:
Thanks for the tight review! Good to come back :) merging now!

PR #549, opened 2021-08-30
Use startWith() instead of slice()

platfowner:
Change summary:
- Use startWith() instead of slice()

Related issue: https://github.com/ainblockchain/ain-blockchain/issues/519


liayoo (approved):
LGTM, thanks!

PR #551, opened 2021-09-01
Apply recursive empty node removal for all parent paths

platfowner:
Change summary:
- Make recursive empty node removal applied for all parent paths
- Apply empty node removal to only relevant child label not all child labels
- Apply empty node removal and state info update at the same time
- Add test cases

Related issue: 

### DIFF
--- a/unittest/state-util.test.js
+++ b/unittest/state-util.test.js
@@ -21,6 +21,7 @@ const {
   deleteStateTreeVersion,
   makeCopyOfStateTree,
   equalStateTrees,
+  removeEmptyNodesForAllRootPaths,
   updateProofHashForStateTree,
   updateProofHashForAllRootPaths,
   verifyProofHashForStateTree,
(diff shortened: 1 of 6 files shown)


unittest/state-util.test.js
```diff
@@ -2481,6 +2482,155 @@ describe("state-util", () => {
     })
   })
 
+  describe("empty nodes", () => {
```
liayoo, line 2485: Could you add a case where a parent on the root path still has some child nodes? For example,
```
{
  "0x0001": {
    "0x0011": {
      "0x0111": {
        "0x1111": null
      },
      "0x0112": {
        "0x1121": "V1121"
      }
    }
  },
  "0x0002": {
    "0x0021": "V0021"
  }
}
```

liayoo (approved):
LGTM with a comment. Thanks!

unittest/state-util.test.js
platfowner: Just updated to have a parent having another child. Thanks.

platfowner:

Issue #552, opened 2021-09-01
[BUG] Empty label radix 


platfowner:
'Overwriting a child' with empty label radix:
```
2021-09-01T18:04:16.189Z [node-0-8081] error: [RADIX_TREE] AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA labelSuffix: 6f6d656f6e, childLabelSuffix: 6f6d656f6e6e6e6e6e6e6e, commonPrefix: 6f6d656f6e, childNewLabel: 6e6e6e6e6e6e, newChildLabel:
2021-09-01T18:04:16.190Z [node-0-8081] error: [RADIX_NODE] [setChild] Overwriting a child with radix label  at: Error
    at RadixNode.setChild (/Users/seo/git/comcomai/blockchain/ain-blockchain/db/radix-node.js:111:18)
    at RadixTree._getRadixNodeForWriting (/Users/seo/git/comcomai/blockchain/ain-blockchain/db/radix-tree.js:113:22)
    at RadixTree._setInTree (/Users/seo/git/comcomai/blockchain/ain-blockchain/db/radix-tree.js:150:23)
    at RadixTree.set (/Users/seo/git/comcomai/blockchain/ain-blockchain/db/radix-tree.js:160:15)
    at StateNode.setChild (/Users/seo/git/comcomai/blockchain/ain-blockchain/db/state-node.js:262:22)
    at Function.getRefForWritingToStateRoot (/Users/seo/git/comcomai/blockchain/ain-blockchain/db/index.js:365:16)
    at Function.writeToStateRoot (/Users/seo/git/comcomai/blockchain/ain-blockchain/db/index.js:395:25)
    at DB.writeDatabase (/Users/seo/git/comcomai/blockchain/ain-blockchain/db/index.js:410:25)
    at DB.setValue (/Users/seo/git/comcomai/blockchain/ain-blockchain/db/index.js:706:10)
    at DB.executeSingleSetOperation (/Users/seo/git/comcomai/blockchain/ain-blockchain/db/index.js:882:23).
[...]

PR #553, opened 2021-09-02
Fix empty label radix bug

platfowner:
Change summary:
- Fix empty label radix bug
- Check validity of input labels

Related issues: 
https://github.com/ainblockchain/ain-blockchain/issues/552

### DIFF
--- a/db/radix-tree.js
+++ b/db/radix-tree.js
@@ -55,7 +55,7 @@ class RadixTree {
     let curNode = this.root
     let labelIndex = 0;
     while (labelIndex < hexLabel.length) {
-      const labelRadix = hexLabel.charAt(labelIndex);
+      const labelRadix = hexLabel.slice(labelIndex, labelIndex + 1);
       if (!curNode.hasChild(labelRadix)) {
         return null;
       }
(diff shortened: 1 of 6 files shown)


db/radix-tree.js
```diff
@@ -91,21 +91,42 @@ class RadixTree {
...
-        internalNode.setChild(newChildLabel.charAt(0), newChildLabel.slice(1), newChild);
-
-        return newChild;
+        if (commonPrefix.length === labelSuffix.length) {
+          // Insert an internal node between curNode and the existing child.
+          const internalNode = new RadixNode();
+          curNode.setChild(labelRadix, commonPrefix, internalNode);
+
+          // Insert the existing child node as a child of the internal node.
+          const childNewLabel = childLabelSuffix.slice(commonPrefix.length);
```
liayoo, line 103: It seems that this sequence of ops (getting a label, getting a label radix, getting a label suffix, inserting as a child) is repeated many times. Could you make a function for this?

db/radix-tree.js
```diff
@@ -73,7 +73,7 @@ class RadixTree {
     let curNode = this.root
     let labelIndex = 0;
     while (labelIndex < hexLabel.length) {
-      const labelRadix = hexLabel.charAt(labelIndex);
+      const labelRadix = hexLabel.slice(labelIndex, labelIndex + 1);
```
liayoo, line 76: question) is this change for a performance increase?

liayoo (approved):
LGTM with some comments & questions. Thanks!

db/radix-tree.js
platfowner: Added _setChildWithLabel() util function. Thanks!

db/radix-tree.js
platfowner: No, it's just a result of debugging trials. Just reverted. Thanks!

PR #558, opened 2021-09-02
Bugfix/function failure code

liayoo:
- Return 105 error code if triggered native function fails
- Deprecate saveAndReturnFuncResult() and result paths
- Use blockTime or lastBlockTimestamp (as opposed to transaction timestamp)
- Validate nonce, timestamp in the tx precheck stage

### DIFF
--- a/db/functions.js
+++ b/db/functions.js
@@ -179,6 +179,9 @@ class Functions {
               // Always pops from the call stack.
               this.popCall();
               triggerCount++;
+              if (CommonUtil.isFailedFuncResultCode(result.code)) {
+                break;
+              }
             }
           }
         } else if (functionEntry.function_type === FunctionTypes.REST) {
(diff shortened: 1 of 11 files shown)


db/functions.js
```diff
@@ -843,65 +808,57 @@ class Functions {
     // transferred directly to the admin account.
     if (value.escrow_key !== undefined) {
       const escrowHoldPath = PathUtil.getEscrowHoldRecordPath(
-          userServiceAccountName, value.target, value.escrow_key, timestamp);
+          userServiceAccountName, value.target, value.escrow_key,
+          context.blockTime !== null ? context.blockTime : this.db.lastBlockTimestamp());
```
platfowner, line 812: In the case of db.lastBlockTimestamp() to be used, what about letting the caller set context.blockTime as db.lastBlockTimestamp()?

integration/he_protocol.test.js
```diff
@@ -161,8 +161,12 @@ describe('HE Protocol', () => {
     describe('Create app', () => {
       it("app creation was successful", async () => {
         const appCreationResult = parseOrLog(syncRequest(
-          'GET', server2 + `/get_value?ref=/manage_app/${appName}/create/1/result`).body.toString('utf-8')).result;
-        expect(appCreationResult.code).to.equal(0);
+          'GET', server2 + `/get_value?ref=/manage_app/${appName}/create/1`).body.toString('utf-8')).result;
+        expect(appCreationResult).to.equal({
```
platfowner, line 165: assert.deepEqual()?

platfowner (approved):
LGTM with some comments.
Thanks for the nice refactoring.

db/functions.js
liayoo: Ahh thanks for the idea! 👍 Updated!

integration/he_protocol.test.js
liayoo: Done!

liayoo:

PR #561, opened 2021-09-07
Add dynamic radix tree, build state info in parallel, tweak computeNodeBytes()

platfowner:
Change summary:
- Add dynamic radix tree with enableDynamicRadixTree flag
- Build state info (proof hash, tree height, size, bytes) in parallel using _buildStateInfo() and _buildRadixInfo()
- Simplify computeNodeBytes() to make things more stable
- eraseStateGas() for stability of integration tests
- Fix LIGHTWEIGHT mode
- Deprecate binary radix label mode

Related issues:


db/state-util.js
```diff
@@ -547,7 +547,8 @@ function renameStateTreeVersion(node, fromVersion, toVersion, isRootNode = true)
  */
 function deleteStateTree(node) {
   let numAffectedNodes = 0;
-  if (FeatureFlags.enableRadixTreeLayers) {
+  // 1. Delete children
+  if (FeatureFlags.enableRadixTreeLayers && node.getRadixTreeEnabled()) {
     const childNodes = node.getChildNodes();
     node.deleteRadixTree();
```
liayoo, line 553: could you explicitly set deleteParent = false & add a note?

db/state-util.js
```diff
@@ -579,26 +580,24 @@ function deleteStateTreeVersion(node) {
...
   }
-  node.resetValue();
-  node.resetProofHash();
-  numAffectedNodes++;
 
-  if (FeatureFlags.enableRadixTreeLayers) {
+  // 1. Delete children
+  if (FeatureFlags.enableRadixTreeLayers && node.getRadixTreeEnabled()) {
     const childNodes = node.getChildNodes();
     node.deleteRadixTree();
```
liayoo, line 587: ditto

db/state-util.js
```diff
@@ -579,26 +580,24 @@ function deleteStateTreeVersion(node) {
...
     for (const child of childNodes) {
       numAffectedNodes += deleteStateTreeVersion(child);
     }
-    if (node.numChildren() === 0) {
-      node.setIsLeaf(true);
-    }
   } else {
     for (const label of node.getChildLabels()) {
       const child = node.getChild(label);
       node.deleteChild(label);
```
liayoo, line 594: could you explicitly set shouldUpdateStateInfo = false & add a note?

db/state-node.js
```diff
@@ -20,23 +22,42 @@ class StateNode {
     // Used for leaf nodes only.
     this.value = null;
     this.parentSet = new Set();
+    this.radixTreeEnabled = !FeatureFlags.enableDynamicRadixTree;
```
liayoo, line 25: should this be FeatureFlags.enableRadixTreeLayers..?

unittest/common-util.test.js
```diff
@@ -77,12 +77,21 @@ describe("CommonUtil", () => {
       expect(CommonUtil.toHexString(false)).to.equal('0x');
     })
 
+    it("when non-string input with withPrefix = false", () => {
+      expect(CommonUtil.toHexString(0, false)).to.equal('');
+    })
+
     it("when string input", () => {
       expect(CommonUtil.toHexString('')).to.equal('0x');
```
minsulee2, line 85: Question) It seems a bit weird to me that the empty string is converted into `0x` here. What do you think just return  empty string here as well?

platfowner: Just changed the default value of withPrefix to false.
It's a convenience option for users and I think it would be less error-prone if it's consistent for input, i.e., adding '0x' prefix to event empty string input. Would it be also consistent with existing use cases?:
https://github.com/ainblockchain/ain-util/blob/master/src/index.ts#L135
https://github.com/ethjs/ethjs-util/blob/master/src/index.js#L121

db/state-util.js
platfowner: Just done. Thanks!

db/state-util.js
platfowner: ditto

db/state-util.js
platfowner: Done. Thanks!

db/state-node.js
platfowner: That looks more precise. Just done. Thanks!

common/constants.js
```diff
@@ -110,6 +112,8 @@ const SERVICE_TREE_SIZE_BUDGET = SERVICE_STATE_BUDGET * MAX_STATE_TREE_SIZE_PER_
 const APPS_TREE_SIZE_BUDGET = APPS_STATE_BUDGET * MAX_STATE_TREE_SIZE_PER_BYTE;
 const FREE_TREE_SIZE_BUDGET = FREE_STATE_BUDGET * MAX_STATE_TREE_SIZE_PER_BYTE;
 const STATE_GAS_COEFFICIENT = 1;
+const NUM_CHILDREN_TO_ENABLE_RADIX_TREE = 6;
```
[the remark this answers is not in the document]
platfowner: Thanks for asking about it. It's a little bit tricky solution. One of the worst case scenario of having the same or very nearby values is when child addition and deletion is repeated so the costly conversion operation is performed very often.
So by giving some buffer zone between the two values, I think we can manage such events' frequency very low.

platfowner:
The comments so far were just addressed. PTAL~ @liayoo @minsulee2

common/constants.js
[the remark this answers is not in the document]
liayoo: Ahhh I see. Thanks for the explanation!

db/state-util.js
```diff
@@ -579,26 +580,25 @@ function deleteStateTreeVersion(node) {
...
+  // 1. Delete children
+  if (FeatureFlags.enableRadixTreeLayers && node.getRadixTreeEnabled()) {
     const childNodes = node.getChildNodes();
-    node.deleteRadixTree();
+    node.deleteRadixTree(true);  // shouldDeleteParent = true
     for (const child of childNodes) {
       numAffectedNodes += deleteStateTreeVersion(child);
     }
-    if (node.numChildren() === 0) {
-      node.setIsLeaf(true);
```
liayoo, line 593: Could you help me understand how deleteStateTree() is different from deleteStateTreeVersion(), and why we need to setIsLeaf(true) for deleteStateTree() and not for deleteStateTreeVersion()?

platfowner: Ah, just removed the same part of deleteStateTreeVersion() as well. Now it's covered by node.reset().

liayoo: Thanks for the update! But I'm still having a hard time understanding why we need both deleteStateTree() and  deleteStateTreeVersion(), since their code is almost identical except for the numParent checking in deleteStateTreeVersion()..

platfowner: Right, deleteStateTreeVersion() can be substituted for deleteStateTree() as numParents() should be always 0 even with enableStateVersionOpt = false. The biggest reason I prefer to keep deleteStateTree() separately however is for conceptual reasons as it's deleting the tree itself not a specific tree version in the case of enableStateVersionOpt = false.

db/radix-node.js
```diff
@@ -211,64 +231,71 @@ class RadixNode {
...
+      }, treeInfo);
     }
-    const initialValues = {
-      treeHeight,
-      treeSize,
-      treeBytes,
+    const proofHash = LIGHTWEIGHT ? '' : CommonUtil.hashString(treeInfo.preimage);
+    return {
+      proofHash,
+      treeHeight: treeInfo.treeHeight,
```
liayoo, line 277: question) do we not need to +1 for the treeHeight and treeSize?

platfowner: We count only state nodes for tree info so it done in StateNode (see: line449 of db/state-node.js).

db/state-util.js
liayoo: I see. How about leaving a note or refactoring the functions? Up to you! Thanks for the explanation.

db/radix-node.js
liayoo: Gotcha! 👌

liayoo (approved):
LGTM with a comment. Thanks!

unittest/common-util.test.js
minsulee2: Oh, thanks for the explain! I was just thinking that `0x` prefix with nothing does not seem natural in a binary analysis perspective. Interesting that the ethjs-utils also does that.

minsulee2:
LGTM!

minsulee2 (approved):

db/state-util.js
platfowner: Just merged them into one by introducing deleteOrphanedOnly option. Thanks!

unittest/common-util.test.js
platfowner: Yes, it's now up to the caller's responsibility. Thanks for sharing the ideas.

platfowner:

PR #563, opened 2021-09-08
Feature/minsulee2/network updates for extension

minsulee2:
Updates:
- The connection of nodes is now made bi-directed.
- Tracker distributes nodes depending on the number of incoming nodes.

How the connection works:
1. A node enters the blockchain network with TrackerMessageTypes.CONNECTION.
2. The node gets candidate nodes list.
3. Tries to connect to the nodes in the list with ip address.
4. Share address one another.
5. After sharing address done successfully, (after `inbound` setup done) then, Sends TrackerMessageTypes.CORRESPOND to connect each other.
6. Make bi-directed connection done.


tracker-server/index.js
```diff
@@ -115,29 +115,52 @@ server.on('connection', (ws) => {
...
-      newManagedPeerInfoList,
-      numLivePeers: getNumAliveNodes() - 1   // except for me.
-    };
-    logger.info(`>> Message to node [${abbrAddr(nodeInfo.address)}]: ` +
-        `${JSON.stringify(msgToNode, null, 2)}`);
-    ws.send(JSON.stringify(msgToNode));
-    printNodesInfo();
+    const parsedMessage = JSON.parse(message);
+    switch(_.get(parsedMessage, 'type')) {
+      case TrackerMessageTypes.CONNECTION:
```
liayoo, line 120: Could we try to break the message types into request & response and make them a little more descriptive?
an idea) CONNECTION -> NEW_PEERS_REQUEST, NEW_PEERS_RESPONSE
CORRESPOND -> PEER_INFO_REQUEST, PEER_INFO_RESPONSE

common/constants.js
```diff
@@ -131,6 +131,17 @@ const MessageTypes = {
   HEARTBEAT: 'HEARTBEAT'
 };
 
+/**
+ * Message types for communication between tracker and node.
+ *
+ * @enum {string}
+ */
+const TrackerMessageTypes = {
+  CONNECTION: 'CONNECTION',
```
platfowner, line 140: minor: what about 'CONNECT' verb like others?

p2p/index.js
```diff
@@ -54,9 +55,9 @@ class P2pClient {
   // maxInbound is a rest of connection after maxOutbound is set.
   initConnections() {
     this.maxOutbound = process.env.MAX_OUTBOUND ?
```
platfowner, line 57: could you update this variable name as well ('targetOutBound')?

tracker-server/index.js
```diff
@@ -155,6 +178,14 @@ function abbrAddr(address) {
   return `${address.substring(0, 6)}..${address.substring(address.length - 4)}`;
 }
 
+function setNodeInfo(ws, nodeInfo) {
```
platfowner, line 181: minor: what about 'setPeerNodes' or 'setWsListAndPeerNodes' as its name? This function seems to be just using the given 'nodeInfo'?

tracker-server/index.js
```diff
@@ -183,10 +214,12 @@ function assignRandomPeers(nodeInfo) {
       !peer.networkStatus.connectionStatus.incomingPeers.includes(nodeInfo.address) &&
       peer.networkStatus.connectionStatus.incomingPeers.length <
           peer.networkStatus.connectionStatus.maxInbound)
-    .map(peer => ({ address: peer.address, url: peer.networkStatus.p2p.url }));
-
-    const shuffled = _.shuffle(candidates);
-    return shuffled.slice(0, maxNumberOfNewPeers);
+    .sort((a, b) =>
+      a.networkStatus.connectionStatus.incomingPeers -
```
platfowner, line 218: nice~

p2p/index.js
```diff
@@ -134,30 +135,63 @@ class P2pClient {
     this.intervalConnection = null;
   }
 
-  updateNodeStatusToTracker() {
-    const updateToTracker = this.getStatus();
+  connectToCorrespondingNode(address) {
```
liayoo, line 138: Could you leave a note of why we need an extra message type (TrackerMessageTypes.CORRESPOND) and how it would be unnecessary once we deprecate the tracker in the near future? Thanks!

tracker-server/index.js
platfowner: +1
Could you make it more clear who makes requests and who responds for the requests with the message type naming?

p2p/index.js
liayoo: Come to think of it, do you think we can just pass the port information (if that's the only info we need) when we connect to the peer so that we don't need to ask the tracker for it if we already have a 1-way connection from the peer?

common/constants.js
```diff
@@ -131,6 +131,17 @@ const MessageTypes = {
...
+  CORRESPOND: 'CORRESPOND',
```
platfowner, line 141: 'RESPOND'?

p2p/server.js
```diff
@@ -426,6 +427,9 @@ class P2pServer {
                 return;
               }
               socket.send(JSON.stringify(payload));
+              if (!this.client.outbound[address]) {
+                this.client.connectToCorrespondingNode(address);
```
platfowner, line 431: It looks it makes bidirectional connection in response to MessageTypes.ADDRESS_REQUEST?
What about introducing another message type?

p2p/index.js
platfowner, line 138: The name sounds like it actually make connection but what it actually does is to report to the tracker server?
Could you rename it appropriately?

p2p/index.js
```diff
@@ -134,30 +135,63 @@ class P2pClient {
     this.intervalConnection = null;
   }
 
-  updateNodeStatusToTracker() {
```
platfowner, line 137: It looks like updateNodeStatusToTracker is divided into pieces.
Wouldn't it make more traffic to the tracker?
As the tracker is getting traffic from all blockchain nodes, I think it need to minimize the traffic size.
What do you think?

platfowner:
Could you share a design doc before making actual changes? I think it would be helpful for faster review and avoiding unnecessary trial and errors. @minsulee2

tracker-server/index.js
minsulee2: Coooooool. I took @liayoo 's idea! Thanks!

common/constants.js
minsulee2: ditto; I took @liayoo 's

tracker-server/index.js
minsulee2: Thanks :)

common/constants.js
minsulee2: ditto; I took @liayoo 's

p2p/index.js
minsulee2: All addressed for this !

tracker-server/index.js
minsulee2: Right! I took `setPeerNodes ` for this! thanks!

p2p/index.js
minsulee2: Yeah, I think so too now. But, As discussed offline with @platfowner , the design will change in the next move. (This part will be removed.) I will leave some notes regarding this! Cheers thank you!

p2p/server.js
minsulee2: This is in response to `TrackerMessageTypes.PEER_INFO_RESPONSE` . As we discussed offline, this part will be updated with new scheme! I will leave some notes for this! Thanks!

p2p/index.js
minsulee2: As mentioned above, I will leave some notes for this as well.

p2p/index.js
minsulee2: It is not actually. The job of `updateNodeStatusToTracker` is now done via `updateNodeInfoToTracker` and the other tracker messages are only called when they get into the network or make connections. Thanks for the nice question!

minsulee2:
@platfowner @liayoo PTAL. Some addressed and Leave some notes. After this PR, I will move on to AIM docs and share the new design with you guys!

tracker-server/index.js
```diff
@@ -155,6 +181,14 @@ function abbrAddr(address) {
   return `${address.substring(0, 6)}..${address.substring(address.length - 4)}`;
 }
 
+function setpeerNodes(ws, nodeInfo) {
```
platfowner, line 184: 'setPeerNodes'?

p2p/server.js
platfowner: I see. Ao 'address and security checks' means direct connection making to the peer? It would be more clear if something like 'direct connection' is emphasized (minor). Thanks.

p2p/index.js
```diff
@@ -134,30 +132,63 @@ class P2pClient {
...
+    const message = {
+      type: TrackerMessageTypes.PEER_INFO_REQUEST,
+      data: address
+    }
+    logger.debug(`\n >> Update to [TRACKER] ${TRACKER_WS_ADDR}: ` +
+        `${JSON.stringify(message, null, 2)}`);
+    this.trackerWebSocket.send(JSON.stringify(message));
+  }
+
+  connectToOtherPeers() {
```
platfowner, line 145: minor: what about sendRequestForNewPeers?

p2p/index.js
platfowner: I see. Then what about sendRequestForPeerInfo? (minor)

p2p/index.js
```diff
@@ -134,30 +132,63 @@ class P2pClient {
...
+    const message = {
+      type: TrackerMessageTypes.NEW_PEERS_REQUEST,
+      data: this.getStatus()
+    };
+    logger.debug(`\n >> Connect to [TRACKER] ${TRACKER_WS_ADDR}: ` +
+        `${JSON.stringify(message, null, 2)}`);
+    this.trackerWebSocket.send(JSON.stringify(message));
+  }
+
+  updateNodeInfoToTracker() {
```
platfowner, line 155: minor: what about updatePeerInfoToTracker?

p2p/index.js
```diff
@@ -134,30 +132,63 @@ class P2pClient {
...
-      if (this.server.node.state === BlockchainNodeStates.STARTING) {
-        await this.startNode(parsedMsg.numLivePeers);
+      const parsedMessage = JSON.parse(message);
+      logger.info(`\n<< Message from [TRACKER]: ${JSON.stringify(parsedMessage, null, 2)}`);
+      switch(_.get(parsedMessage, 'type')) {
+        case TrackerMessageTypes.NEW_PEERS_RESPONSE:
+          const data = parsedMessage.data;
+          this.connectToPeers(data.newManagedPeerInfoList);
+          if (this.server.node.state === BlockchainNodeStates.STARTING) {
+            await this.startNode(data.numLivePeers);
```
platfowner, line 181: minor: what about startBlockchainNode for clearer meaning?

tracker-server/index.js
minsulee2: Fixed!

p2p/index.js
minsulee2: I should've changed it. I like the name! Changed! Thanks!

p2p/index.js
minsulee2: Updated!

p2p/index.js
minsulee2: It is clearer! Updated!

minsulee2:
PTAL! All addressed!

platfowner (approved):
(Could you take a look at the remaining comments? one on 'connectToCorrespondingNode' and the other on a note with 'anddress and security checks')
Thanks!

p2p/server.js
minsulee2: Updated the notes! Thanks!

p2p/index.js
minsulee2: I like the name! Thanks!

minsulee2:
Thanks for the fast review! merging it now!

PR #564, opened 2021-09-10
Feature/minsulee2/network graph

minsulee2:
Using d3 package, Draw graph of ain-blockchain network.
- Simply run `node tools/network/index.js`.
- It will show a topology of nodes in the network.
- Go to `localhost:8000`


tools/network/index.js
```diff
@@ -0,0 +1,65 @@
...
+const buildGraphData = (peerNodes) => {
+  const data = { nodes: [], links: [] };
+  const peerNodeIdMap = { };
+
+  Object.keys(peerNodes).forEach((peerNode, i) => {
+    Object.assign(peerNodeIdMap, { [peerNode]: i });
+    data.nodes.push({ address: abbrAddr(peerNode) });
+  });
+
+  Object.values(peerNodes).forEach(nodeInfo => {
```
liayoo, line 25: You could combine the two loops around Object.keys() & Object.values() into one by using Object.entries() like so:
```
Object.entries(peerNodes).forEach(([key, val], index) => {
    // do stuff
});
```

liayoo (approved):
LGTM with some comments. Thanks for the nice tool! Maybe we can extend this & show the graph on insight as well! (one day..) 🙂

tools/network/README.md
```diff
@@ -0,0 +1,4 @@
+# How to Run The Graph board
```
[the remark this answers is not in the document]
minsulee2: I like network-topology! Done!

tools/network/index.js
minsulee2: Oh, I see. I am just not used to using entries :) updated!! it's better !

minsulee2:
@liayoo Thanks for the fast review!! All addressed! merging now 👍

PR #565, opened 2021-09-10
Revamp check-in service

liayoo:
- Updated check-in native functions & db paths to match the check-out service
- Deprecated remote transactions
- Added network name, chain id & token id to check-in & token bridge db paths
- Added check-in tool scripts & removed outdated cross-shard tool scripts
- Added prevVal (previous value snapshot) as an argument to triggerFunctions()
- [Design docs](https://docs.google.com/document/d/1rb9dS0mtgPw6ndGljCK15BIf_wCmrIW8zzgTexLoefg/edit#)


genesis-configs/he-shard/genesis_token.json
```diff
@@ -4,14 +4,15 @@
...
     "AIN": {
-      "0": {
-        "network_id": 1,
-        "token_pool": "0xBBB2219cD5eACc54Ce95deF7a67dDe71C8241891",
-        "min_checkout_per_request": 1000,
-        "max_checkout_per_request": 10000,
-        "max_checkout_per_day": 1000000,
-        "token_exchange_rate": 10,
-        "token_exchange_scheme": "FIXED"
+      "1": {
```
platfowner, line 7: what about just using '0' (testnet and default)?

genesis-configs/sim-shard/genesis_token.json
```diff
@@ -4,14 +4,15 @@
...
     "AIN": {
-      "0": {
-        "network_id": 1,
-        "token_pool": "0xB0B0d133E6a82A982396Fa7E4B4F9eC85d1DFBF4",
-        "min_checkout_per_request": 1000,
-        "max_checkout_per_request": 10000,
-        "max_checkout_per_day": 1000000,
-        "token_exchange_rate": 10,
-        "token_exchange_scheme": "FIXED"
+      "1": {
```
platfowner, line 7: ditto?

tools/checkin/sendOpenCheckinTx.js
```diff
@@ -0,0 +1,49 @@
...
+let config = {};
+
+function buildOpenCheckinTxBody(fromAddr, tokenAmount, checkinId) {
+  return {
+    operation: {
+      type: 'SET_VALUE',
+      ref: `/checkin/requests/ETH/3/0xB16c0C80a81f73204d454426fC413CAe455525A7/${fromAddr}/${checkinId}`,
+      value: {
+        amount: tokenAmount,
+        sender: config.senderAddr
```
platfowner, line 12: Is this an ETH address, who will send ETH to the pool?
If so, could you leave some comments here and in the _openCheckin function code?

common/path-util.js
```diff
@@ -181,42 +184,58 @@ class PathUtil {
...
+        PredefinedDbPaths.CHECKIN_STATS_PENDING, PredefinedDbPaths.CHECKIN_TOKEN_POOL, tokenPoolAddr]);
   }
 
-  static getCheckinPayloadPathFromValuePath(valuePath) {
-    const branchPath = CommonUtil.formatPath(valuePath.slice(0, -3));
-    return PathUtil.getCheckinPayloadPath(branchPath);
+  static getCheckinPendingAmountPerSenderPath(networkName, chainId, tokenId, sender) {
+    return CommonUtil.formatPath([
+        PredefinedDbPaths.CHECKIN, PredefinedDbPaths.CHECKIN_STATS,
+        PredefinedDbPaths.CHECKIN_STATS_PENDING, networkName, chainId, tokenId, sender]);
```
platfowner, line 208: Could you check that the following scenario is handled properly?
- User A send checkin request with sender X and amount Y
- User B send checkin request with sender X and amount Y
- Sender A send Y token amount to the pool
?

db/functions.js
```diff
@@ -911,124 +904,180 @@ class Functions {
...
-  // TODO(platfowner): Support refund feature.
-  _openCheckin(value, context) {
-    const parsedValuePath = context.valuePath;
-    const payloadTx = _.get(value, 'payload', null);
-    const txHash = CommonUtil.hashSignature(payloadTx.signature);
-    if (!this.tp || !this.db.isNodeDb) {
-      // It's not the backupDb
-      logger.info(`  =>> Skip sending signed transaction to the parent blockchain: ${txHash}`);
-      return this.returnFuncResult(context, FunctionResultCode.SUCCESS);
+  updatePendingCheckin(networkName, chainId, tokenId, sender, tokenPool, amount, isIncrease, context) {
```
platfowner, line 907: What about more descriptive name like 'updateStatsForPendingCheckin'?

db/functions.js
```diff
@@ -911,124 +904,180 @@ class Functions {
...
+              PathUtil.getCheckinPendingAmountPerTokenPoolPath(tokenPool), amount, context))) {
+        return FunctionResultCode.INTERNAL_ERROR;
+      }
     }
-    if (!this.validateCheckinParams(context.params)) {
-      return this.returnFuncResult(context, FunctionResultCode.FAILURE);
+    return true;
+  }
+
+  updateCompleteCheckin(user, amount, context) {
```
platfowner, line 936: ditto 'updateStatsForCompletedCheckin'?

db/functions.js
```diff
@@ -911,124 +904,180 @@ class Functions {
...
+  }
+
+  validateCheckinAmount(networkName, chainId, tokenId, sender, amount, tokenPool) {
+    // NOTE(liayoo): pending amounts do NOT include the request's amount yet.
+    const tokenPoolBalance = this.db.getValue(PathUtil.getAccountBalancePath(tokenPool)) || 0;
+    const pendingPerTokenPool = this.db.getValue(
+        PathUtil.getCheckinPendingAmountPerTokenPoolPath(tokenPool)) || 0;
+    const pendingSender = this.db.getValue(
+        PathUtil.getCheckinPendingAmountPerSenderPath(networkName, chainId, tokenId, sender)) || 0;
+    if (pendingSender > 0) {
```
platfowner, line 964: minor: what about checking this before fetching other values like tokenPoolBalance and pendingPerTokenPool?

db/functions.js
```diff
@@ -911,124 +904,180 @@ class Functions {
...
+    }
+    return this.returnFuncResult(context, FunctionResultCode.SUCCESS);
   }
 
-  _closeCheckin(value, context) {
-    if (!this.tp || !this.db.isNodeDb) {
-      // It's not the backupDb
-      logger.info('  =>> Skip sending transfer transaction to the shard blockchain');
+  _cancelCheckin(value, context) {
+    if (value !== null) {
```
platfowner, line 1015: What does this mean: value !== null? Could you add some comments?

common/path-util.js
liayoo: Thanks for asking! In that case, if user A & B are checking in with the same networkName, chainId & tokenId, the later request will be rejected since there is already a pending request with the same networkName, chainId, tokenId & sender. Let me know if you have any other questions!

db/functions.js
liayoo: It means that it's not a SET_VALUE for a cancel, but a request. The cancel should only happen if the value is null (likewise, _openCheckin should only run if the value is NOT null). I'll add some notes on this!

common/path-util.js
platfowner: So sender is a key. got it, thanks!

genesis-configs/he-shard/genesis_token.json
liayoo: Ah, my mistake. Done, thanks!

genesis-configs/sim-shard/genesis_token.json
liayoo: Done!

tools/checkin/sendOpenCheckinTx.js
liayoo: That's right. Added some comments on it!

db/functions.js
liayoo: Done, thanks!

db/functions.js
liayoo: Updated!

db/functions.js
liayoo: Great idea. Done!

db/functions.js
liayoo: I think there were some timing issues on github - anyway, added some comments on _openCheckin and _cancelCheckin :)

liayoo:
Addressed all comments (at last! sorry for the prolonged delay), PTAL @platfowner

platfowner (approved):
LGTM
Thanks for the nice refactoring!

platfowner:
Let me merge it shortly..

PR #567, opened 2021-09-13
Add echo-bot (http://echo-bot.ainetwork.ai) to the function triggering whitelist

platfowner:
Change summary:
- Add echo-bot (http://echo-bot.ainetwork.ai) to the function triggering whitelist


liayoo (approved):
LGTM, thanks!

PR #572, opened 2021-09-17
Fix intervalCount bug in EventCounter.

platfowner:


traffic/event-counter.js
```diff
@@ -2,20 +2,19 @@ const _ = require('lodash');
 
 class EventCounter {
   constructor(
-      intervalMs = 60000,  // 1 min
-      maxIntervals = 180, // 3 hours
+      intervalMs,  // 1 min
```
liayoo, line 5: question) are the comments still valid? (1 min, 3 hrs)

liayoo (approved):
LGTM with a comment, thanks!

traffic/event-counter.js
platfowner: Just removed. Thanks!

Issue #577, opened 2021-09-24
[BUG] Blockchain node dies with '[CONSENSUS] Offenses and evidence don't match' error:


platfowner:
Blockchain node dies with '[CONSENSUS] Offenses and evidence don't match' error:
```
2021-09-23T19:07:48.757Z [node-2-8080] error: [CONSENSUS] Offenses and evidence don't match: {} / {"0x03AAb7b6f16A92A1dfe018Fe34ee420eb098B98A":[{"transactions":[{"tx_body":{"operation":{"type":"SET","op_list":[{"type":"SET_VALUE","ref":"/consensus/number/10581/propose","value":{"number":10581,"epoch":20201435,"validators":{"0x00ADEc28B6a845a085e03591bE7550dd68673C1C":{...
```
![offences_evidence_mismatch](https://user-images.githubusercontent.com/46211617/134621500-7db5113b-5070-4cd1-a773-133ad83fd4ec.jpg)

@liayoo  Could you take a look?

platfowner:
Notably many big log files are created as a result.

platfowner:
Let me close this issue. @liayoo

PR #583, opened 2021-09-24
Fix "offenses and evidence don't match" error

liayoo:
Updated offenses getting logic in getOffensesFromProposalTx (). (after blockNumber > 1000, proposal tx becomes a SET operation)

Fixes https://github.com/ainblockchain/ain-blockchain/issues/577


platfowner (approved):
LGTM

PR #584, opened 2021-09-24
Make deleteRadixTreeVersion() directly call deleteStateTreeVersion()

platfowner:
Change summary:
- Make deleteRadixTreeVersion() directly call deleteStateTreeVersion() instead of using getChildNodes()

Background:
- One lesson learned from many previous experiments was that gathering full child node information using getChildNodes() or getChildLabels() is a very costly operation, and we need to avoid it if possible. This PR actually does it by making deleteRadixTreeVersion() directly call deleteStateTreeVersion().

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/517
- https://github.com/ainblockchain/ain-blockchain/issues/519

Performance test: 
50tps


minsulee2 (approved):
LGTM!

liayoo (approved):
LGTM, thanks!

PR #597, opened 2021-10-01
Remove one step connecting one another.

minsulee2:
Updates:
- Now each node **directly** asks bi-connection via peerInfo.
- Leave TODO comments for next works.

To avoid too big PR, Small updates just has done. I am also working on with AIM-021! No worries!


p2p/index.js
```diff
@@ -312,6 +298,7 @@ class P2pClient {
   sendAddress(socket) {
```
platfowner, line 298: What about giving a new name e.g. sendPeerInfo() as it's doing so now?

platfowner (approved):
**Please make sure the code is kept working and the tests pass all the time.**
Thanks!

p2p/index.js
minsulee2: Updated! But it will be renamed again as it is going to do more than that in the next update!

minsulee2:
@platfowner Thanks for the fast review! Before moving on to the next updates, I will check some tests first!!

PR #606, opened 2021-10-08
Fix log file prefix for nodes with accounts injected with keystores.

### LINKED ISSUE #603: [BUG] Need to updated log file prefix when keystore file is used
Currently, the server type is identified as a tracker server when a keystore file is used instead of ACCOUNT_INDEX.

log_file_prefix

liayoo:
- Updated to use `node-${PORT}` prefix for nodes that start with KEYSTORE_FILE_PATH env var
- Fixes #603


platfowner (approved):
LGTM
Thanks!

PR #615, opened 2021-10-12
Bugfix/liayoo/handle invalid tx sig

### LINKED ISSUE #612: Not returning errors for txs with invalid signatures
We're just returning the execution result when a tx has an invalid signature. We should return an appropriate error code for the case, and furthermore, we should check signatures before calling executeTransaction().

liayoo:
- Return error codes for invalid tx sig & invalid tx cases
- Check tx sig before calling executeTransaction()
- Fixes #612


consensus/index.js
```diff
@@ -1029,6 +1029,10 @@ class Consensus {
       if (blockInfo.block.number < lastNotarizedBlock.number) {
         continue;
       }
+      if (!ConsensusUtil.isValidConsensusTx(blockInfo.proposal)) {
+        logger.info(`[${LOG_HEADER}] Invalid consensus tx: ${JSON.stringify(blockInfo.proposal)}`);
```
minsulee2, line 1033: Can it be logger.error?

tx-pool/index.js
```diff
@@ -26,22 +26,9 @@ class TransactionPool {
     this.txCountTotal = 0;
   }
 
+  // NOTE(platfowner): A transaction needs to be converted to an executable form
+  //                   before being added.
   addTransaction(tx, isExecutedTx = false) {
-    // NOTE(platfowner): A transaction needs to be converted to an executable form
-    //                   before being added.
-    if (!Transaction.isExecutable(tx)) {
```
platfowner, line 32: What about keeping this if-block for safety?

node/index.js
```diff
@@ -434,6 +434,16 @@ class BlockchainNode {
           logger, 2, `[${LOG_HEADER}] Blockchain node is NOT in SERVING mode: ${this.state}`, 0);
     }
     const executableTx = Transaction.toExecutable(tx);
+    if (!Transaction.isExecutable(executableTx)) {
+      return CommonUtil.logAndReturnTxResult(
+          logger, 5,
+          `[${LOG_HEADER}] Invalid transaction: ${JSON.stringify(executableTx, null, 2)}`);
+    }
+    if (!LIGHTWEIGHT) {
+      if (!Transaction.verifyTransaction(executableTx)) {
```
minsulee2, line 443: minor question: Does it make sense that verifyTransaction() first and isExecutable later? I think checking executable is not necessary if verification fails.

platfowner: fyi, isExecutable() is very light-weighted operation than verifyTransaction().

PR #617, opened 2021-10-13
Add client api for getting block info (finalized_at)

liayoo:
- Added numberToBlockInfo mapping in Blockchain (`blockNumber -> { finalized_at: <number> }`)
- Added get_block_info_by_number client API


blockchain/index.js
```diff
@@ -141,6 +152,15 @@ class Blockchain {
     logger.info(`[${LOG_HEADER}] Successfully added block ${block.number} to chain.`);
   }
 
+  updateNumberToBlockInfo(block) {
+    this.numberToBlockInfo[block.number] = {
+      finalized_at: Date.now()
```
platfowner, line 157: question: what is the use case in mind of finalized_at?
wondering if it's just okay to have different values depending on blockchain node or whether the block was reloaded by restarting?

platfowner: ah, it's for tta certification. got it, thanks!

Issue #613, opened 2021-10-12
Need state proof verification tool


platfowner:
We need a state proof verification tool.
- Input: state proofs
- Output: verified or not verified

PR #621, opened 2021-10-13
Add tools for state proof verification

platfowner:
Change summary:
- Simplify state proof structure
- Add verifyStateProof() util function
- Add tools/proof-hash/verifyProof.js for state proof verification
- Move verifyBlock.js to tools/proof-hash/

Related issues:
https://github.com/ainblockchain/ain-blockchain/issues/613

### DIFF
--- a/db/state-util.js
+++ b/db/state-util.js
@@ -677,14 +677,14 @@ function getProofHashOfRadixNode(stateProofHash, subProofList) {
  * 
  * Returns { proofHash, isStateProof } when successful, otherwise null.
  */
-function verifyStateProofRecursive(proof) {
+function verifyStateProofInternal(proof) {
   let stateProofHash = null;
   let radixProofHash = null;
   const subProofList = [];
   for (const [label, value] of Object.entries(proof)) {
     let proofHash = null;
     if (CommonUtil.isDict(value)) {
-      const subProof = verifyStateProofRecursive(value);
+      const subProof = verifyStateProofInternal(value);
       if (subProof === null) {
         return null;
       }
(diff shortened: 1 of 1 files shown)


db/state-util.js
```diff
@@ -636,6 +638,78 @@ function getProofOfStatePath(root, fullPath) {
...
+    return {
+      proofHash: radixProofHash,
+      isStateProof: false,
+    }
+  } else {
+    return null;  // proof hash mismatch!
+  }
+}
+
+function verifyStateProof(proof) {
```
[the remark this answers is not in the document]
platfowner: ditto

platfowner:

db/state-util.js
```diff
@@ -632,10 +641,112 @@ function getProofOfStatePathRecursive(fullPath, node, index) {
...
+
+/**
+ * Verifies a state path.
+ * 
+ * @param {Object} proof state proof
+ * 
+ * Returns root proof hash if successful, otherwise null.
+ */
+function verifyStateProof(proof) {
+  const verified = verifyStateProofRecursive(proof);
```
liayoo, line 743: Can verified.isStateProof be false, and does it affect how to interpret the result?

platfowner: isStateProof is for the purpose of internal use. If the input proof is from /get_state_proof or ain_getStateProof, it will always return isStateProof = false since the root node will always be a root radix node of the state node of the given 'ref' path. 

The current structure of the state proof is somewhat complicated and confusing. I have an idea to simplify the current structure of the state proof, which I'm preparing for the next PR.

PR #627, opened 2021-10-17
Add CORS_WHITELIST env var, enable cors.

liayoo:
- Allow CORS
- Accept CORS_WHITELIST env var (comma-separated list of domains)
- Set some default CORS_WHITELIST domains (ainetwork.ai, ainize.ai)
- Internal ref: https://comcomai.slack.com/archives/CBKK76P9S/p1634442856023100

### DIFF
--- a/common/common-util.js
+++ b/common/common-util.js
@@ -728,6 +728,16 @@ class CommonUtil {
     if (!CommonUtil.isArray(resList)) return [];
     return resList.map((res) => DB.trimExecutionResult(res));
   }
+
+  static getCorsWhitelist(input) {
+    let whitelist = ['https://ainetwork.ai', 'https://ainize.ai', /\.ainetwork\.ai$/,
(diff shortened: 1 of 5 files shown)


common/common-util.js
```diff
@@ -728,6 +728,16 @@ class CommonUtil {
     if (!CommonUtil.isArray(resList)) return [];
     return resList.map((res) => DB.trimExecutionResult(res));
   }
+
+  static getCorsWhitelist(input) {
+    let whitelist = ['https://ainetwork.ai', 'https://ainize.ai', /\.ainetwork\.ai$/,
```
platfowner, line 733: what about defining it as a constant e.g. DEFAULT_CORS_WHITELIST?

liayoo: Done, thanks! Also, updated the logic to overwrite the default setting if CORS_WHITELIST env var is given.

PR #628, opened 2021-10-17
Fix state proof verification bug

platfowner:
Change summary:
- Fix state proof verification bug by fixing the order of children in radix proof hash computation
- Set enableReceiptPathPrefixLayers=true always and remove the flag
- Set enableRadixTreeLayers=true always and remove the flag

Related issue:
https://github.com/ainblockchain/ain-blockchain/issues/626

### DIFF
--- a/db/radix-node.js
+++ b/db/radix-node.js
@@ -279,13 +279,13 @@ class RadixNode {
   getChildLabelRadices() {
     // NOTE(platfowner): Sort child nodes by label radix for stability.
     return [...this.radixChildMap.keys()]
-        .sort((a, b) => a[0].localeCompare(b[0]));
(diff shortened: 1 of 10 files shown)


db/radix-node.js
```diff
@@ -277,11 +277,16 @@ class RadixNode {
   }
 
   getChildLabelRadices() {
-    return [...this.radixChildMap.keys()];
+    // NOTE(platfowner): Sort child nodes by label radix for stability.
+    return [...this.radixChildMap.keys()]
+        .sort((a, b) => a[0].localeCompare(b[0]));
```
liayoo, line 282: question) why do we need to compare only the chars at index 0?

minsulee2: I am wondering as well!

Issue #649, opened 2021-10-26
[BUG] Invalid state proof hash


platfowner:
Error log:
```
2021-10-26T09:24:57.758Z [node-0-8081] error: [NODE] [executeBlockOnDb] Invalid state proof hash (58): 0x9346dbd4dce2979a47c6e8ec2246de136016cffa9fd57adbf0f89b6d0af66c3f, 0x733fa7f672adcb6d6d7954e70ff4d1440a35bff173c5a5a893ab2dc131916e16
```

How to reproduce:
```
sh start_servers_afan_local.sh

kill 62747

MIN_NUM_VALIDATORS=3 ACCOUNT_INDEX=0 CONSOLE_LOG=true ENABLE_DEV_SET_CLIENT_API=true ENABLE_TX_SIG_VERIF_WORKAROUND=true ENABLE_GAS_FEE_WORKAROUND=true BLOCKCHAIN_DATA_DIR=~/ain_blockchain_data node ./client/index.js
```

liayoo:
Possibly related: 
```
error: [NODE] [applyBlocksToDb] Failed to validate state proof of block
```

platfowner:
Result of experiment with latest updates (https://github.com/ainblockchain/ain-blockchain/pull/684 + block in snapshot):
- **restart in fast sync mode succeeded**!:
```
bash start_node_genesis_gcp.sh summer 0 0 --keep-code 
```
![restart_fast_block_in_snapshot](https://user-images.githubusercontent.com/46211617/141879148-f344d60f-b196-4720-a4e4-3fcdf9f9f917.jpg)


- **restart in full sync mode failed**!:
```
bash start_node_genesis_gcp.sh summer 0 0 --keep-code --full-sync
```
![restart_full_block_in_snapshot](https://user-images.githubusercontent.com/46211617/141879178-9a2bbb56-4e8e-4bc7-b0ce-8fab896eb800.jpg)

What's observed is that the node gets stuck after getting NEW_PEERS_RESPONSE from the tracker server:
```
2021-11-16T01:21:41.469Z [node-0-8080] info: [NODE] Now node in STARTING state!
2021-11-16T01:21:41.470Z [node-0-8080] info: [NODE] [setAccountAndInitShardSetting] Initializing a new blockchain n
ode with account: 0x000AF024FEDb636294867bEff390bCE6ef9C5fc4
[...]

PR #663, opened 2021-11-02
Fix lastBlockNumber = -1 for removeOldReceipts when loading blocks.

liayoo:
Fixed lastBlockNumber being -1 when loading more than ON_MEMORY_CHAIN_LENGTH  blocks. 
It was causing issues with removeOldReceipts() & thus state hashes.

Related: #649 

- With that said, if a node is fast-syncing, it would have a similar issue as it may not have old blocks to get the transactions from. We need a way to get around this!

- This fixed the state hash issues with loading blocks, but I still see state hash issues when merging blocks from the peers (after successfully loading blocks)

### DIFF
--- a/blockchain/index.js
+++ b/blockchain/index.js
@@ -150,6 +150,11 @@ class Blockchain {
 
     this.chain.push(block);
     logger.info(`[${LOG_HEADER}] Successfully added block ${block.number} to chain.`);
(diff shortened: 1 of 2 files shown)


platfowner (approved):
LGTM
Thanks!

Issue #664, opened 2021-11-02
[BUG] No 'Invalid state proof hash' error log 


platfowner:
No 'Invalid state proof hash' error log with:
```
bash restart_node_gcp.sh dev 0 0 full
```

platfowner:
Fixed.

PR #665, opened 2021-11-02
Fix no 'Invalid state proof hash' error logging bug

platfowner:
Change summary:
- Handle blockchain node initialization failure with BlockchainNodeStates.STOPPED
- Use class in logger implementation
- Finish logging instead of exiting (exitWithStackTrace() -> finishWithStackTrace()) 
- Include blockNumber in function context

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/664


liayoo (approved):
LGTM, thanks!

PR #666, opened 2021-11-03
Add enableHardCodedStateGC feature flag.

liayoo:
Added Add enableHardCodedStateGC feature flag that determines whether to garbage-collect old data (e.g. receipts, consensus, shard state proof hash reports). 
Will be removed or turned on once state proof hash bug is fixed (or state rule is implemented).

Related issues:


platfowner (approved):
LGTM
Thanks!

PR #669, opened 2021-11-05
Add full node snapshotting with enableFullNodeSnapshots flag

platfowner:
Summary:
- Add full node snapshotting with enableFullNodeSnapshots flag (fromSnapshotObject(), toSnapshotObject()) 
- Add prefix to both state label ('#state:') and radix label ('#radix:') in state proof

Related issues:


db/radix-node.js
```diff
@@ -546,13 +547,89 @@ class RadixNode {
     return numAffectedNodes;
   }
 
+  static parseChildStateNode(obj) {
```
liayoo, line 550: how about parseChildStateNodeFromSnapshot() ?

db/index.js
```diff
@@ -157,6 +164,18 @@ class DB {
     return true;
   }
 
+  /**
+   * Replaces the state root.
+   *
+   * @param {StateNode} newRoot new root to replace with
+   */
+  replaceStateRoot(newRoot) {
+    const newVersion = newRoot.getVersion();
```
liayoo, line 173: question) would there be a need to check if the newVersion doesn't have a conflicting version?

common/constants.js
```diff
@@ -34,6 +34,8 @@ const FeatureFlags = {
   enableTrafficMonitoring: true,
   // Enables state info updates.
   enableStateInfoUpdates: true,
+  // Enables full node snapshots.
+  enableFullNodeSnapshots: true,
```
liayoo, line 38: how about enableFullDbSnapshots or enableDetailedSnapshots?

db/index.js
platfowner: Just added the conflict check. Thanks!

db/radix-node.js
platfowner: Changed to getChildStateNodeFromSnapshotObject(). PTAL

common/constants.js
platfowner: Changed to enableRadixLevelSnapshots. PTAL

platfowner:
Comments addressed, PTAL! @liayoo

db/index.js
```diff
@@ -157,6 +164,23 @@ class DB {
...
+   * Replaces the state root.
+   *
+   * @param {StateNode} newRoot new root to replace with
+   */
+  replaceStateRoot(newRoot) {
+    const LOG_HEADER = 'replaceStateRoot';
+    const newVersion = newRoot.getVersion();
+    if (this.stateManager.hasVersion(newVersion)) {
+      CommonUtil.finishWithStackTrace(
+          logger, `[${LOG_HEADER}] State version already exists: ${newVersion}`);
```
liayoo, line 177: should we process.exit() here?

liayoo (approved):
LGTM with a comment. Thanks for handling this!

db/index.js
platfowner: Actually, finishWithStackTrace() tries to do similar with the some technical difficulties:
- If we call process.exit() directly somewhere in the code, the corresponding critical error messages might be skipped with the current winston logger's structure.
- So, we're just trying to stop logging after the critical log messages.

platfowner:

db/index.js
liayoo: Ah, right right. I was thinking of the old convention. Will keep that in mind, thanks!

PR #673, opened 2021-11-08
Feature/liayoo/function url wildcard

liayoo:
- Allow wildcards in REST function URLs using [match-url-wildcard](https://www.npmjs.com/package/match-url-wildcard)
- Updated URL validity checks to follow the WHATWG URL Standard ([node.js URL](https://nodejs.org/dist/latest-v16.x/docs/api/url.html#the-whatwg-url-api), [WHATWG valid url string](https://url.spec.whatwg.org/#valid-url-string))
- Examples
  - http://localhost:3000 => matches http://localhost:3000/*
  - http://127.0.0.1:3000 => matches http://127.0.0.1:3000/*
  - http://ainetwork.ai => matches http://ainetwork.ai/*
  - https://ainetwork.ai => matches https://ainetwork.ai/*
  - https://\*.ainetwork.ai => matches https://\*.ainetwork.ai/\*

### DIFF
--- a/unittest/rule-util.test.js
+++ b/unittest/rule-util.test.js
@@ -245,13 +245,41 @@ describe("RuleUtil", () => {
       expect(util.isValidUrl('0x')).to.equal(false);
       expect(util.isValidUrl('0x6af1ec8d4f0a55bac328cb20336ed0eff46fa6334ebd112147892f1b15aafc8')).to.equal(false);
       expect(util.isValidUrl('ainetwork.ai')).to.equal(false);
+      expect(util.isValidUrl('https://*.ainetwork.ai')).to.equal(false);
     })
 
     it("when valid input", () => {
       expect(util.isValidUrl('http://ainetwork.ai')).to.equal(true);
       expect(util.isValidUrl('https://ainetwork.ai')).to.equal(true);
-      expect(util.isValidUrl('https://*.ainetwork.ai')).to.equal(true);
       expect(util.isValidUrl('https://ainetwork.ai/some/api')).to.equal(true);
+    })
+  })
+
+  describe("isValidUrlWhitelistItem", () => {
+    it("when invalid input", () => {
+      expect(util.isValidUrlWhitelistItem(true)).to.equal(false);
+      expect(util.isValidUrlWhitelistItem(false)).to.equal(false);
+      expect(util.isValidUrlWhitelistItem(0)).to.equal(false);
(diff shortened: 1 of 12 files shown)


unittest/rule-util.test.js
```diff
@@ -227,6 +227,34 @@ describe("RuleUtil", () => {
...
+      expect(util.isValidUrl({a: 'A'})).to.equal(false);
+      expect(util.isValidUrl('0x')).to.equal(false);
+      expect(util.isValidUrl('0x6af1ec8d4f0a55bac328cb20336ed0eff46fa6334ebd112147892f1b15aafc8')).to.equal(false);
+      expect(util.isValidUrl('ainetwork.ai')).to.equal(false);
+    })
+
+    it("when valid input", () => {
+      expect(util.isValidUrl('http://ainetwork.ai')).to.equal(true);
+      expect(util.isValidUrl('https://ainetwork.ai')).to.equal(true);
+      expect(util.isValidUrl('https://*.ainetwork.ai')).to.equal(true);
```
platfowner, line 253: is this considered as a valid url?

liayoo: With the node.js's WHATWG URL API, it is considered valid, which works well with our new wildcard support for the function URL whitelist. 
On the other hand, for the function_url's in function configs, we could maybe apply a stricter rule?

liayoo: Ah, I see! In that case, no. For actual API calls, there shouldn't be asterisks in the domain names (only alphanumerics & '-' allowed). But for some reason, most URL validators (including node.js's URL) seem to be ok with them. For example, the old package we've been using, valid-url, would have validated `https://*.ainetwork.ai` to be a well-formed URL.
So, if we want to apply a stricter rule for the function_url, which is actually used for API calls, we could try some other methods (maybe something like [these](https://stackoverflow.com/questions/161738/what-is-the-best-regular-expression-to-check-if-a-string-is-a-valid-url)?)

platfowner: I see, thanks for the explanation. If so, I think we can have a separate function for validating whitelist items, e.g. isValidUrlWhitelistItem() that allows wildcards while keeping isValidUrl() not allowing wildcards. What do you think?

liayoo: Sounds great! Added isValidUrlWhitelistItem() and some more tests. PTAL~ @platfowner

platfowner (approved):
LGTM
Thanks for the nice feature!

Issue #614, opened 2021-10-12
[BUG] Rule attack with `process.exit()` function


cshcomcom:
**Describe the bug**
When I set the rule to `'write': 'process.exit(1)',` and send SET_VALUE tx, the node died.

**Screenshots**
![image](https://user-images.githubusercontent.com/74223287/136885833-ef138d14-d137-49a8-af8d-9f44ef61a34f.png)

platfowner:
Fixed.

PR #674, opened 2021-11-08
Add token-level write rule validity checking logic

platfowner:
Change summary:
- Add token-level validity checking logic for write rules using a javascript tokenization library (spree)
  (Notably only top-level tokens (e.g. auth, newData not auth.addr nor newData.proposer) are checked)
- Add tools/tokenizeRule.js util tool for rule tokenization
- Directory name change: diff-json -> json-diff

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/614


db/state-util.js
```diff
@@ -298,7 +375,7 @@ function isValidFunctionInfo(functionInfoObj) {
 /**
  * Checks the validity of the given function configuration.
  */
-function isValidFunctionConfig(functionConfigObj) {
+function isValidFunctionConfig(configPath, functionConfigObj) {
```
liayoo, line 378: Unused arg (configPath)?

db/state-util.js
```diff
@@ -352,15 +431,15 @@ function isValidOwnerPermissions(ownerPermissionsObj) {
...
+    logger.info(`[${LOG_HEADER}] Owner permission is in a non-standard format:\n${diffLines}\n`);
   }
   return isIdentical;
 }
 
 /**
  * Checks the validity of the given owner configuration.
  */
-function isValidOwnerConfig(ownerConfigObj) {
+function isValidOwnerConfig(configPath, ownerConfigObj) {
```
liayoo, line 442: ditto

db/state-util.js
```diff
@@ -401,30 +480,39 @@ function isValidOwnerConfig(ownerConfigObj) {
   return { isValid: true, invalidPath: '' };
 }
 
-function isValidConfigTreeRecursive(stateTreeObj, path, configLabel, stateConfigValidator) {
+/**
+ * Checks the validity of the given state tree with the given config label and
+ * state config validator.
+ * 
+ * @param {*} stateTreeObj state tree to check the validity
```
liayoo, line 487: missing @param for treePath?

db/state-util.js
```diff
@@ -198,6 +224,56 @@ function sanitizeRuleConfig(rule) {
...
+function getTopLevelIdTokens(tokenList) {
+  let withProceedingDot = false;
+  return tokenList.filter((token) => {
+    const isTopLevelIdToken = token.type === 'Identifier' && !withProceedingDot;
+    withProceedingDot = token.type === 'Punctuator' && token.value === '.';
+    return isTopLevelIdToken;
+  }).map((token) => token.value);
+}
+
+function isValidWriteRule(parsedRulePath, ruleString) {
```
liayoo, line 248: I'm thinking of adding this same function to ain-js so that users / insight can check write rules before sending SET_RULE txs. What do you think?!
cc: @minsulee2

db/state-util.js
platfowner: Right. For simplicity this function has the same parameters as isValidRuleConfig() to be used in isValidConfigTreeRecursive().

db/state-util.js
platfowner: ditto

db/state-util.js
platfowner: Just added. Thanks.

db/state-util.js
platfowner: I think that's a good idea. As we the blockchain is not returning friendly-enough message due to the complexity of code, what we can try quickly is just to copy the related parts to ain-js?

platfowner:

db/index.js
```diff
@@ -708,15 +710,15 @@ class DB {
...
-    const isValidObj = isValidJsObjectForStates(value);
-    if (!isValidObj.isValid) {
-      return CommonUtil.returnTxResult(101, `Invalid object for states: ${isValidObj.invalidPath}`, 1);
-    }
     const parsedPath = CommonUtil.parsePath(valuePath);
     const isValidPath = isValidPathForStates(parsedPath);
     if (!isValidPath.isValid) {
       return CommonUtil.returnTxResult(102, `Invalid path: ${isValidPath.invalidPath}`, 1);
     }
+    const isValidObj = isValidJsObjectForStates(value);
```
liayoo, line 718: question) are the orders changed to perform lighter checks first?

platfowner: Ah, it was just for aligning the order of code with setRule() where the order was slightly changed for reusing 'parsedPath' variable.
In my experience, bugs often occur with subtle disparities between similar functions, so I'm trying to avoid differences unless necessary.

liayoo (approved):
LGTM, thanks for handling the issue! 👍

platfowner:
Thanks for the review. Passed all tests and now being merged..

PR #675, opened 2021-11-09
account injection using ain-util

ehgmsdk20:
Changed account injection to use mnemonicToAccount of ain-util.


platfowner (approved):
LGTM
Thanks!

PR #677, opened 2021-11-10
Do not set IP_ADDR_LIST for no account injection options case

platfowner:
Change summary:
- Do not set IP_ADDR_LIST for no account injection options case

Background:
- The following error occurs without account injection options:
```
deploy_blockchain_genesis_gcp.sh: line 81: ./testnet_ip_addresses/summer.txt: No such file or directory
```


liayoo (approved):
LGTM, thanks!

Issue #658, opened 2021-11-01
Need API nodes separate from block proposers


platfowner:
We need separate API nodes to protect block proposers.

platfowner:
Done.

PR #678, opened 2021-11-11
Add ENABLE_JSON_RPC_API and revamp deploy scripts

platfowner:
Change summary:
- Add ENABLE_JSON_RPC_API
- Rename: ENABLE_DEV_SET_CLIENT_API -> ENABLE_DEV_CLIENT_SET_API
- Revamp deploy scripts:
  - Add --reset, --restart options to deploy_blockchain_genesis_gcp.sh, deploy_blockchain_incremental_gcp.sh
  - Remove reset scripts, restart scripts
  - Leave server start commands to start_commands.txt 
 
Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/658


deploy_blockchain_genesis_gcp.sh
```diff
@@ -15,63 +14,73 @@ if [[ "$1" = 'spring' ]] || [[ "$1" = 'summer' ]] || [[ "$1" = 'dev' ]] || [[ "$
...
 ACCOUNT_INJECTION_OPTION=""
+RESET_RESTART_OPTION=""
 
-if [[ "$#" -gt 3 ]]; then
-    parse_options "$4"
-    if [[ "$#" = 5 ]]; then
-        parse_options "$5"
-    fi
-fi
+number=4
```
liayoo, line 69: suggestion) how about `OPTION_INDEX`?

deploy_blockchain_genesis_gcp.sh
```diff
@@ -209,32 +221,72 @@ if [[ $SETUP_OPTION = "--setup" ]]; then
     gcloud compute ssh $NODE_6_TARGET_ADDR --command ". setup_blockchain_ubuntu.sh" --project $PROJECT_ID --zone $NODE_6_ZONE
 fi
 
-# ssh into each instance, install packages and start up the server
+printf "\nStarting blockchain servers...\n\n"
+if [[ $RESET_RESTART_OPTION = "--reset" ]]; then
```
liayoo, line 225: looks good~

common/constants.js
```diff
@@ -41,7 +41,10 @@ const FeatureFlags = {
 // ** Environment variables **
 const DEBUG = CommonUtil.convertEnvVarInputToBool(process.env.DEBUG);
 const CONSOLE_LOG = CommonUtil.convertEnvVarInputToBool(process.env.CONSOLE_LOG);
-const ENABLE_DEV_SET_CLIENT_API = CommonUtil.convertEnvVarInputToBool(process.env.ENABLE_DEV_SET_CLIENT_API);
+const ENABLE_DEV_CLIENT_SET_API =
```
liayoo, line 44: question) I believe we're planning to disable this by default in the future. When that happens, what would be the best way to check the status of the proposer nodes? Should we use the p2p_getRouteStatus for that as well, or maybe allow some status/health check APIs to be always enabled?

deploy_blockchain_genesis_gcp.sh
platfowner: Changed to ARG_INDEX. PTAL~

common/constants.js
platfowner: It's already disabled by default and explicitly set false in deploy scripts (dev, staging, prod). Note that it turns off only set APIs so I think get APIs like the status checking is not affected. Let me know if it's not the answer to your question.

platfowner:
All comments address, PTAL~ @liayoo

json_rpc/index.js
```diff
@@ -28,466 +29,473 @@ const PathUtil = require('../common/path-util');
...
-            code: 2,
-            message: `Missing properties.`
-          }
-        }));
-      } else {
-        const createdTx = Transaction.create(args.tx_body, args.signature);
-        if (!createdTx) {
+  // Default API methods
+  let methods = {
+    // TODO(minsulee2): Add p2p json rpc API methods here.
```
platfowner, line 34: fyi, @minsulee2

common/constants.js
liayoo: Ah, I was thinking we were going to disable or deprecate all client APIs. If we're going to leave the getter client APIs, then this makes sense!

liayoo (approved):
LGTM with a comment. Thanks! The scripts are getting prettier ✨

common/constants.js
platfowner: I think get APIs are not very harmful unless they consumes too much computation power or network bandwidth. So I think we need to check that before the big launch.

platfowner:
Addressed all comments visible so far, please let me know if there are any other hidden comments.

PR #679, opened 2021-11-11
Genesis account revamping with 100 genesis accounts

platfowner:
Change summary:
- Increase the number of sample genesis accounts from 10 to 100
- Move and update some tools
- Add NUM_GENESIS_ACCOUNTS to avoid unnecessary performance drop
- Move genesis timestamp and genesis shares to genesis params from genesis_accounts.json file


liayoo (approved):
LGTM, thanks!

Issue #680, opened 2021-11-12
[BUG] Need to make account injection json rpc APIs always available 


platfowner:
Without this, deployment --keystore or --mnemonic option fails with ENABLE_JSON_RPC_API=false configuration.

platfowner:
Fixed.

PR #681, opened 2021-11-12
Enable/disable JSON RPC transaction APIs with ENABLE_JSON_RPC_TX_API flag

platfowner:
Change summary:
- Enable/disable JSON RPC transaction APIs with ENABLE_JSON_RPC_TX_API flag
- Set NUM_GENESIS_ACCOUNTS=10
- Fix testnet genesis params file 

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/680


json_rpc/index.js
```diff
@@ -29,472 +29,481 @@ const PathUtil = require('../common/path-util');
...
+      trafficStatsManager.addEvent(TrafficEventTypes.JSON_RPC_SET);
+      let result = false;
+      if (await p2pServer.node.injectAccountFromHDWallet(args.encryptedMnemonic, args.index)) {
+        result = true;
+        p2pServer.client.run();
+      }
+      return addProtocolVersion({ result });
+    },
+
+    // TODO(minsulee2): Add p2p json rpc API methods here.
```
liayoo, line 409: should `p2p_getRouteStatus` be added under get methods or another group (e.g. p2p methods?)
cc: @minsulee2

platfowner: It's in `const systemSetMethods =`, which is always enabled regardless of ENABLE_JSON_API_APP_SET_API flag value.

liayoo: I think the naming (systemSetMethods) threw me off, since p2p_getRouteStatus seems to be a get API?

platfowner: Right, it was not clear enough. Tried to apply two groups: non-tx methods and tx methods with new flag name: ENABLE_JSON_RPC_TX_API.
PTAL~ @liayoo

liayoo: Thanks for the clarification! Looks good!

liayoo: + we should discuss which client & json-rpc APIs to enable on the API servers~

liayoo (approved):
LGTM, thanks!

platfowner:

PR #682, opened 2021-11-12
Add punctuator token blacklist for write rules

platfowner:
Change summary:
- Add punctuator token blacklist for write rules
- Do not allow assignment punctuator ('=')

Background:
- Value assignment could be harmful in javascript

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/614


liayoo (approved):
LGTM, thanks!

PR #686, opened 2021-11-12
Add /get_env dev client API for env variables

platfowner:
Change summary:
- Add /get_env dev client API for env variables


liayoo (approved):
LGTM, thanks!

Issue #687, opened 2021-11-15
[BUG] "No notarized block" error in restarting


platfowner:
Branch:
https://github.com/ainblockchain/ain-blockchain/pull/684

How to reproduce:
```bash start_node_genesis_gcp.sh summer 0 0 --keep-code```

Error logs:
```
2021-11-15T10:20:29.621Z [node-0-8080] error: [NODE] [loadAndExecuteChainOnDb] Failed to validate and execute block
 10001: ConsensusException: No notarized block at number 10000 with hash 0x832084c910d27b3c0db9662177d9b1188f04fd26
b395e9f7d22c70da9db8ee8b
    at Function.getPrevBlockInfo (/home/ain-blockchain/consensus/index.js:458:13)
    at Function.validateAndExecuteBlockOnDb (/home/ain-blockchain/consensus/index.js:767:37)
    at BlockchainNode.loadAndExecuteChainOnDb (/home/ain-blockchain/node/index.js:526:19)
    at BlockchainNode.init (/home/ain-blockchain/node/index.js:207:17)
    at P2pClient.startBlockchainNode (/home/ain-blockchain/p2p/index.js:272:29)
    at WebSocket.<anonymous> (/home/ain-blockchain/p2p/index.js:232:24)
    at WebSocket.emit (events.js:314:20)
    at Receiver.receiverOnMessage (/home/ain-blockchain/node_modules/ws/lib/websocket.js:983:20)
    at Receiver.emit (events.js:314:20)
    at Receiver.dataMessage (/home/ain-blockchain/node_modules/ws/lib/receiver.js:517:14)
Error
    at Function.logErrorWithStackTrace (/home/ain-blockchain/common/common-util.js:664:33)
    at Function.finishWithStackTrace (/home/ain-blockchain/common/common-util.js:674:16)
    at BlockchainNode.loadAndExecuteChainOnDb (/home/ain-blockchain/node/index.js:534:20)
    at BlockchainNode.init (/home/ain-blockchain/node/index.js:207:17)
    at P2pClient.startBlockchainNode (/home/ain-blockchain/p2p/index.js:272:29)
    at WebSocket.<anonymous> (/home/ain-blockchain/p2p/index.js:232:24)
    at WebSocket.emit (events.js:314:20)
    at Receiver.receiverOnMessage (/home/ain-blockchain/node_modules/ws/lib/websocket.js:983:20)
    at Receiver.emit (events.js:314:20)
[...]

PR #689, opened 2021-11-16
Add block and root_proof_hash to blockchain snapshots

platfowner:
Change summary:
- Add block to blockchain snapshots and add it to bc.chain[] in node restarting
- Add root_proof_hash to blockchain snapshots and verify it in node restarting 
- Finalize loaded state version in node restarting
- Remove enableRadixLevelSnapshots feature flag

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/687

### DIFF
--- a/db/index.js
+++ b/db/index.js
@@ -10,6 +10,7 @@ const {
   OwnerProperties,
   RuleProperties,
   StateInfoProperties,
+  BlockchainSnapshotProperties,
   ShardingProperties,
   GenesisAccounts,
   GenesisSharding,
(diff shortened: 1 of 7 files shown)


db/index.js
```diff
@@ -181,6 +183,9 @@ class DB {
     this.stateManager.setRoot(newVersion, newRoot);
     this.stateVersion = newVersion;
     this.stateRoot = newRoot;
+    if (!this.stateManager.finalizeVersion(newVersion)) {
```
liayoo, line 186: suggestion) how about renaming the function to replaceStateRootAndFinalize() or replaceStateRootAndFinalizeVersion() ?

db/index.js
```diff
@@ -181,6 +183,9 @@ class DB {
...
+      logger.error(`[${LOG_HEADER}] Failed to finalize version: ${newVersion}`);
```
liayoo, line 187: CommonUtil.finishWithStackTrace()?

liayoo (approved):
LGTM with some comments. Thanks for handling this issue!

db/index.js
platfowner: Replaced the function with resetDbWithSnapshot(). PTAL

db/index.js
platfowner: Done, thanks!

platfowner:

Issue #702, opened 2021-11-18
Need to improve randomness for proposer selection


liayoo:
The current seed for the pseudorandom number generator uses the last block's last_votes_hash and the epoch. 
One idea @platfowner suggested was to incorporate unpredictable values from validators by including a nonce/timestamp in votes.

platfowner:
Implementation in place.

PR #704, opened 2021-11-18
Improve randomness for proposer selection

liayoo:
- Added timestamps to votes to introduce more randomness in proposer selection
- Related: https://github.com/ainblockchain/ain-blockchain/issues/702


common/constants.js
```diff
@@ -221,6 +221,7 @@ const PredefinedDbPaths = {
   CONSENSUS_REWARDS_UNCLAIMED: 'unclaimed',
   CONSENSUS_REWARDS_CUMULATIVE: 'cumulative',
   CONSENSUS_STAKE: 'stake',
+  CONSENSUS_TIMESTAMP: 'timestamp',
```
platfowner, line 224: what about allowing a random string with CONSENSUS_VOTE_NONCE ('vote_nonce') or something?

db/rule-util.js
```diff
@@ -287,8 +287,12 @@ class RuleUtil {
...
   }
 
-  validateConsensusVoteData(newData, userAddr, blockHash, lastBlockNumber, getValue) {
-    if (!this.isDict(newData) || !this.isBool(newData.is_against) || !this.isNumber(newData.stake) || newData.block_hash !== blockHash) {
+  validateConsensusVoteData(newData, userAddr, blockHash, lastBlockNumber, currentTime, getValue) {
+    if (!this.isDict(newData) || !this.isBool(newData.is_against) || !this.isNumber(newData.stake)
+        || newData.block_hash !== blockHash) {
+      return false;
+    }
+    if (newData.timestamp !== currentTime) {
```
platfowner, line 295: If we allow a random string, I think we don't need to care about its value?

consensus/index.js
```diff
@@ -921,16 +926,18 @@ class Consensus {
...
       ref: PathUtil.getConsensusVotePath(proposalBlock.number, proposalBlock.hash, myAddr),
       value: {
         [PredefinedDbPaths.CONSENSUS_BLOCK_HASH]: proposalBlock.hash,
         [PredefinedDbPaths.CONSENSUS_STAKE]: validatorInfo.stake,
         [PredefinedDbPaths.CONSENSUS_IS_AGAINST]: false,
+        [PredefinedDbPaths.CONSENSUS_TIMESTAMP]: timestamp,
       }
     };
-    const voteTx = this.node.createTransaction({ operation, nonce: -1, gas_price: 1 });
+    const voteTx = this.node.createTransaction({ operation, nonce: -1, gas_price: 1, timestamp });
```
platfowner, line 940: Creating a vote tx with timestamp looks just fine.

consensus/index.js
```diff
@@ -960,9 +968,10 @@ class Consensus {
         [PredefinedDbPaths.CONSENSUS_STAKE]: validatorInfo.stake,
         [PredefinedDbPaths.CONSENSUS_IS_AGAINST]: true,
         [PredefinedDbPaths.CONSENSUS_OFFENSE_TYPE]: ValidatorOffenseTypes.INVALID_PROPOSAL,
+        [PredefinedDbPaths.CONSENSUS_TIMESTAMP]: timestamp,
```
platfowner, line 971: Using a timestamp as a vote nonce looks just fine too.

common/constants.js
liayoo: Oh, you mean another field (vote_nonce) on top of the timestamp?

platfowner: I mean replacing timestamp with nonce. It looks the current timestamp is included in 'value'. Couldn't the same value be extracted from the tx itself if necessary?

db/rule-util.js
liayoo: I thought it could help mitigate [grinding attacks](https://dyor-crypto.fandom.com/wiki/Grinding_Attack), although we are getting random values from multiple parties already. Also, since the name is 'timestamp', I thought it'd be reasonable to expect it to equal the tx's timestamp. 
As you said, we could introduce another field ('vote_nonce') and allow it to truly be random?

platfowner: Thanks for sharing the ideas. If the timestamp itself is set by the node runners, couldn't it be used in grinding attacks any ways? Also, if timestamp is included in vote txs, won't it automatically affect the random seed without being added to 'value' explicitly? Please correct me if it's wrong.

liayoo: Oh, you're absolutely right about the second point! I got confused with the computation of last_tx_hash...
Regarding the grinding attacks, I thought that it could be random enough but also has restricted space (kind of acting as a restriction to the grinding?). Anyway, since point #2, I'll replace the timestamp with vote_nonce. Thanks!

common/constants.js
liayoo: Right, updated to vote_nonce & changed getTimestampFromVoteTx() to use timestamp in tx_body!

liayoo:

platfowner (approved):
LGTM

consensus/index.js
```diff
@@ -838,6 +839,10 @@ class Consensus {
       // FIXME: ask for the block from peers
       return false;
     }
+    if (voteTimestamp < block.timestamp) {
+      logger.debug(`[${LOG_HEADER}] Invalid timestamp in vote: ${JSON.stringify(voteTx)}`);
```
minsulee2, line 843: logger.error()?

unittest/consensus.test.js
```diff
@@ -39,14 +39,16 @@ describe("Consensus", () => {
     );
     addBlock(node1, [tx], [], {});
     const lastBlock = node1.bc.lastBlock();
+    const timestamp = lastBlock.timestamp + 1;
```
minsulee2, line 42: Can I ask why plus 1 ?

minsulee2 (commented):
Left some comments!

consensus/index.js
liayoo: This is something other validators can trigger, so I try not to log as an error.

unittest/consensus.test.js
liayoo: Votes should naturally come after the block, thus the +1.

consensus/index.js
minsulee2: Thanks!

unittest/consensus.test.js
minsulee2: I see. Thanks for the explain!

minsulee2 (approved):
LGTM!

integration/function.test.js
```diff
@@ -28,28 +28,28 @@ const DB = require('../db');
 const ENV_VARIABLES = [
   {
     MIN_NUM_VALIDATORS: 4, ACCOUNT_INDEX: 0, DEBUG: false, CONSOLE_LOG: false,
-    ENABLE_DEV_CLIENT_SET_API: true, ENABLE_GAS_FEE_WORKAROUND: true,
+    ENABLE_DEV_CLIENT_SET_API: true, ENABLE_GAS_FEE_WORKAROUND: true, ENABLE_EXPRESS_RATE_LIMIT: false,
```
platfowner, line 31: Oh, thanks for fixing this! @liayoo

integration/blockchain.test.js
```diff
@@ -33,25 +33,25 @@ const { Block } = require('../blockchain/block');
 const ENV_VARIABLES = [
   {
     ACCOUNT_INDEX: 0, MIN_NUM_VALIDATORS: 4, DEBUG: false, CONSOLE_LOG: false,
-    ENABLE_DEV_CLIENT_SET_API: true, ENABLE_GAS_FEE_WORKAROUND: true,
+    ENABLE_DEV_CLIENT_SET_API: true, ENABLE_GAS_FEE_WORKAROUND: true, ENABLE_EXPRESS_RATE_LIMIT: false,
```
platfowner, line 36: Oh, thanks for fixing this!! @liayoo

liayoo:
Thanks for the reviews! Ran the tests, will be merging as soon as the performance test pipeline is fixed~

PR #709, opened 2021-11-19
Enhance address request and response

minsulee2:
Address signing and verification has already done a lot and peer disconnection if not pass some checks has been done as well. There are no many updates have to be done :)

Updates:
- Do not send peerInfo message if errors
- Check client ip address in the server side for safety
- Use p2pState
-


platfowner (approved):
LGTM
Thanks!

minsulee2:
Thanks for the review! merging now!

Issue #672, opened 2021-11-07
[BUG] null state versions


platfowner:
Some states in the db have null state versions:
e.g. http://localhost:8081/get_final_state_snapshot?include_version=true shows:
```
            "accounts": {
                "0xAAAf6f50A0304F12119D218b94bea8082642515B": {
                    "balance": 690000000,
                    "#version:balance": "START",
                    "#version": "START"
                },
                "0x00ADEc28B6a845a085e03591bE7550dd68673C1C": {
                    "balance": 900000,
                    "#version:balance": "START",
                    "#version": null
                },
                "0x01A0980d2D4e418c7F27e1ef539d01A5b5E93204": {
                    "balance": 900000,
                    "#version:balance": "START",
                    "#version": null
                },
                "0x02A2A1DF4f630d760c82BE07F18e5065d103Fa00": {
                    "balance": 1000000,
                    "#version:balance": "START",
                    "#version": null
                },
                "0x03AAb7b6f16A92A1dfe018Fe34ee420eb098B98A": {
                    "balance": 1000000,
                    "#version:balance": "START",
                    "#version": null
                },
                "0x04A456C92A880cd59D7145C457475515a6f6E0f2": {
                    "balance": 1000000,
                    "#version:balance": "START",
                    "#version": null
                },
                "0x05A1247A7400f0C2A893611adD1505743552c631": {
                    "balance": 1000000,
                    "#version:balance": "START",
                    "#version": null
                },
                "0x06AD9C8F611f1e9d9CACD4738167A51aA2e80a1A": {
                    "balance": 1000000,
                    "#version:balance": "START",
                    "#version": null
                },
                "0x07A43138CC760C85A5B1F115aa60eADEaa0bf417": {
                    "balance": 1000000,
                    "#version:balance": "START",
                    "#version": null
                },
                "0x08Aed7AF9354435c38d52143EE50ac839D20696b": {
                    "balance": 1000000,
                    "#version:balance": "START",
                    "#version": null
                },
                "0x09A0d53FDf1c36A131938eb379b98910e55EEfe1": {
                    "balance": 1000000,
                    "#version:balance": "START",
                    "#version": null
                },
                "#version": "START"
            },
```

platfowner:
Fixed.

PR #710, opened 2021-11-19
Fix null state version bug

platfowner:
Change summary:
- Fix null state version bug by setting radix tree versions properly
- Rename: FeatureFlags -> DevFlags

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/672


liayoo (approved):
LGTM, thanks!

Issue #698, opened 2021-11-18
Organize genesis params


minsulee2:
Some variables (like urls) are mixed in blockchain and network. We need to discuss and reorganize it in the right way.

platfowner:
Now we have blockchain_params.json and node_params.json.
Thanks!

PR #721, opened 2021-11-23
Introduce BlockchainConfigs for flags and constants

platfowner:
Change summary:
- Introduce BlockchainConfigs for flags and constants in constants.js
- Include BlockchainConfigs in /get_config result
- Remove duplicated env from getRuntimeInfo() result
- Use string instead of direct reg exp for CORS_WHITELIST
- Fix a undefined variable bug

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/698

### DIFF
--- a/common/common-util.js
+++ b/common/common-util.js
@@ -610,14 +610,14 @@ class CommonUtil {
    * @returns
    */
   static getTotalGasCost(gasPrice, gasAmount) {
-    const { MICRO_AIN } = require('./constants');
+    const { BlockchainConfigs } = require('./constants');
     if (!CommonUtil.isNumber(gasPrice)) {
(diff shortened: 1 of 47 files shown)


liayoo (approved):
LGTM, thanks!

common/common-util.js
```diff
@@ -779,6 +779,14 @@ class CommonUtil {
     }
     return jsonDiff.diffString(base, target, { color: "" });
   }
+
+  static getRegexpList(strList) {
```
liayoo, line 783: 👍👍

platfowner:
Thanks for the fast review. Will be merged immediately and further comments will be addressed later if available.

Issue #725, opened 2021-11-25
Peer address whitelisting


platfowner:
We need PEER_ADDR_WHITELIST flag in BlockchainConfigs for protecting (or isolating) the 5 block producers from public nodes. It needs to be applied to:
- Peer request rejection from the nodes not in the list
- Peer candidate filtering in peer discovery

platfowner:
Fixed by https://github.com/ainblockchain/ain-blockchain/pull/826
thanks!

PR #726, opened 2021-11-25
Support testnet-sandbox env

liayoo:
Background:
- A new gcp project named testnet-sandbox has been created. It will be initially used for the extensibility experiments @minsulee2 and I are conducting, but I think we can use it for future experiments as well!

Changes:
- Added blockchain_params.json, genesis_block.json.gz for testnet-sandbox
- Updated deploy & start gcp scripts to support testnet-sandbox
- Added a fix for the yarn permission issue in setup scripts

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/731
- https://github.com/ainblockchain/ain-blockchain/issues/727


blockchain-configs/testnet-sandbox/blockchain_params.json
```diff
@@ -0,0 +1,100 @@
...
+    }
+  },
+  "blockchain": {
+    "TRANSACTION_POOL_TIMEOUT_MS": 3600000,
+    "TRANSACTION_TRACKER_TIMEOUT_MS": 86400000,
+    "NETWORK_ID": 0,
+    "CHAIN_ID": 0,
+    "HOSTING_ENV": "gcp",
+    "TRACKER_WS_ADDR": "ws://35.221.146.70:5000",
+    "P2P_PEER_CANDIDATE_URL": "http://130.211.244.169:8080/json-rpc",
```
platfowner, line 28: what about using https://sandbox-api.ainetwork.ai/json-rpc? I can quickly set it up if you like.

liayoo: Thanks, that'd be nice! We figured we don't need API nodes or load balancing for the extensibility test, but we could just set it up now to match other envs.

liayoo: Oh and I just realized I used the node 0's URL here..

setup_blockchain_ubuntu.sh
```diff
@@ -20,6 +20,10 @@ printf 'npm --version\n'
 npm --version
 
 
+printf 'Fix yarn permission issues'
+sudo chown -R $USER:$GROUP ~/.npm
```
platfowner, line 24: what about trying to use '--userconfig=/dev/null' as suggested below?
https://stackoverflow.com/questions/64201625/ignore-local-npmrc-config-file-during-npm-install

blockchain-configs/testnet-sandbox/blockchain_params.json
liayoo: After I set up a load balancer, could you help me configure the subdomain?

platfowner: Right, once we have peer address whitelisting for nodes 0 ~ 4, we'll need to use a load balancer + hostname URL instead of the first node.
https://github.com/ainblockchain/ain-blockchain/issues/725

Please ping me once the load balancer part is done.

liayoo: Will do! Thanks! 🙌

minsulee2: Ohhhhh, Thanks for setting up!

setup_blockchain_ubuntu.sh
liayoo: Hmm.. I tried, but the option doesn't seem to work? (I tried `sudo npm install -g yarn --userconfig /dev/null` and `yarn -v --userconfig /dev/null`) What do you think about using yarn with sudo? (e.g. `sudo yarn install`)

platfowner: If that fixes the issue, it looks like a better approach than chown. Feel free to update so. Thanks!

liayoo: Updated~

platfowner (approved):
LGTM
Thanks!

Issue #730, opened 2021-11-25
[BUG] Cannot process the vote or it's invalid


platfowner:
Reset command line:
```
deploy_blockchain_incremental_gcp.sh summer <Username> 0 --reset
```

Error messages:
```
...
2021-11-25T15:09:15.821Z [node-8080] error: [CONSENSUS] [handleConsensusMessage] Already have the block in my block
 pool
2021-11-25T15:09:15.975Z [node-8080] error: [CONSENSUS] [handleConsensusMessage] Already have the block in my block
 pool
2021-11-25T15:09:27.821Z [node-8080] error: [CONSENSUS] No notarized block at number 118 with hash 0xaf79f54f7d57d4
5d03e5d00798dd325d56d9c315dd148b012927e112f20ec72e
2021-11-25T15:09:27.918Z [node-8080] error: [CONSENSUS] [handleConsensusMessage] Already have the block in my block
 pool
2021-11-25T15:09:27.991Z [node-8080] error: [CONSENSUS] [handleConsensusMessage] Already have the block in my block
 pool
2021-11-25T15:09:28.097Z [node-8080] error: [CONSENSUS] [handleConsensusMessage] Already have the block in my block
 pool
2021-11-25T15:09:42.705Z [node-8080] error: [CONSENSUS] [handleConsensusMessage] Cannot process the vote or it's in
valid: {"tx_body":{"operation":{"type":"SET_VALUE","ref":"/consensus/number/130/0xa474755fd6b1d811cb5b0d5d3dabed044
d98ff5134b246479b1735eaef204419/vote/0x003AD6FdB06684175e7D95EcC36758B014517E4b","value":{"block_hash":"0xa474755fd
6b1d811cb5b0d5d3dabed044d98ff5134b246479b1735eaef204419","stake":10000000,"is_against":false,"vote_nonce":163785298
2303}},"nonce":-1,"gas_price":1,"timestamp":1637852982303},"signature":"0x8bb035b82a01c0478dd75e8aad22da952593fa5cc
f75833ae804360437d2f3ec67d427337d78c67e0d522dbfcbd433ade128d42adc1352c169fc1d3f27de5f1b697a8f1f38a7ac956b7bb130b001
7e273a63bda9c00aa59e8d03bc322a9571d11b","hash":"0x8bb035b82a01c0478dd75e8aad22da952593fa5ccf75833ae804360437d2f3ec"
[...]

PR #737, opened 2021-11-26
Bugfix/liayoo/cannot process vote issue

liayoo:
- Reduced "Cannot process vote.." errors by checking the block number in the vote earlier
  - May not fix the root cause though
- Tried to use the catch-up info received even though you think you're fully synced (L570 in p2p/index.js)
  - May solve the root of the issue. We'll see!
  - Randomly refreshing peer connection still seems like a good tool to have~
- Fixed my URL ending up in inbound, outbound peer lists

Related: #730


block-pool/index.js
```diff
@@ -30,13 +30,24 @@ class BlockPool {
...
     this.numberToBlockSet = {};
+    this.heighestSeenBlockNumber = -1;
   }
 
   getLongestNotarizedChainHeight() {
     return this.longestNotarizedChainTips.length > 0 ?
         this.hashToBlockInfo[this.longestNotarizedChainTips[0]].block.number : this.node.bc.lastBlockNumber();
   }
 
+  setHeighestSeenBlockNumber(blockNumber) {
```
platfowner, line 41: minor: what about 'updateHighestSeenBlockNumber()' as it's not always overwriting the value?

block-pool/index.js
```diff
@@ -291,6 +302,7 @@ class BlockPool {
       this.numberToBlockSet[block.number] = new Set();
     }
     this.numberToBlockSet[block.number].add(block.hash);
+    this.setHeighestSeenBlockNumber(block.number);
```
platfowner, line 305: could you consider moving this to one level upper (i.e. to addSeenBlock())? The naming seems to be well matched (addSeenBlock <-> setHighestSeenBlockNumber())

consensus/consensus-util.js
```diff
@@ -63,6 +63,20 @@ class ConsensusUtil {
     }
   }
 
+  static getBlockNumberFromConsensusTx(tx) {
+    const op = _get(tx, 'tx_body.operation');
+    if (!tx || !op) return null;
```
platfowner, line 68: minor: what about splitting into two if-blocks (one for tx the other for op)?

p2p/index.js
```diff
@@ -631,7 +632,8 @@ class P2pClient {
    * @param {string} peerCandidateUrl should be something like http(s)://xxx.xxx.xxx.xxx/json-rpc
    */
   async connectWithPeerCandidateUrl(peerCandidateUrl) {
-    if (!peerCandidateUrl || peerCandidateUrl === '') {
+    const myUrl = _.get(this.server.urls, 'p2p.url', '');
+    if (!peerCandidateUrl || peerCandidateUrl === '' || peerCandidateUrl === myUrl) {
```
platfowner, line 636: nice! now maybe the exceptional P2P_PEER_CANDIDATE_URL setting for the api nodes in the deployment scripts can be reverted.

p2p/index.js
```diff
@@ -647,24 +649,20 @@ class P2pClient {
       logger.error(`Invalid peer candidate json rpc url from peer candidate url (${peerCandidateUrl}).`);
       return;
     }
-
     this.peerCandidates[peerCandidateJsonRpcUrl] = { queriedAt: Date.now() };
     const peerCandidateUrlList = _.get(peerCandidateInfo, 'peerCandidateUrlList', []);
-    peerCandidateUrlList.forEach(url => {
-      if (!this.peerCandidates[url] && this.isValidJsonRpcUrl(url)) {
+    peerCandidateUrlList.forEach((url) => {
+      if (url !== myUrl && !this.peerCandidates[url] && this.isValidJsonRpcUrl(url)) {
```
platfowner, line 655: could you apply this checking to the peerCandidateJsonRpcUrl above (line 652) as well?

p2p/index.js
platfowner, line 636: peerCandiateUrl is a json rpc url, so it needs to be compared with 'jsonRpc.url'?

p2p/server.js
```diff
@@ -580,8 +580,12 @@ class P2pServer {
             const updatePeerInfo = parsedMessage.data;
             const addressFromSocket = getAddressFromSocket(this.inbound, socket);
             // Keep updating both inbound and outbound.
-            this.inbound[addressFromSocket].peerInfo = updatePeerInfo;
-            this.client.outbound[addressFromSocket].peerInfo = updatePeerInfo;
+            if (this.inbound[addressFromSocket]) {
```
minsulee2, line 583: Thanks for adding this!

p2p/index.js
minsulee2: Oh, that's right! But it also be considered that `localhost` and `172.30.1.1` is in the same manner.

block-pool/index.js
liayoo: SGTM, thanks!

block-pool/index.js
liayoo: Thanks for the nice idea! Updated!

consensus/consensus-util.js
liayoo: Done~

p2p/index.js
liayoo: Thanks for catching it! Done!

p2p/index.js
liayoo: Right, I've updated the names to better distinguish the two as well. Thanks!!

liayoo: Ah, thanks for the idea! I've added `P2P_PEER_CANDIDATE_URL  === myJsonRpcUrl` to the genesis node conditions & removed `P2P_PEER_CANDIDATE_URL=''` in the start script. However, it still seems necessary to keep the empty string settings for local scripts since P2P_PEER_CANDIDATE_URL is set as localhost whereas the p2p url uses the actual ip address.

platfowner: Great, thanks!
This is somewhat out of the scope but can P2P_PEER_CANDIDATE_URL be changed to just PEER_CANDIDATE_JSON_RPC_URL (adding 'json rpc' explicitly)? 
cc: @minsulee2

platfowner: Great to know that, thanks!

platfowner (approved):
LGTM with some additional comments.
Thanks!

p2p/index.js
liayoo: Ah, that's a good idea! Changed the name to `PEER_CANDIDATE_JSON_RPC_URL`~

liayoo:

PR #738, opened 2021-11-29
Add checkout fees, update min/max amounts, ..

liayoo:
- Added checkout_fee_rate (0.1%)
- Increased min & max checkout amounts
- Changed eth token bridge configs for testnet-* from eth mainnet to ropsten
- Added a blockchain_params.json placeholder for the mainnet


minsulee2 (approved):
LGTM tho. What about double checking with @platfowner ?

blockchain-configs/afan-shard/blockchain_params.json
```diff
@@ -9,9 +9,10 @@
           "0": {
             "network_id": 1,
             "token_pool": "0xBBB2219cD5eACc54Ce95deF7a67dDe71C8241891",
-            "min_checkout_per_request": 1000,
-            "max_checkout_per_request": 10000,
+            "min_checkout_per_request": 10000,
+            "max_checkout_per_request": 100000,
```
platfowner, line 13: question: is this correct that we don't need per-day max as a user cannot create multiple pending requests?

db/functions.js
```diff
@@ -1193,8 +1194,9 @@ class Functions {
     if (amountValidated !== true) {
       return this.returnFuncResult(context, amountValidated);
     }
+    const fee = amount * checkoutFeeRate;
```
platfowner, line 1197: minor: what about defining e.g. 'transferAmount' like 'refundAmount' here or uring 'fee' in _closeCheckout as well?

platfowner (approved):
LGTM with some minor comments and questions.
Thanks!

blockchain-configs/afan-shard/blockchain_params.json
liayoo: It's actually used for check-outs. Check-ins do not allow a user to create multiple pending requests with the same <network name, chain id, token id, sender address>, but for check-outs, there are no such limits.

platfowner: Got it, thanks!

db/functions.js
liayoo: Defined transferAmount to match refundAmount. Thanks!

liayoo:

PR #740, opened 2021-11-29
Fix lastBlockNumber fetch in wait_until_node_synced_gcp.sh

platfowner:
Change summary:
- Fix lastBlockNumber fetch in wait_until_node_synced_gcp.sh

### DIFF
--- a/json_rpc/index.js
+++ b/json_rpc/index.js
@@ -64,8 +64,7 @@ module.exports = function getMethods(node, p2pServer, eventHandler, minProtocolV
 
     ain_getRecentBlockNumber: function(args, done) {
       trafficStatsManager.addEvent(TrafficEventTypes.JSON_RPC_GET);
-      const block = node.bc.lastBlock();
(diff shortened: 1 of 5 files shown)


json_rpc/index.js
```diff
@@ -64,8 +64,7 @@ module.exports = function getMethods(node, p2pServer, eventHandler, minProtocolV
 
     ain_getRecentBlockNumber: function(args, done) {
```
platfowner, line 65: @liayoo Just realized that we can use similar names for both json rpc api and dev client api. which one do you prefer among 'last block number' and 'recent block number'?

liayoo: Ah, thanks for pointing it out! I like "last block number", since we've been using the term more than "recent block number"~

liayoo (approved):
LGTM, thanks!

json_rpc/index.js
platfowner: Got it, I'll make changes so.

platfowner: Done, thanks!

PR #753, opened 2021-12-02
fix HE protocol integration test

shyun-comcom:
Fix HE protocol integration test
* fix typo
* Update function call gas amount


platfowner (approved):
LGTM
Thanks for fixing this!

PR #754, opened 2021-12-02
Feature/liayoo/on chain blockchain params

liayoo:
- Added non-overwritable blockchain params to the state (on-chain)
- Blockchain params can be accessed with getBlockchainParam() and getAllBlockchainParamsFromState()
- Separated NodeConfiigs from BlockchainConfigs (NodeConfigs = each node can set differently. Basically env vars)
- Re-org'ed blockchain_params.json
- Changed blockchain param names to lower cases
- Introduced blockchain params for 1-node, 2-nodes, 3-nodes 

TODOs
- Use the on-chain values instead of GenesisToken/GenesisSharding
- Add TX_POOL_* as env vars (which will become overwritable & included in the NodeConfigs)
- Update README.md
- Deprecate ACCOUNT_INDEX


node/index.js
```diff
@@ -387,15 +391,43 @@ class BlockchainNode {
...
-      result.limit = BlockchainConfigs.TX_POOL_SIZE_LIMIT;
+      result.limit = this.getBlockchainParam(BlockchainParamsCategories.RESOURCE, 'tx_pool_size_limit');
       result.used = this.tp.getPoolSize();
     }
     return result;
   }
 
+  getBlockchainParam(category, name, blockNumber = null, stateVersion = null) {
+    return DB.getBlockchainParam(category, name,
+        blockNumber !== null ? blockNumber : this.bc.lastBlockNumber(),
```
platfowner, line 406: This seems to be an out-of-scope topic, but what do you think about changing 'lastBlockNumber' to 'finalBlockNumber'? in the contest of this change (i.e., final block number with final root)?

blockchain-configs/base/genesis_rules.json
```diff
@@ -11,6 +11,11 @@
       }
     }
   },
+  "blockchain_params": {
+    ".rule": {
+      "write": "util.isAppAdmin('consensus', auth.addr, getValue) === true"
```
platfowner, line 16: isAppAdmin() not blockchain owner!!

blockchain-configs/base/blockchain_params.json
```diff
@@ -20,33 +20,20 @@
...
+    "chain_id": 0
   },
   "consensus": {
-    "MIN_STAKE_PER_VALIDATOR": 10000000,
-    "MAX_STAKE_PER_VALIDATOR": 10000000,
-    "MIN_NUM_VALIDATORS": 3,
-    "MAX_NUM_VALIDATORS": 9,
-    "EPOCH_MS": 3000,
-    "GENESIS_WHITELIST": {
+    "min_stake_per_validator": 10000000,
```
platfowner, line 26: as discussed offline what about something like 'min_stake_for_propose' and 'max_...'?

blockchain-configs/base/blockchain_params.json
```diff
@@ -61,31 +48,47 @@
...
+    "genesis_timestamp": 1571819318272,
+    "num_genesis_accounts": 10
+  },
   "network": {
-    "P2P_MESSAGE_TIMEOUT_MS": 600000,
-    "TARGET_NUM_OUTBOUND_CONNECTION": 3,
-    "MAX_NUM_INBOUND_CONNECTION": 6,
-    "REQUEST_BODY_SIZE_LIMIT": "100mb"
+    "network_id": 0,
+    "p2p_message_timeout_ms": 600000,
```
platfowner, line 58: as discussed offline, this can be moved to NodeConfigs?

common/constants.js
```diff
@@ -74,10 +65,64 @@ BlockchainConfigs.DEFAULT_CORS_WHITELIST = [
...
+if (!fs.existsSync(NodeConfigs.BLOCKCHAIN_DATA_DIR)) {
+  fs.mkdirSync(NodeConfigs.BLOCKCHAIN_DATA_DIR, { recursive: true });
+}
+
 
+// ** Blockchain configs **
+const BlockchainConfigs = {};
+// *** Genesis ***
+BlockchainConfigs.BASE_BLOCKCHAIN_CONFIGS_DIR = 'blockchain-configs/base';
+BlockchainConfigs.CUSTOM_BLOCKCHAIN_CONFIGS_DIR = process.env.BLOCKCHAIN_CONFIGS_DIR ?
```
platfowner, line 78: as discussed offline, need to be moved to NodeConfigs?

node/index.js
```diff
@@ -387,15 +391,43 @@ class BlockchainNode {
...
+
+  getAllBlockchainParamsFromState() {
+    const params = DB.getValueFromStateRoot(
+        this.stateManager.getFinalRoot(), PathUtil.getBlockchainParamsRootPath()) || {};
+    const token = DB.getValueFromStateRoot(this.stateManager.getFinalRoot(), PredefinedDbPaths.TOKEN) || {};
+    const sharding = DB.getValueFromStateRoot(this.stateManager.getFinalRoot(), PredefinedDbPaths.SHARDING) || {};
+    return Object.assign(params, { token }, { sharding });
+  }
+
+  getBandwidthBudgets() {
```
platfowner, line 418: as discussed offline, this can be moved to tx-pool/index.js?

common/constants.js
```diff
@@ -74,10 +65,64 @@ BlockchainConfigs.DEFAULT_CORS_WHITELIST = [
...
+  network: {
+    tracker_update_json_rpc_url: true,
+    peer_candidate_json_rpc_url: true,
+    target_num_outbound_connection: true,
+    max_num_inbound_connection: true,
+  }
+  // TODO(liayoo): Add tx_pool_* resource params
+};
+
+function setOverwritableBlockchainParams() {
```
platfowner, line 105: as discussed offline:
- extract node_params.json in a flat structure from blockchain_params.json 
- BlockchainConfigs -> BlockchainConsts 
- keep using blockchain_params.json and getBlockchainParam()

node/index.js
platfowner: cc: @minsulee2

liayoo: Hmm, that sounds nice, but also sounds like a huge change in itself! :) (APIs, code, pipeline, insight, ...) I'll leave a TODO for now and work on it as a next item (& let other people know about the change!)

blockchain-configs/mainnet/blockchain_params.json
```diff
@@ -20,59 +20,62 @@
     }
```
platfowner, line 20: minor: it seems this config file needs to be cloned (maybe from testnet-prod) and addjusted again when we're really ready for the mainnet launch. so what about not paying much attention to it for now by dropping it?

blockchain-configs/token-bridge/blockchain_params.json
```diff
@@ -0,0 +1,85 @@
+{
```
platfowner, line 1: question: what is the use case of this config file?

In this context, what do you think about adding a dedicated field, e.g., "description" for briefly describing the config file?

client/index.js
```diff
@@ -716,7 +719,10 @@ app.get('/get_chain_id', (req, res) => {
   trafficStatsManager.addEvent(TrafficEventTypes.CLIENT_API_GET, latency);
   res.status(200)
     .set('Content-Type', 'application/json')
-    .send({ code: 0, result: BlockchainConfigs.CHAIN_ID })
+    .send({
+      code: 0,
+      result: node.getBlockchainParam(BlockchainParamsCategories.BLOCKCHAIN, 'chain_id')
```
platfowner, line 724: could you move this line in between `const beginTime = ...` and `const latency = ...`?

common/network-util.js
```diff
@@ -63,8 +63,8 @@ async function sendSignedTx(endpoint, params) {
 }
 
 // FIXME(minsulee2): this is duplicated function see: ./tools/util.js
-async function signAndSendTx(endpoint, tx, privateKey) {
-  const { txHash, signedTx } = CommonUtil.signTransaction(tx, privateKey, BlockchainConfigs.CHAIN_ID);
+async function signAndSendTx(endpoint, tx, privateKey, chainId) {
```
platfowner, line 66: thanks for fixing this! 
cc: @minsulee2

blockchain-configs/mainnet/blockchain_params.json
liayoo: Right, that's why I haven't created a genesis block for it. I'll just leave the folder empty for now then! Thanks!

blockchain-configs/token-bridge/blockchain_params.json
liayoo: Ah, this is a config I've been using while developing the token bridge servers. I'll remove this and add a description for each param file~

client/index.js
liayoo: Done!

common/constants.js
liayoo: Done!

common/constants.js
liayoo: Done!

blockchain-configs/base/blockchain_params.json
liayoo: Done!

blockchain-configs/base/blockchain_params.json
liayoo: Done!

blockchain-configs/base/genesis_rules.json
liayoo: Right!! (Discussed offline to keep using the consensus app admin, which is initially the genesis_addr)

integration/consensus.test.js
```diff
@@ -175,6 +176,31 @@ describe('Consensus', () => {
   });
 
   describe('Validators', () => {
+    before(async () => {
+      // Update max_num_validators to 4
```
platfowner, line 180: 👍

common/network-util.js
liayoo: Ah, actually chain id was already in place, but there are some other functions that need to be fixed. Let me quickly fix them..!

node/index.js
platfowner: fyi, just learned in R48 that 'ain_getLastBlockNumber' is used by the ain-benchmark (see https://github.com/ainblockchain/ain-benchmark/pull/57) and the insight (constants/const.js).
cc: @cshcomcom @minsulee2

p2p/index.js
```diff
@@ -53,25 +52,27 @@ class P2pClient {
     this.p2pState = P2pNetworkStates.STARTING;
     this.peerConnectionsInProgress = {};
     this.chainSyncInProgress = null;
+    this.epochMs = node.getBlockchainParam(BlockchainParamsCategories.CONSENSUS, 'epoch_ms');
```
platfowner, line 55: In such a situation that epoch_ms  was changed by the consensus admin, the blockchain node owner (or operator) needs to restart the node?
In this context, what do you think about moving the getBlockchainParam() function calls in each function (no restart needed)?

tools/genesis-file/createGenesisBlock.js
```diff
@@ -418,16 +473,26 @@ function processArguments() {
...
-  console.log('  MIN_NUM_VALIDATORS      The minimum number of validators.');
-  console.log('\nExample:');
-  console.log('  BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/base node createGenesisBlock.js');
+  console.log('\nExamples:');
+  console.log('  node createGenesisBlock.js'); // Same as BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/base
+  console.log('  BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/1-node node createGenesisBlock.js');
+  console.log('  BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/2-nodes node createGenesisBlock.js');
+  console.log('  BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/3-nodes node createGenesisBlock.js');
+  console.log('  BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/afan-shard node createGenesisBlock.js');
+  console.log('  BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/he-shard node tools/genesis-file/createGenesisBlock.js');
```
platfowner, line 489: does this mean that this script can be run from ain-blockchain not from ain-blockchain/toos/genesis-file?

node/index.js

p2p/index.js
liayoo: I thought that epoch_ms should not be changed without a protocol version upgrade & node restart as it's a critical component of the consensus process.. curious to hear your thoughts!

tools/genesis-file/createGenesisBlock.js
liayoo: That's correct~

common/constants.js
```diff
@@ -74,85 +65,135 @@ BlockchainConfigs.DEFAULT_CORS_WHITELIST = [
...
+NodeConfigs.PEER_CANDIDATE_JSON_RPC_URL = process.env.PEER_CANDIDATE_JSON_RPC_URL;
+NodeConfigs.TX_POOL_SIZE_LIMIT = process.env.TX_POOL_SIZE_LIMIT ?
+    Number(process.env.TX_POOL_SIZE_LIMIT) : undefined;
+NodeConfigs.TX_POOL_SIZE_LIMIT_PER_ACCOUNT = process.env.TX_POOL_SIZE_LIMIT_PER_ACCOUNT ?
+    Number(process.env.TX_POOL_SIZE_LIMIT_PER_ACCOUNT) : undefined;
+NodeConfigs.TX_POOL_TIMEOUT_MS = process.env.TX_POOL_TIMEOUT_MS ?
+    Number(process.env.TX_POOL_TIMEOUT_MS) : undefined;
+NodeConfigs.TX_TRACKER_TIMEOUT_MS = process.env.TX_TRACKER_TIMEOUT_MS ?
+    Number(process.env.TX_TRACKER_TIMEOUT_MS) : undefined;
+const OverwritableNodeParamsDefaultValues = {
```
platfowner, line 100: Could you consider this?
- Move these default values to blockchain-configs/base/node_params.json
- Use the node_params.json as a base as it used to be with overwriteBlockchainParams(), and overwrite its values with the env variables
- Let the node_params.json at the custom blockchain configs dir be used if available instead of the node_params.json at the base blockchain configs dir

This might give simplicity in the inheritance structure of the config values, i.e., base node_params.json -> custom node_params.json -> env variables?

blockchain-configs/base/blockchain_params.json
```diff
@@ -20,33 +20,20 @@
...
-    "MIN_NUM_VALIDATORS": 3,
-    "MAX_NUM_VALIDATORS": 9,
-    "EPOCH_MS": 3000,
-    "GENESIS_WHITELIST": {
+    "min_stake_for_proposer": 10000000,
+    "max_stake_for_proposer": 10000000,
+    "min_num_validators": 3,
+    "max_num_validators": 9,
+    "epoch_ms": 3000,
+    "genesis_whitelist": {
```
platfowner, line 31: as discussed offline, if this means a whitelist for proposers, what about 'genesis_proposer_whitelist' more explicitly? 
And if it's acceptable, changing `CONSENSUS_WHITELIST: 'whitelist',` to `CONSENSUS_PROPOSER_WHITELIST: 'proposer_whitelist',` as well?

p2p/index.js
platfowner: Thanks for sharing the ideas.
Just checked that all the on-chain params saved locally as of now are:
- genesisAddr, epochMs, networkId, chainId

I totally agree that these shouldn't be changed in the future without protocol version upgrades.
If the overhead is negligible, however would it be safer not to have such coding patterns for future developers who might not be aware of such background?

tools/genesis-file/createGenesisBlock.js
platfowner: good to know that!

db/index.js
```diff
@@ -537,6 +539,26 @@ class DB {
     };
   }
 
+  static getBlockchainParam(category, name, blockNumber, stateRoot) {
```
platfowner, line 542: minor: would it make sense to get a combined path (e.g. 'consensus/epoch_ms') of category and name directly for developer's convenience by defining `getBlockchainParam(paramPath, blockNumber, stateRoot)`?
e.g. DB.getBlockchainParam(BlockchainParamsCategory.CONSENSUS, 'epoch_ms', ...) -> DB.getBlockchainParam('consensus/epoch_ms', ...)?

node/index.js
```diff
@@ -387,15 +391,30 @@ class BlockchainNode {
...
       result.used = this.tp.getPoolSize();
     }
     return result;
   }
 
+  // TODO(liayoo): Rename lastBlockNumber to finalBlockNumber.
+  getBlockchainParam(category, name, blockNumber = null, stateVersion = null) {
+    return DB.getBlockchainParam(category, name,
+        blockNumber !== null ? blockNumber : this.bc.lastBlockNumber(),
+        stateVersion ? this.stateManager.getRoot(stateVersion) : this.stateManager.getFinalRoot());
```
platfowner, line 407: This is just a question for future discussion (maybe in the next meeting?):
- Would it be better to use the state root of the previous block of the current block in the block pool if we can make it?

p2p/index.js
liayoo: Hmm. Should those be considered exceptions & not added to the state then, and always use the genesis values?

common/constants.js
liayoo: Yeah, I actually went for your method first and realized that it was nicer to have the fields explicitly defined in constants.js since the editor lets us know if we're importing correctly that way. Come to think of it, I guess having the same values & structure to each node_params.json and having a base json file can be better for future devs. Reverted to the original design. Thanks for your input!

blockchain-configs/base/blockchain_params.json
liayoo: Done & done~

db/index.js
liayoo: Done. Removing the category enum actually made the code much simpler. Thanks!

p2p/index.js
platfowner: Oh, I assumed such a case that a new developer tried to follow the pattern of saving the value in the constructor for a less sensitive parameter like 'resource/batch_tx_list_size_limit' and then the committee decided to increase its value by the consensus admin. 
In this case, the code needs to be updated and deployed to the most of the blockchain nodes to work properly with the new param value.
I was just thinking of avoiding such scenarios by applying a kind of rule that blockchain params need to be fetched always online. please let me know if you think it's too much though..

liayoo: Ah, no I think that's fair! I updated to always use getBlockchainParam(). What do you think about treating `genesis_addr, epoch_ms, network_id, chain_id` as exceptions within the function though? In order to make sure that we use the genesis values?

node/index.js
liayoo: Oh, it took me a while to get it, but it totally makes sense! Since the get param is for a specific block, not any incoming txs.. I updated to use the tempDb's blockNumberSnapshot and stateVersion (which will essentially be the same as the state that we're basing a new block from)

p2p/index.js
platfowner: Oh, I think now I understand what you mean. If we keep those values in BlockchainConsts instead of blockchain-params.json, they cannot be changed by just state writing. Instead, such changes need to be included in a source code version release which explicitly requires node restart. 

I think it's a good idea for at least the three parameters: genesisAddr, networkId, and chainId.

BTW, do you know what actually happens when we change epochMs online, e.g. 3 secs to 4 secs? What about 3 secs to 3.1 secs? Would it be a good test case of the sandbox experiment? or obviously be just a disaster to the network?

common/constants.js
platfowner: Great, thanks!

common/constants.js
```diff
@@ -24,48 +24,39 @@ const DevFlags = {
...
-    Number(process.env.MAX_BLOCK_NUMBERS_FOR_RECEIPTS) : 1000;
-BlockchainConfigs.ACCOUNT_INJECTION_OPTION = process.env.ACCOUNT_INJECTION_OPTION || null;
-BlockchainConfigs.KEYSTORE_FILE_PATH = process.env.KEYSTORE_FILE_PATH || null;
-BlockchainConfigs.ENABLE_STATUS_REPORT_TO_TRACKER =
+NodeConfigs.ACCOUNT_INDEX = process.env.ACCOUNT_INDEX || null;
+NodeConfigs.PORT = process.env.PORT || getPortNumber(8080, 8080);
+NodeConfigs.P2P_PORT = process.env.P2P_PORT || getPortNumber(5000, 5000);
+NodeConfigs.EVENT_HANDLER_PORT = process.env.EVENT_HANDLER_PORT || getPortNumber(6000, 6000);
+NodeConfigs.LIGHTWEIGHT = CommonUtil.convertEnvVarInputToBool(process.env.LIGHTWEIGHT);
+NodeConfigs.SYNC_MODE = process.env.SYNC_MODE || 'full';
```
platfowner, line 51: Would it be possible 1) to add all the NodeConfigs params in this file to node-params.json with their default values (e.g. SYNC_MODE='full') and 2) let overwriteNodeConfigs() does the env value overwriting? 

It looks like there are many exceptional cases for the overwriting e.g. CommonUtil.convertEnvVarInputToBool() for boolean values, so for 2), I think we have some quick options:
a) Just don't try to make one magical function do all of them. Instead just rely on manual settings for necessary params only as we're doing in this file now (, i.e., just drop overwriteNodeConfigs())
b) Move the setting logic here into overwriteNodeConfigs() with if-else branching, e.g.
```
  const NodeConfigs = getBlockchainConfig('node_params.json');
    for (const [param, valFromNodeParams] of Object.entries(NodeParams)) {
    const valFromEnvVar = process.env[param];
    if (param === 'DEBUG') {
      NodeConfigs[param] = CommonUtil.convertEnvVarInputToBool(valFromEnvVar);
    } else if (...) {
    ...
    } else {
      NodeConfigs[param] = CommonUtil.isIntegerString(valFromEnvVar) ? Number(valFromEnvVar) : valFromEnvVar;
    }
```
?

node/index.js
platfowner: Oh, I think you don't have to change anything in this PR regarding this topic. It's just a kind of pre-check for my rough ideas for a refactoring in the future: 
- Adding prev block info to the block pool?
- Officially adding block number and the epoch to state version?
- Moving database under state version manager?

p2p/index.js
platfowner: (offline discussion) 
It looks like change of epochMs won't be easy (almost impossible?).
Then some questions we need to address regarding epochMs at this moment:
- When the network size increases dramatically (e.g. to 10k nodes), can this be guaranteed that the 3 secs epochMs won't be a limit to the network expansion?
  - Can we make sure that the voters (validators of high staking values) will do their jobs well in 3 secs?
  - Can we make sure that non-voter validators (validators of low staking values) won't be mixed up with the delayed p2p messages?

liayoo: I've moved the unchangeable blockchain params to BlockchainConsts (still set in blockchain_params.json though, for convenience..) PTAL!

common/constants.js
liayoo: Great idea! I've added all of them to the node_params.json, except for the BLOCKCHAIN_CONFIGS_DIR and GENESIS_BLOCK_DIR (needed to get the node_params.json file itself)

liayoo:

blockchain-configs/base/node_params.json
```diff
@@ -1,9 +1,39 @@
...
+    "\\.afan\\.ai$",
+    "http://localhost:3000"
+  ],
+  "DEBUG": false,
+  "ENABLE_DEV_CLIENT_SET_API": false,
+  "ENABLE_EVENT_HANDLER": false,
+  "ENABLE_EXPRESS_RATE_LIMIT": true,
+  "ENABLE_GAS_FEE_WORKAROUND": true,
+  "ENABLE_JSON_RPC_TX_API": true,
+  "ENABLE_TX_SIG_VERIF_WORKAROUND": false,
```
platfowner, line 21: minor: sorting?

common/constants.js
```diff
@@ -24,135 +24,112 @@ const DevFlags = {
...
-    process.env.BLOCKCHAIN_CONFIGS_DIR : null;
-BlockchainConfigs.GENESIS_BLOCK_DIR =
-    path.resolve(__dirname, '..', process.env.BLOCKCHAIN_CONFIGS_DIR || BlockchainConfigs.BASE_BLOCKCHAIN_CONFIGS_DIR);
+const BlockchainConsts = {};
+// *** Genesis ***
 const BlockchainParams = getBlockchainConfig('blockchain_params.json');
+BlockchainConsts.CHAIN_ID = BlockchainParams.blockchain.chain_id;
+BlockchainConsts.EPOCH_MS = BlockchainParams.consensus.epoch_ms;
+BlockchainConsts.GENESIS_ADDR = BlockchainParams.genesis.genesis_addr;
+BlockchainConsts.NETWORK_ID = BlockchainParams.network.network_id;
```
platfowner, line 66: Ah, if the 4 constants still need to be placed in a blockchain-params.json for easy value inheritance/overriding, would it be simpler just not to add them to BlockchainConsts?

liayoo: Oh do you mean just accessing the values like so? `BlockchainParams.blockchain.chain_id`

common/constants.js
```diff
@@ -24,135 +24,112 @@ const DevFlags = {
   enableTrafficMonitoring: true,
 };
 
+// ** Node configs, set for individual nodes by env vars **
+const NodeConfigs = {};
+NodeConfigs.BLOCKCHAIN_CONFIGS_DIR = process.env.BLOCKCHAIN_CONFIGS_DIR || 'blockchain-configs/base';
```
platfowner, line 29: To be included in the /get_config results, could 'blockchain-configs/base' be added as a value of NodeConfigs as well? e.g.
```
NodeConfigs.BASE_BLOCKCHAIN_CONFIGS_DIR = 'blockchain-configs/base';
NodeConfigs.BLOCKCHAIN_CONFIGS_DIR = process.env.BLOCKCHAIN_CONFIGS_DIR || null;
```
(If it looks weird to add it to NodeConfigs, to BlockchainConsts instead, i.e., `BlockchainConsts.BASE_BLOCKCHAIN_CONFIGS_DIR = 'blockchain-configs/base';`?)
Please let me know how it looks like.

common/constants.js
platfowner: Looks great, thanks! 👍

common/constants.js
platfowner: Ah, I mean just going back to getBlockchainParam() if we cannot avoid them to be added to the states? Let me know which one looks better to you.

common/constants.js
liayoo: Hmm.. I don't follow why we want the default value to be in the /get_config result?

common/constants.js
liayoo: Ah, I see! I actually like having them in the BlockchainConsts since that way, it's clearer that they're constant, and it's simpler to access than passing them as function arguments. I can exclude them from being in the state, if you think that would cause some confusion?

blockchain-configs/base/node_params.json
liayoo: Oops, thanks for catching that! Fixed!

common/constants.js
```diff
@@ -24,135 +24,112 @@ const DevFlags = {
...
+BlockchainConsts.VARIABLE_LABEL_PREFIX = '$';
+BlockchainConsts.STATE_INFO_PREFIX = '#';
+BlockchainConsts.KEYS_ROOT_DIR = path.resolve(NodeConfigs.BLOCKCHAIN_DATA_DIR, 'keys');
+// *** Error Codes ***
+BlockchainConsts.TX_NONCE_ERROR_CODE = 900;
+BlockchainConsts.TX_TIMESTAMP_ERROR_CODE = 901;
+// *** Token Denominations ***
+BlockchainConsts.MILLI_AIN = 10**-3; // 1,000 milliain = 1 ain
+BlockchainConsts.MICRO_AIN = 10**-6; // 1,000,000 microain = 1 ain
+// *** State Budgets ***
```
platfowner, line 116: could you move these values (lines 117 - 127) to blockchain-params.json?

common/constants.js
```diff
@@ -24,135 +24,112 @@ const DevFlags = {
...
-  'https://*.afan.ai',
-  'http://localhost:3000'
-];
+BlockchainConsts.SNAPSHOTS_INTERVAL_BLOCK_NUMBER = 1000; // How often the snapshot is generated.
+BlockchainConsts.MAX_NUM_SNAPSHOTS = 10; // Maximum number of snapshots to be kept.
+BlockchainConsts.HASH_DELIMITER = '#';
+BlockchainConsts.VARIABLE_LABEL_PREFIX = '$';
+BlockchainConsts.STATE_INFO_PREFIX = '#';
+BlockchainConsts.KEYS_ROOT_DIR = path.resolve(NodeConfigs.BLOCKCHAIN_DATA_DIR, 'keys');
+// *** Error Codes ***
```
platfowner, line 110: I seems these error code definitions are not used any more. Could you remove them for me?

common/constants.js
platfowner: Actually the original reason to introduce BlockchainConfigs was to expose config details via API. One typical example when it's useful I think is when some blockchain nodes show abnormalities (esp. when this affects the performance of the overall network) but they're not owned by us. If we can check their config details without inspecting their source code, I think it could help understand what's happening and how we can handle them?

BTW, string 'blockchain-configs/base' is used multiple times in the source code and basically it can be defined as a constant from SWE perspective?

common/constants.js
platfowner: Somehow blockchain params can be seen more immutable values than those in BlockchainConsts as the former can only be changed by the consensus admin while the latter by the blockchain node owners?

common/constants.js
```diff
@@ -24,135 +24,97 @@ const DevFlags = {
...
-  'https://*.ainetwork.ai',
-  'https://*.ainize.ai',
-  'https://*.afan.ai',
-  'http://localhost:3000'
-];
+BlockchainConsts.SNAPSHOTS_INTERVAL_BLOCK_NUMBER = 1000; // How often the snapshot is generated.
+BlockchainConsts.MAX_NUM_SNAPSHOTS = 10; // Maximum number of snapshots to be kept.
+BlockchainConsts.HASH_DELIMITER = '#';
+BlockchainConsts.VARIABLE_LABEL_PREFIX = '$';
+BlockchainConsts.STATE_INFO_PREFIX = '#';
```
platfowner, line 108: Could you move lines 106 - 108 to blockchain-params.json?

common/constants.js
```diff
@@ -24,135 +24,97 @@ const DevFlags = {
...
+BlockchainConsts.KEYS_ROOT_DIR = path.resolve(NodeConfigs.BLOCKCHAIN_DATA_DIR, 'keys');
+// *** Token Denominations ***
+BlockchainConsts.MILLI_AIN = 10**-3; // 1,000 milliain = 1 ain
```
platfowner, line 111: It looks like this is not used any more. Could you remove it for me?

common/constants.js
platfowner: (offline discussion) to "genesis" section of blockchain-params.json?

common/constants.js
```diff
@@ -24,135 +24,97 @@ const DevFlags = {
...
+BlockchainConsts.MICRO_AIN = 10**-6; // 1,000,000 microain = 1 ain
```
platfowner, line 112: (offline discussion) Give new name e.g. GAS_PRICE_UNIT and move to blockchain-params.json ("resource" section?)?

common/constants.js
```diff
@@ -24,135 +24,97 @@ const DevFlags = {
...
+// *** Traffic ***
+BlockchainConsts.TRAFFIC_DB_INTERVAL_MS = 60000;  // 1 min
+BlockchainConsts.TRAFFIC_DB_MAX_INTERVALS = 60;  // 1 hour
```
platfowner, line 115: (offline discussion) to node-params.json?

common/constants.js
```diff
@@ -24,135 +24,97 @@ const DevFlags = {
...
+// *** Functions ***
+BlockchainConsts.REST_FUNCTION_CALL_TIMEOUT_MS = 10000; // 10 sec
```
platfowner, line 117: (offline discussion) to node-params.json?

common/constants.js
platfowner: (offline discussion) 'blockchain-configs/base' to BlockchainConsts?

common/constants.js
liayoo: done!

common/constants.js
liayoo: done!

common/constants.js
liayoo: done!

common/constants.js
liayoo: done!

common/constants.js
liayoo: done!

common/constants.js
liayoo: done!

common/constants.js
liayoo: done!

common/constants.js
liayoo: done!

liayoo:
Woah... this PR is getting out of hand.. 🤯 Thank you so much for all the support, review, & patience! Addressed all comments. PTAL~

p2p/index.js
```diff
@@ -127,10 +124,11 @@ class P2pClient {
 
   getConfig() {
     return {
-      blockchainParams: BlockchainParams,
+      blockchainParams: this.server.node.getAllBlockchainParamsFromState(),
       env: process.env,
       devFlags: DevFlags,
-      blockchainConfigs: BlockchainConfigs,
+      BlockchainConsts,
```
platfowner, line 130: minor: what about `blockchainConsts: BlockchainConsts,`?

platfowner (approved):
Now the code looks clean and clear. Thanks for the thorough and patient reorganization!

blockchain-configs/1-node/blockchain_params.json
```diff
@@ -0,0 +1,87 @@
+{
+  "_description": "Blockchain params for a 1-node network. Mostly used for unit tests.",
```
platfowner, line 2: just a qq: can the property order in this file affect the state proof hash?

liayoo: Thanks for asking! If the params within a category (e.g. consensus/resource/genesis) are reordered, yes, it would change the proof hash.

platfowner: Right, thanks for the answer!

liayoo:
Thanks for the review! Will merge once the benchmark is updated cc: @cshcomcom

p2p/index.js
liayoo: done, thanks!

PR #757, opened 2021-12-07
Feature/csh/event handler convert logic

cshcomcom:
## Description
- Add convert functions
- Fix bugs

### DIFF
--- a/event-handler/index.js
+++ b/event-handler/index.js
@@ -51,7 +51,7 @@ class EventHandler {
     if (!clientFilterId) {
       throw Error(`Can't get client filter ID from node filter ID (nodeFilterId: ${nodeFilterId})`);
     }
-    return `${clientFilterId}`;
+    return clientFilterId;
   }
 
   clientFilterIdToNodeFilterId(clientFilterId, channelId) {
--- a/event-handler/index.js
+++ b/event-handler/index.js
@@ -46,8 +46,20 @@ class EventHandler {
     // TODO(cshcomcom): Implement
   }
 
+  convertNodeFilterIdClientFilterIdTo(nodeFilterId) {
+    const [channelId, clientFilterId] = nodeFilterId.split(':');
+    if (!clientFilterId) {
+      throw Error(`Can't get client filter ID from node filter ID (nodeFilterId: ${nodeFilterId})`);
+    }
+    return `${clientFilterId}`;
+  }
+
+  convertClientFilterIdToNodeFilterId(clientFilterId, channelId) {
+    return `${channelId}:${clientFilterId}`;
+  }
+
   createAndRegisterEventFilter(clientFilterId, channelId, eventType, config) {
-    const eventFilterId = `${channelId}:${clientFilterId}`;
+    const eventFilterId = this.convertClientFilterIdToNodeFilterId(clientFilterId, channelId);
     if (this.eventFilters[eventFilterId]) {
       throw Error(`Event filter ID ${eventFilterId} is already in use`);
     }
--- a/event-handler/index.js
+++ b/event-handler/index.js
@@ -35,9 +35,8 @@ class EventHandler {
       const eventFilter = this.eventFilters[eventFilterId];
       const eventFilterBlockNumber = _.get(eventFilter, 'config.block_number', -1);
       if (eventFilterBlockNumber === -1) {
-        continue;
-      }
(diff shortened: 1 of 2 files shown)


event-handler/index.js
```diff
@@ -47,8 +46,20 @@ class EventHandler {
...
 
+  convertNodeFilterIdClientFilterIdTo(nodeFilterId) {
+    const [channelId, clientFilterId] = nodeFilterId.split(':');
+    if (!clientFilterId) {
+      throw Error(`Can't get client filter ID from node filter ID (nodeFilterId: ${nodeFilterId})`);
+    }
+    return `${clientFilterId}`;
+  }
+
+  convertClientFilterIdToNodeFilterId(clientFilterId, channelId) {
```
liayoo, line 57: ditto (`clientFilterIdToNodeFilterId()`)

event-handler/index.js
```diff
@@ -35,9 +35,8 @@ class EventHandler {
       const eventFilter = this.eventFilters[eventFilterId];
       const eventFilterBlockNumber = _.get(eventFilter, 'config.block_number', -1);
```
platfowner, line 36: as '-1' is a block number actually used in the blockchain (as an initial value), what about using null instead?

event-handler/index.js
platfowner, line 57: what does 'node filter' mean? it looks like just a concatenation of channel id and client filter id?
is the purpose of this concatenation to make the id unique in the blockchain node?

cshcomcom: Yes! It just exists to avoid collisions.

event-handler/index.js
cshcomcom: ah. Thanks I changed it to null! Done

event-handler/index.js
```diff
@@ -47,8 +46,20 @@ class EventHandler {
     // TODO(cshcomcom): Implement
   }
 
+  convertNodeFilterIdClientFilterIdTo(nodeFilterId) {
```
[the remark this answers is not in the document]
cshcomcom: Thanks! Looks better. Done

cshcomcom:
All comments are addressed! PTAL @liayoo @platfowner

event-handler/index.js
platfowner: I see. If that's the meaning, what about 'globalFilterId' instead of 'nodeFilterId'?

event-handler/index.js
```diff
@@ -47,8 +46,20 @@ class EventHandler {
     // TODO(cshcomcom): Implement
   }
 
+  nodeFilterIdToClientFilterId(nodeFilterId) {
+    const [channelId, clientFilterId] = nodeFilterId.split(':');
+    if (!clientFilterId) {
+      throw Error(`Can't get client filter ID from node filter ID (nodeFilterId: ${nodeFilterId})`);
+    }
+    return `${clientFilterId}`;
```
platfowner, line 54: `return clientFilterId;`?

event-handler/index.js
```diff
@@ -47,8 +46,20 @@ class EventHandler {
...
+  }
+
+  clientFilterIdToNodeFilterId(clientFilterId, channelId) {
```
platfowner, line 57: what about just 'getGlobalFilterId()' as it's not a one-to-one mapping?
And placing channelId param first as it comes first in the return value?

event-handler/index.js
```diff
@@ -47,8 +46,20 @@ class EventHandler {
     // TODO(cshcomcom): Implement
   }
 
+  nodeFilterIdToClientFilterId(nodeFilterId) {
```
platfowner, line 49: minor: 'getClientFilterIdFromGlobalFilterId()' seems to be an option for its name.

event-handler/index.js
cshcomcom: 'Global' word looks nice to me.

cshcomcom: I changed function name and params order! Done

event-handler/index.js
cshcomcom: Done

event-handler/index.js
cshcomcom: Thanks! Done

platfowner (approved):
LGTM
Thanks!

cshcomcom:
Thanks for reviews! Now merging! I'll keep an eye for comments after merged 👀

PR #760, opened 2021-12-08
Fix flaky test cases with created_at tx field value

platfowner:
Change summary:
- Fix flaky test cases with created_at tx field value


liayoo (approved):
LGTM, thanks!

Issue #761, opened 2021-12-08
[BUG] Grafana too slow in sandbox


platfowner:
We need to optimize the monitoring data size.

platfowner:
Fixed.

PR #763, opened 2021-12-08
Move to node-params.json and customize traffic params for sandbox environment

platfowner:
Change summary:
- Move to node-params.json and customize traffic params for sandbox environment (leave only '1m' stats period for sandbox)
- Move p2p params to node-params.json

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/761


p2p/index.js
```diff
@@ -31,15 +31,6 @@ const {
 } = require('../common/network-util');
 
 // NOTE(minsulee2): consider updating TRACKER_UPDATE_INTERVAL_MS to 2.5 minutes.
```
liayoo, line 33: could you remove this note or move it to L180? cc: @minsulee2

liayoo (approved):
LGTM with a comment, thanks!

p2p/index.js
minsulee2: Can you remove this? I don't think this NOTE is valid anymore !

minsulee2 (commented):
LGTM!

minsulee2 (approved):

p2p/index.js
platfowner: It was an issue for the traffic volume from the nodes to the tracker, and I think it could be still an issue even with json rpc API. Updated it accordingly. PTAL

platfowner:
Let me merge it now, thanks!

Issue #762, opened 2021-12-08
[BUG] Integration test run is too slow


platfowner:

platfowner:
Fixed, thanks!

PR #767, opened 2021-12-08
Set epoch_ms to 1000 for test blockchain configs

platfowner:
Change summary:
- Set epoch_ms to 1000 for test blockchain configs

Result:
- The run time of integration/node.test.js reduced from 19 min to 8 min in my experiment.

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/762


liayoo:
Nice! Could you check in the updated genesis_block.json.gz files as well?

platfowner:
> Nice! Could you check in the updated genesis_block.json.gz files as well?

Right. Just done, thanks!

liayoo (approved):
LGTM, thanks!

Issue #745, opened 2021-11-30
GCP logging error


liayoo:
In some of the testnet-sandbox nodes, the following error logs have been observed (in error_logs.txt)

```
(node:11314) UnhandledPromiseRejectionWarning: Error: Total timeout of API google.logging.v2.LoggingServiceV2 exceeded 60000 milliseconds before any response was received.
    at repeat (/home/ain-blockchain/node_modules/google-gax/build/src/normalCalls/retries.js:66:31)
    at Timeout._onTimeout (/home/ain-blockchain/node_modules/google-gax/build/src/normalCalls/retries.js:101:25)
    at listOnTimeout (internal/timers.js:554:17)
    at processTimers (internal/timers.js:497:7)
(node:11314) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:11314) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

platfowner:
error_logs.txt of summer node-1 in R48:
```
SyntaxError: Unexpected token @ in JSON at position 0
    at JSON.parse (<anonymous>)
    at createStrictSyntaxError (/home/ain-blockchain/node_modules/body-parser/lib/types/json.js:158:10)
    at parse (/home/ain-blockchain/node_modules/body-parser/lib/types/json.js:83:15)
    at /home/ain-blockchain/node_modules/body-parser/lib/read.js:121:18
    at invokeCallback (/home/ain-blockchain/node_modules/raw-body/index.js:224:16)
    at done (/home/ain-blockchain/node_modules/raw-body/index.js:213:7)
    at IncomingMessage.onEnd (/home/ain-blockchain/node_modules/raw-body/index.js:273:7)
    at IncomingMessage.emit (events.js:314:20)
    at endReadableNT (_stream_readable.js:1241:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)
SyntaxError: Unexpected token u in JSON at position 1
    at JSON.parse (<anonymous>)
[...]

PR #772, opened 2021-12-10
Do not call winstonLogger.debug() if DEBUG=false.

liayoo:
It seems the winston library has some issues with garbage collecting when a log function with a level lower than the current level is called. Fixed this by not calling winstonLogger.debug() if DEBUG env var is not true.

Without the changes:
Screen Shot 2021-12-09 at 2 14 11 PM

With the changes:
Screen Shot 2021-12-09 at 6 08 20 PM

\+ It also seems that bunyan, on the other hand, doesn't have this issue. I'll conduct some quick research on it & share it in the next meeting. We can compare & discuss which library to use!

Related: https://github.com/ainblockchain/ain-blockchain/issues/745

### DIFF
--- a/logger/index.js
+++ b/logger/index.js
@@ -1,6 +1,7 @@
 /* eslint new-cap: "off" */
 const winston = require('winston');
 const { getWinstonLevels, getWinstonColors, getWinstonTransports } = require('./winston-util');
+const { NodeConfigs } = require('../common/constants');
 
 const winstonLogger = new winston.createLogger({
   levels: getWinstonLevels(),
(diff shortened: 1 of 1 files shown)


logger/index.js
```diff
@@ -30,7 +31,7 @@ class Logger {
   }
 
   debug(text) {
-    if (!isFinished) {
+    if (!isFinished && NodeConfigs.DEBUG) {
```
minsulee2, line 34: was it logging even if DEBUG is off?

minsulee2 (approved):
LGTM!

logger/index.js
liayoo: No, but I suspect that when it's **not** logged, the gc doesn't work properly..

minsulee2: I see. Thanks for the explain!

platfowner (approved):
LGTM
Thanks for handling this!

platfowner:
Wow, thanks for sharing!

liayoo:

PR #774, opened 2021-12-11
Fix check-in, check-out bugs

liayoo:
- Add sender_proof to check-in requests to make sure that the requester actually owns the 'sender' ethereum key
  - sender_proof is a message containing the ref, amount, sender, timestamp, and nonce that's signed by the sender eth key
- Check that check-in, check-out ids don't exist in the history
  - Note that this alone cannot enforce unique keys though, since history is garbage collected. I'll update the bridge server to close a request with status = false if there's already a record in history with the same id
- Change check-in, check-out request status type to boolean


db/functions.js
```diff
@@ -964,6 +966,24 @@ class Functions {
...
+  validateCheckinSenderProof(ref, amount, sender, senderProof, chainId, tx) {
+    try {
+      const body = {
+        ref,
+        amount,
+        sender,
+        timestamp: _.get(tx, 'tx_body.timestamp'),
+        nonce: _.get(tx, 'tx_body.nonce'),
+      };
+      if (!ainUtil.ecVerifySig(stringify(body), senderProof, sender, chainId) === true) {
```
platfowner, line 978: question: is this correct that 'sender' is a ethereum address but 'sendProof' is generated using ain-util (not ethereum signing library)?

liayoo: You're right! I've updated to use web3-eth-accounts's sign & recover functions. Thanks!

db/functions.js
```diff
@@ -991,7 +1012,7 @@ class Functions {
     const tokenId = context.params.token_id;
     // NOTE(liayoo): `sender` is the address on `networkName` that will send `tokenId` tokens to the pool.
     // For example, with the Eth token bridge, it will be an Ethereum address that will send ETH to the pool.
-    const { amount, sender } = value;
+    const { amount, sender, sender_proof: senderProof } = value;
```
platfowner, line 1015: could you add a note for the 'senderProof' as well?

tools/checkin/sendOpenCheckinTx.js
```diff
@@ -1,21 +1,34 @@
...
+  const ref = `/checkin/requests/ETH/3/0xB16c0C80a81f73204d454426fC413CAe455525A7/${fromAddr}/${checkinId}`;
+  const timestamp = Date.now();
+  const body = {
+    ref,
+    amount: tokenAmount,
+    sender: config.senderAddr,
+    timestamp,
+    nonce: -1,
+  };
+  const senderProof = ainUtil.ecSignMessage(stringify(body), Buffer.from(config.senderPrivateKey, 'hex'), 3);
```
platfowner, line 19: need to be updated as well?

platfowner: and could you add a note for 'senderProof' here as well?

db/functions.js
liayoo: done!

tools/checkin/sendOpenCheckinTx.js
liayoo: right, thanks! done!

platfowner (approved):
LGTM
Thanks for handling the issues.

PR #775, opened 2021-12-11
Deprecate GenesisToken and GenesisSharding

liayoo:
Deprecated GenesisToken and GenesisSharding & updated to use on-chain values.


db/index.js
```diff
@@ -555,6 +554,13 @@ class DB {
     if (blockNumber <= 0) {
       return BlockchainParams[category][name];
     }
+    if (category === 'sharding') {
```
platfowner, line 557: can't we just move the sharding params and the token params under /blockchain_params?

liayoo: Apologies for the delayed updates! I've migrated `/sharding/config` to `/blockchain_params/sharding` and `/token` to `/blockchain_params/token`. 
I decided to keep the `/sharding/shard` path as it is so that `/blockchain_params` doesn't become too crowded & it's easier to manage shards. Let me know if you feel differently.

liayoo:
Addressed all comments, PTAL @platfowner~

platfowner (approved):
LGTM
Thanks for the nice refactoring!

liayoo:

PR #777, opened 2021-12-13
Use bunyan as the default logger, add error catching

liayoo:
- Updated to use bunyan as the default logger
- Added enableWinstonLogger to DevFlags (default = false)
- Added try-catch for logging
- Increased node version in setup script to 16
- Updated some deploy & local scripts


logger/winston-util.js
```diff
@@ -94,14 +94,17 @@ const getWinstonDailyErrorFileTransport = () => {
...
-      ];
+  if (NodeConfigs.LIGHTWEIGHT) {
+    return [getWinstonDailyErrorFileTransport()];
+  }
+  const transports = [
+    getWinstonDailyCombinedFileTransport(),
+    getWinstonDailyErrorFileTransport(),
+  ];
+  if (NodeConfigs.CONSOLE_LOG) {
+    transports.push(getWinstonConsoleTransport());
```
platfowner, line 105: can 'silent' line (line 56) be removed now?

liayoo: right, done. thanks!

platfowner (approved):
LGTM
Thanks!

PR #778, opened 2021-12-13
Consensus tx clean-up in tx pool & fix bugs

liayoo:
- Added consensus tx clean-up for discarded blocks
- Fixed stakeTx not setting bug
- Updated to return fast if block is already notarized in tryUpdateNotarized()
- Added a nonce check to isValidConsensusTx()
- Reduced MAX_CONSENSUS_LOGS_IN_STATES to 100 and renamed as MAX_INVALID_BLOCKS_ON_MEM
- Updated to try to get block from the on-mem chain by hash in getSnapDb().


block-pool/index.js
```diff
@@ -468,6 +471,12 @@ class BlockPool {
   }
 
   cleanUpForBlockHash(blockHash) {
+    if (_get(this.hashToBlockInfo[blockHash], 'block')) {
```
platfowner, line 474: minor: what about adding variable definition e.g. `const block = _get(this.hashToBlockInfo[blockHash], 'block');`?

block-pool/index.js
```diff
@@ -468,6 +471,12 @@ class BlockPool {
...
+      this.node.tp.cleanUpConsensusTxsForBlock(this.hashToBlockInfo[blockHash].block);
+    }
+    if (_get(this.hashToInvalidBlockInfo[blockHash], 'block')) {
```
platfowner, line 477: minor: ditto?

tx-pool/index.js
```diff
@@ -410,6 +410,51 @@ class TransactionPool {
...
+    }
+  }
+
+  cleanUpConsensusTxsForBlock(block) {
+    const consensusTxs = new Set(block.last_votes.map((tx) => tx.hash));
+    this.addEvidenceTxsToTxHashSet(consensusTxs, block.evidence);
+    this.updateTxPoolWithTxHashSet(consensusTxs, {}, {});
+  }
+
   cleanUpForNewBlock(block) {
```
platfowner, line 458: could you update the test case in tx-pool.test.js by adding some evidences to blocks?

platfowner (approved):
LGTM with some comments.
Thanks!

tx-pool/index.js
liayoo: Added a test case for cleanUpConsensusTxsForBlock() with some evidence votes. Thanks!

block-pool/index.js
liayoo: done, thanks!

block-pool/index.js
liayoo: ditto

liayoo:
Thanks for the review~ Now merging!

PR #785, opened 2021-12-16
Fix consensus message clean-up from tx pool

liayoo:
- Removed block/invalid block checking before consensus txs clean-ups
- Included some missing consensus txs (proposal, valid block votes) to the clean-ups
- Updated to call addToNumberToBlockSet() in addSeenVote() to make sure we clean up the votes
- Added explicit Number type casting

...I haven't seen consensus txs being collected in the tx pool in my local setting with these changes, but let me test with 100 nodes in the sandbox.


liayoo:
Updates: I was able to confirm that old consensus txs no longer accumulate in the tx pool. Tested with 30 active validators, 70 passive validators, at 5 tps for ~12 hrs.

platfowner (approved):
LGTM
Thanks!

liayoo:

PR #787, opened 2021-12-17
Deprecate ACCOUNT_INDEX

liayoo:
- Deprecated ACCOUNT_INDEX node param
- Added private_key account injection option
- Introduced UNSAFE_PRIVATE_KEY node param
- Added ain_injectAccountFromPrivateKey json-rpc api
- Updated scripts


node/index.js
```diff
@@ -131,6 +167,31 @@ class BlockchainNode {
     }
   }
 
+  verifyNodeAccountSignature(message, signature) {
```
platfowner, line 170: used no where?

liayoo: Ah.. I was working on 2 features at the same time & some changes got mixed up :') This should be included in #794.. Will fix soon!

node/index.js
```diff
@@ -73,14 +73,30 @@ class BlockchainNode {
...
+      case 'keystore':
+        if (!NodeConfigs.KEYSTORE_FILE_PATH) {
+          throw Error(
+              `[${LOG_HEADER}] Must specify KEYSTORE_FILE_PATH as a process env or in node_params.json`);
+        }
+        break;
+      case 'mnemonic':
+        break;
+      case 'private_key':
+        // NOTE(liayoo): Could either specify UNSAFE_PRIVATE_KEY or inject an account using APIs.
```
platfowner, line 86: it looks like there are two different use cases: 1) with private key injection and 2) with NodeConfigs.UNSAFE_PRIVATE_KEY value.
what about separating them e.g. by using ACCOUNT_INJECTION_OPTION = null (default value) for case 2) while using 'private_key' for the other case?

deploy_blockchain_genesis_gcp.sh
```diff
@@ -106,47 +112,49 @@ then
     [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1 # handle exits from shell or function but don't exit interactive shell
 fi
 
+# Read node ip addresses
+IFS=$'\n' read -d '' -r -a IP_ADDR_LIST < ./testnet_ip_addresses/$SEASON.txt
```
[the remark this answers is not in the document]
liayoo: Thanks for asking! Now that the deploy scripts need to always use an account injection method (whether it be private key/mnemonic/keystore), we should always have the IP address files.

node/index.js
liayoo: Right, that'd be less ambiguous. Updated as you suggested, PTAL~

deploy_blockchain_sandbox_gcp.sh
```diff
@@ -79,6 +86,16 @@ then
     [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1 # handle exits from shell or function but don't exit interactive shell
 fi
 
+function inject_account() {
+    local node_index="$1"
+    local node_ip_addr=${IP_ADDR_LIST[${node_index}]}
+    local GENESIS_ACCOUNTS_PATH="blockchain-configs/base/genesis_accounts.json"
+    printf "\n* >> Injecting an account for node $node_index ********************\n\n"
+    printf "node_ip_addr='$node_ip_addr'\n"
+    PRIVATE_KEY=$(cat $GENESIS_ACCOUNTS_PATH | jq -r '.others.[$node_index].private_key')
```
platfowner, line 95: 👍

deploy_blockchain_genesis_gcp.sh
```diff
@@ -106,47 +117,49 @@ then
...
         } | node inject_account_gcp.js $node_ip_addr $ACCOUNT_INJECTION_OPTION
+    else
+        printf "\n* >> Injecting an account for node $node_index ********************\n\n"
+        printf "node_ip_addr='$node_ip_addr'\n"
+        local GENESIS_ACCOUNTS_PATH="blockchain-configs/base/genesis_accounts.json"
+        if [[ "$SEASON" = "spring" ]] || [[ "$SEASON" = "summer" ]]; then
+            GENESIS_ACCOUNTS_PATH="blockchain-configs/testnet-prod/genesis_accounts.json"
+        fi
+        PRIVATE_KEY=$(cat $GENESIS_ACCOUNTS_PATH | jq -r '.others['$node_index'].private_key')
+        printf "PRIVATE_KEY=$PRIVATE_KEY\n"
```
platfowner, line 161: could you skip printing out private keys just in case (as you know, this shouldn't be the case of testnet prod or mainnet though)?

platfowner (approved):
Thanks for nice refactoring!
And now we're supporting four account injection methods~

deploy_blockchain_genesis_gcp.sh
liayoo: Right! Thanks for pointing it out. Removed.

liayoo:

Issue #779, opened 2021-12-14
Need to handle subtree rules and owners properly


platfowner:

platfowner:
Fixed, thanks!

PR #788, opened 2021-12-17
Handle subtree owners properly

platfowner:
Change summary:
- Do NOT allow non-empty subtree owners in set_rule, set_function, and set_owner
- Return owner permission-related code and error messages
- Include subtree config path list in the error message

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/779


db/index.js
```diff
@@ -2289,8 +2412,16 @@ class DB {
   }
 
   checkPermission(config, auth, permission) {
-    const permissions = this.getOwnerPermissions(config, auth);
-    return !!(permissions && permissions[permission] === true);
+    const permissionsObj = this.getOwnerPermissions(config, auth);
```
liayoo, line 2415: minor) how about using only singular/only plural?

liayoo (approved):

db/index.js
platfowner: Just applied some renaming:
- permissionsObj -> permissionObj
- permissionsString -> permissionString

Please let me know if your comment was not interpreted correctly.

PR #791, opened 2021-12-19
Add propose/vote p2p message latency to traffic stats

platfowner:
Change summary:
- Add propose/vote p2p message latency to traffic stats
- Remove traffic stats manager from BlockchainNode constructor

### DIFF
--- a/common/constants.js
+++ b/common/constants.js
@@ -569,6 +569,8 @@ const TrafficEventTypes = {
   PROPOSE_BEFORE_BLOCK: 'propose_before_block',
   VOTE_BEFORE_BLOCK: 'vote_before_block',
   VOTE_AFTER_PROPOSE: 'vote_after_propose',
+  PROPOSE_P2P_MESSAGE: 'propose_p2p_message',
+  VOTE_P2P_MESSAGE: 'vote_p2p_message',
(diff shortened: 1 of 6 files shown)


consensus/index.js
```diff
@@ -220,6 +222,20 @@ class Consensus {
     return true;
   }
 
+  addTrafficEventsForProposalTx(proposalTx) {
```
liayoo, line 225: could you move these to the consensus-util.js?

common/constants.js
```diff
@@ -569,6 +569,8 @@ const TrafficEventTypes = {
   PROPOSE_BEFORE_BLOCK: 'propose_before_block',
```
liayoo, line 569: minor) `// Latency` -> `// Consensus messages`?

liayoo (approved):

minsulee2 (approved):
LGTM!

common/constants.js
platfowner: Just changed to 'Consensus & block producing'. PTAL~

consensus/index.js
platfowner: Done, thanks!

PR #793, opened 2021-12-19
Allow writing with subtree owners in partial set cases (with isPartialSet = true)

platfowner:
Change summary:
- Allow writing rules / functions / owners with subtree owners in partial set cases (with isPartialSet = true)
- Do not return empty error_messages

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/779

### DIFF
--- a/common/common-util.js
+++ b/common/common-util.js
@@ -255,8 +255,8 @@ class CommonUtil {
     if (args.is_global !== undefined) {
       options.isGlobal = CommonUtil.toBool(args.is_global);
     }
-    if (args.is_partial_set !== undefined) {
-      options.isPartialSet = CommonUtil.toBool(args.is_partial_set);
+    if (args.is_merge !== undefined) {
+      options.isMerge = CommonUtil.toBool(args.is_merge);
     }
     return options;
   }
--- a/common/common-util.js
+++ b/common/common-util.js
@@ -247,6 +247,9 @@ class CommonUtil {
     if (args.include_proof !== undefined) {
       options.includeProof = CommonUtil.toBool(args.include_proof);
     }
+    if (args.is_partial_set !== undefined) {
+      options.isPartialSet = CommonUtil.toBool(args.is_partial_set);
+    }
     return options;
   }
 
--- a/common/common-util.js
+++ b/common/common-util.js
@@ -247,9 +247,6 @@ class CommonUtil {
     if (args.include_proof !== undefined) {
       options.includeProof = CommonUtil.toBool(args.include_proof);
     }
-    if (args.is_partial_set !== undefined) {
-      options.isPartialSet = CommonUtil.toBool(args.is_partial_set);
-    }
     return options;
   }
 
@@ -257,6 +254,9 @@ class CommonUtil {
     const options = {};
     if (args.is_global !== undefined) {
       options.isGlobal = CommonUtil.toBool(args.is_global);
+    }
+    if (args.is_partial_set !== undefined) {
+      options.isPartialSet = CommonUtil.toBool(args.is_partial_set);
     }
     return options;
   }
--- a/db/state-util.js
+++ b/db/state-util.js
@@ -608,6 +608,16 @@ function hasConfigLabelOnly(stateTreeObj, configLabel) {
 }
 
 /**
+ * Returns a config at the given path if available, otherwise null.
+ */
+function getConfigAtPath(stateTreeObj, configPath) {
+  if (!CommonUtil.isDict(stateTreeObj)) {
+    return null;
+  }
+  return CommonUtil.getJsObject(stateTreeObj, configPath);
+}
+
+/**
  * Returns a new rule tree created by applying the rule change to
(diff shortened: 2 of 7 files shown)


liayoo:
Could you help me understand the use case of partial sets? Like if we're going to allow users to specify is_partial_set, why can't we just allow it always? (without specifying is_partial_set option)

platfowner:
> Could you help me understand the use case of partial sets? Like if we're going to allow users to specify is_partial_set, why can't we just allow it always? (without specifying is_partial_set option)

Thanks for asking. is_partial_set options is used for only eval APIs (ain_evalOwner or /eval_owner) APIs.
For set operations (SET_RULE, SET_FUNCTION, SET_OWNER), it's auto-detected by apply functions (applyRuleChange(), applyFunctionChange(), applyOwnerChange()).
For now this kind of auto-detection is impossible for the eval APIs as users are not allowed to pass the actual rule / function / owner config.
Let me check if we can allow such values as a parameter after this PR.

common/common-util.js
```diff
@@ -255,6 +255,9 @@ class CommonUtil {
     if (args.is_global !== undefined) {
       options.isGlobal = CommonUtil.toBool(args.is_global);
     }
+    if (args.is_partial_set !== undefined) {
```
liayoo, line 258: as discussed offline, how about `is_merge`?

liayoo:
> > Could you help me understand the use case of partial sets? Like if we're going to allow users to specify is_partial_set, why can't we just allow it always? (without specifying is_partial_set option)
> 
> Thanks for asking. is_partial_set options is used for only eval APIs (ain_evalOwner or /eval_owner) APIs. For set operations (SET_RULE, SET_FUNCTION, SET_OWNER), it's auto-detected by apply functions (applyRuleChange(), applyFunctionChange(), applyOwnerChange()). For now this kind of auto-detection is impossible for the eval APIs as users are not allowed to pass the actual rule / function / owner config. Let me check if we can allow such values as a parameter after this PR. (Just added a todo)

I see! Thanks for the detailed explanation. A to-do sounds good for now~

common/common-util.js
platfowner: Just updated, thanks!

platfowner:

db/state-util.js
```diff
@@ -607,34 +607,53 @@ function hasConfigLabelOnly(stateTreeObj, configLabel) {
...
  * the current rule tree.
  * 
  * @param {Object} curRuleTree current rule tree (to be modified by this rule)
  * @param {Object} ruleChange rule change
  */
-// NOTE(platfowner): Partial set is applied only when the current rule tree has
+// NOTE(platfowner): Config merge is applied only when the current rule tree has
 // .rule property and the rule change has .rule property as the only property.
 function applyRuleChange(curRuleTree, ruleChange) {
+  // 1. Config overwriting case:
```
liayoo, line 630: thanks for the comments! 👍

db/state-util.js
```diff
@@ -607,34 +607,53 @@ function hasConfigLabelOnly(stateTreeObj, configLabel) {
...
-    return CommonUtil.isDict(ruleChange) ?
+    const newRuleConfig = CommonUtil.isDict(ruleChange) ?
         JSON.parse(JSON.stringify(ruleChange)) : ruleChange;
+    return {
+      isMerge: false,
+      ruleConfig: newRuleConfig,
+    };
   }
-  const ruleChangeMap = CommonUtil.getJsObject(ruleChange, [PredefinedDbPaths.DOT_RULE]);
+  // 2. Config merge case:
```
liayoo, line 640: minor) how about no changes for this case & config merge case for the below?

db/state-util.js
```diff
@@ -607,34 +607,53 @@ function hasConfigLabelOnly(stateTreeObj, configLabel) {
...
  * 
  * @param {Object} curRuleTree current rule tree (to be modified by this rule)
  * @param {Object} ruleChange rule change
  */
-// NOTE(platfowner): Partial set is applied only when the current rule tree has
+// NOTE(platfowner): Config merge is applied only when the current rule tree has
 // .rule property and the rule change has .rule property as the only property.
 function applyRuleChange(curRuleTree, ruleChange) {
+  // 1. Config overwriting case:
   if (!hasConfigLabel(curRuleTree, PredefinedDbPaths.DOT_RULE) ||
```
liayoo, line 631: I think this condition is covered with !hasConfigLabelOnly() check?

platfowner: Ah, this is checked for 'curRuleTree' and the other one is for 'ruleChange'.

db/state-util.js
platfowner: just done, thanks!

db/state-util.js
liayoo: Right, missed that. Thanks!

liayoo (approved):
LGTM, thanks!

platfowner:
Thanks for the fast review, now merging..

PR #796, opened 2021-12-20
Enforce check-in, check-out ids to be same as the tx timestamp

liayoo:
To make sure they're unique, only allow the check-in/check-out ids that are the same as the tx timestamp.


db/rule-util.js
```diff
@@ -272,14 +273,15 @@ class RuleUtil {
     if (!_.isEqual(request, newData.request)) {
       return false;
     }
-    return this.isDict(newData.response) && this.isValidHash(newData.response.tx_hash) &&
-        this.isBool(newData.response.status);
+    return this.isDict(newData.response) && this.isBool(newData.response.status) &&
+        (newData.response.tx_hash === undefined || this.isValidHash(newData.response.tx_hash));
```
platfowner, line 277: minor: would it be helpful for the readers if some comments about the undefined case is available?

db/rule-util.js
```diff
@@ -242,11 +242,12 @@ class RuleUtil {
...
   }
 
-  validateCheckoutRequestData(networkName, chainId, tokenId, userAddr, checkoutId, newData, getValue) {
+  validateCheckoutRequestData(networkName, chainId, tokenId, userAddr, checkoutId, newData, currentTime, getValue) {
     const { PredefinedDbPaths } = require('../common/constants');
     const PathUtil = require('../common/path-util');
     if (!this.isDict(newData) || !this.isNumber(newData.amount) || newData.amount <= 0 ||
-        !this.isString(newData.recipient) || !this.isNumber(newData.fee_rate)) {
+        !this.isString(newData.recipient) || !this.isNumber(newData.fee_rate) ||
+        Number(checkoutId) !== currentTime) {
```
platfowner, line 250: would it be helpful if some comments are available about the restriction on the checkoutId?

db/rule-util.js
```diff
@@ -272,14 +273,15 @@ class RuleUtil {
...
+        (newData.response.tx_hash === undefined || this.isValidHash(newData.response.tx_hash));
   }
 
-  validateCheckinRequestData(networkName, chainId, tokenId, userAddr, checkinId, newData, getValue) {
+  validateCheckinRequestData(networkName, chainId, tokenId, userAddr, checkinId, newData, currentTime, getValue) {
     const PathUtil = require('../common/path-util');
     if (!this.isDict(newData) || !this.isNumber(newData.amount) || newData.amount <= 0 ||
-        !this.isString(newData.sender) || !this.isString(newData.sender_proof)) {
+        !this.isString(newData.sender) || !this.isString(newData.sender_proof) ||
+        Number(checkinId) !== currentTime) {
```
platfowner, line 284: ditto

platfowner (approved):
LGTM with some comments.
Thanks!

db/rule-util.js
liayoo: Right, done!

db/rule-util.js
liayoo: Done!

db/rule-util.js
liayoo: ditto

Issue #811, opened 2021-12-23
[BUG] Invalid IP from addToDevClientApiIpWhitelist.js


platfowner:
When I tried with my macbook's ip address, it returned some error messages:
```
node addToDevClientApiIpWhitelist.js http://<node 0 ip address>8:8080 0 private_key 192.168.35.158
Enter private key: 
Result: { code: 30501, message: 'Invalid IP: 192.168.35.158' }
```

platfowner:
I see it's fixed, thanks!

PR #812, opened 2021-12-23
Fix ip validity checking

liayoo:
- Fixed a bug in IP checking
- Added a wildcard case to ain_addToDevClientApiIpWhitelist

Fixes https://github.com/ainblockchain/ain-blockchain/issues/811

### DIFF
--- a/json_rpc/api-access.js
+++ b/json_rpc/api-access.js
@@ -44,7 +44,7 @@ module.exports = function getApiAccessApis(node) {
         }));
         return;
       }
-      if (net.isIPv4(args.message.ip)) {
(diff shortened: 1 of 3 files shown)


platfowner:
Could you add some empty value check to addToDevClientApiIpWhitelist.js?
It seems that undefined is currently interpreted as 'null' value.

liayoo:
> Thanks for handling this issue. Could you add some empty value check to addToDevClientApiIpWhitelist.js? It seems that undefined is currently interpreted as 'null' value.

That'd be nice! Added the checks. PTAL~

platfowner (approved):
LGTM
Thanks the quick fix!

PR #819, opened 2021-12-24
Update p2p state

minsulee2:
- Update p2p state correctly
- Make starting node actively find other peers
- Remove PEER_CANDIDATE_JSON_RPC_URL=''


p2p/util.js
```diff
@@ -43,10 +43,16 @@ function removeSocketConnectionIfExists(connectionObj, address) {
 
 function closeSocketSafe(connections, socket) {
   const address = getAddressFromSocket(connections, socket);
-  removeSocketConnectionIfExists(connections, address);
+  _removeSocketConnectionIfExists(connections, address);
   socket.close();
 }
 
+function closeCorrespondingConnection(connections, address) {
```
platfowner, line 50: minor: it looks like the the socket is converted to an address again in closeSocketSafe.
what about introducing a util function e.g. closeSocketSafeByAddress(connections, address) instead?

p2p/index.js
```diff
@@ -429,6 +439,11 @@ class P2pClient {
     return true;
   }
 
+  clearAllConnectionsInProgress(socket) {
+    removePeerConnection(this.peerConnectionsInProgress, socket.url);
```
platfowner, line 443: minor: what about defining removePeerConnection() as a class method in each of the client and the server, and changing it's name to something like removeFromPeerConnectionsInProgress()?

p2p/index.js
```diff
@@ -429,6 +439,11 @@ class P2pClient {
     return true;
   }
 
+  clearAllConnectionsInProgress(socket) {
```
platfowner, line 442: it looks like this function is handling only 1 input (not all)?
if it's true, then what about clearPeerConnectionInProgress(socket) or something?

p2p/index.js
```diff
@@ -668,6 +678,25 @@ class P2pClient {
     }
   }
 
+  isTheSameJsonRpcUrl(peerCandidateJsonRpcUrl, myJsonRpcUrl) {
+    if (NodeConfigs.HOSTING_ENV === 'local') {
+      const peerCandidateUrl = new URL(peerCandidateJsonRpcUrl);
+      const myUrl = new URL(myJsonRpcUrl);
+      return CommonUtil.isValidPrivateUrl(peerCandidateUrl.hostname) &&
```
platfowner, line 685: question: does this validity check required? do we have any corner cases for this?

p2p/index.js
```diff
@@ -668,6 +678,25 @@ class P2pClient {
     }
   }
 
+  isTheSameJsonRpcUrl(peerCandidateJsonRpcUrl, myJsonRpcUrl) {
+    if (NodeConfigs.HOSTING_ENV === 'local') {
```
platfowner, line 682: question: do we need to handle 'local' env special? just comparing both hostname and port all the time is not applicable?

p2p/index.js
```diff
@@ -668,6 +678,25 @@ class P2pClient {
     }
   }
 
+  isTheSameJsonRpcUrl(peerCandidateJsonRpcUrl, myJsonRpcUrl) {
```
platfowner, line 681: s/is/are/?
If it's not applicable to only json rpc urls, what about generalizing to e.g. 'areTheSameUrls' or 'areIdenticalUrls' and moving it to util.js?

p2p/index.js
```diff
@@ -668,6 +678,25 @@ class P2pClient {
...
+          peerCandidateUrl.port === myUrl.port;
+    } else {
+      return peerCandidateJsonRpcUrl === myJsonRpcUrl;
+    }
+  }
+
+  updatePeerCandidates() {
+    Object.values(this.outbound).forEach(peer => {
+      const jsonRpcUrl = _.get(peer, 'peerInfo.networkStatus.urls.jsonRpc.url');
+      this.peerCandidates[jsonRpcUrl] = { queriedAt: null };
```
platfowner, line 696: minor: it looks like it's more like an initialization?
if so, what about something like 'initPeerCandidates'?

p2p/index.js
```diff
@@ -262,12 +264,20 @@ class P2pClient {
...
+    this.updateP2pState();
+  }
+
   disconnectRandomPeer() {
     if (Object.keys(this.outbound) === 0) {
       return;
     }
-    const randomPeer = _.shuffle(Object.values(this.outbound))[0];
-    closeSocketSafe(this.outbound, randomPeer.socket);
+    const biDirectedConnections = Object.keys(this.outbound).filter(address => {
```
platfowner, line 276: s/biDirected/bidirected/?

and could you leave some note on why we pick a random peer from bidirected connections only?

p2p/index.js
```diff
@@ -262,12 +264,20 @@ class P2pClient {
     }
   }
 
+  closeSocketWithP2pStateUpdate(socket) {
+    closeSocketSafe(this.outbound, socket);
```
platfowner, line 268: what about closing inbound socket together here?

p2p/index.js
```diff
@@ -243,7 +242,10 @@ class P2pClient {
       return NodeConfigs.PEER_CANDIDATE_JSON_RPC_URL;
     } else {
       const notQueriedCandidateEntries = peerCandidatesEntries.filter(([, value]) => {
-        return value.queriedAt === null;
+        // NOTE(minsulee2): this gets stuck if the never queried node gets offline. To avoid this,
+        // the node which queried more than 5 minutes ago can also be considered as notQueried.
+        return value.queriedAt === null ? true :
+            Date.now() - value.queriedAt > NodeConfigs.PEER_CANDIDATE_RECONNECTION_THRESHOLD_MS;
```
platfowner, line 248: minor: 'reconnection' somewhat sounds like disconnect + connect. what about something like 'RETRY' instead?

blockchain-configs/afan-shard/node_params.json
```diff
@@ -37,6 +37,7 @@
   "P2P_WAIT_FOR_ADDRESS_TIMEOUT_MS": 10000,
   "PEER_CANDIDATE_JSON_RPC_URL": "http://localhost:9001/json-rpc",
   "PEER_CANDIDATES_CONNECTION_INTERVAL_MS": 20000,
+  "PEER_CANDIDATE_RECONNECTION_THRESHOLD_MS": 50000,
```
platfowner, line 40: any reason to apply shorter value for afan-shard? what about applying just the same value?

p2p/util.js
minsulee2: Right!! I built `closeSocketSafeByAddress` Thanks!

p2p/index.js
minsulee2: Updated to `clearPeerConnectionInProgress`! Thanks!

p2p/index.js
minsulee2: Yes, Yes. When `PEER_CANDIDATE_JSON_RPC_URL` is set, it is something like `https://localhost/json-rpc`. But, `peerCandidateUrl` is set something like `https://172.0.0.0` so that, both values are not the same if directly compare both local values.

p2p/index.js
minsulee2: Mentioned the reason above! thanks!

p2p/index.js
minsulee2: This is only used for comapring json rpc url for now tho, it can be generalized. Updated! thanks!

p2p/index.js
minsulee2: Ohhhh, I like it thanks!

p2p/index.js
minsulee2: Done!

p2p/index.js
minsulee2: I love RETRY!!

blockchain-configs/afan-shard/node_params.json
minsulee2: Oh, It is mistake! Fixed!

p2p/index.js
minsulee2: Done!

p2p/index.js
minsulee2: Done!!

minsulee2:
PTAL! All addressed!

p2p/util.js
```diff
@@ -16,131 +16,139 @@ const {
...
+    }
+  }
 
-function signMessage(messageBody, privateKey) {
-  if (!CommonUtil.isDict(messageBody)) {
-    logger.error('The message body must be the object type.');
-    return null;
+  static closeSocketSafe(connections, socket) {
+    const address = getAddressFromSocket(connections, socket);
+    P2pUtil._removeSocketConnectionIfExists(connections, address);
```
platfowner, line 47: closeSocketSafeByAddress() can be reused?

p2p/util.js
```diff
@@ -16,131 +16,139 @@ const {
...
-  const body = _.get(message, 'data.body');
-  if (!body || !CommonUtil.isDict(body)) {
-    logger.error('Data body is not included in the message.');
-    return false;
-  }
-  const signature = _.get(message, 'data.signature');
-  if (!signature) {
-    logger.error('Data signature is not included in the message.');
-    return false;
+class P2pUtil {
```
platfowner, line 19: could you change the file name to p2p-util.js as well?

p2p/index.js
platfowner: oh, i mean by changing to a class member method, the fist parameter (this.peerConnectionsInProgress) can be skipped?

p2p/index.js
platfowner: got it, could you add some notes with the example urls?

p2p/index.js
platfowner: 👌

p2p/index.js
```diff
@@ -678,7 +689,13 @@ class P2pClient {
     const myP2pUrl = _.get(this.server.urls, 'p2p.url', '');
     const myJsonRpcUrl = _.get(this.server.urls, 'jsonRpc.url', '');
     if (!peerCandidateJsonRpcUrl || peerCandidateJsonRpcUrl === '' ||
-        peerCandidateJsonRpcUrl === myJsonRpcUrl) {
+        P2pUtil.areIdenticalUrls(peerCandidateJsonRpcUrl, myJsonRpcUrl)) {
+      delete this.peerCandidates[peerCandidateJsonRpcUrl];
+      // NOTE(minsulee2): in case of either a starting node or a seed node
+      if (P2pUtil.areIdenticalUrls(NodeConfigs.PEER_CANDIDATE_JSON_RPC_URL,
+          _.get(this.server.urls, 'jsonRpc.url', ''))) {
+        this.initPeerCandidates();
```
platfowner, line 697: btw, what do you think about generalizing this logic a little bit?
- whenever an outbound connection is made, try to add the peer's json rpc url to this.peerCandidates

p2p/index.js
```diff
@@ -262,12 +252,26 @@ class P2pClient {
...
+  }
+
   disconnectRandomPeer() {
     if (Object.keys(this.outbound) === 0) {
       return;
     }
-    const randomPeer = _.shuffle(Object.values(this.outbound))[0];
-    closeSocketSafe(this.outbound, randomPeer.socket);
+    // NOTE(minsulee2): To avoid breaking connections between ADDRESS_REQUEST and ADDRESS_RESPONSE,
+    // nodes, which is successful bi directed connections, are picked to disconnect.
```
platfowner, line 269: what about 'bidirectionally connected peers are ...'?

p2p/index.js
platfowner: (is it possible to do it for new inbound connections as well?)

p2p/util.js
minsulee2: Done!

p2p/index.js
minsulee2: Done!

p2p/util.js
minsulee2: Updated!

p2p/index.js
minsulee2: I thought the same way tho, I checked the NOTE at line 708 in p2p/index.js that `this.peerCandidates` comes from response. So that, `initPeerCandidates` has been made.

p2p/util.js
minsulee2: Oh, My mistake. It has other case that it is not working with closeSocketSafeByAddress. I left a comment regarding this. Thanks!

minsulee2:
PTAL! All addressed!

p2p/index.js
platfowner: Oh, I mean instead of calling initPeerCandidates() from connectWithPeerCandidateUrl(), can peer info be added to this.peerCandidates whenever new connections are made, e.g. when MessageTypes.ADDRESS_RESPONSE is received on both the client and the server side?

p2p/util.js
platfowner: I see. some notes would be a big help for readers.

p2p/index.js
minsulee2: Oh, I got this! Updated! added initPeerCandidates only in the server-side cuz the client works based on peerCandidates.

minsulee2:
PTAL again!

p2p/index.js
platfowner: minor) now it can be generalized to setPeerCandidate(peerInfo, queriedAt) and used in line 706, 718 as well?

platfowner (approved):
Thanks for the nice fix and refactoring.

p2p/index.js
minsulee2: Updated! Thanks for the nice review!!

minsulee2:
Thanks thanks!! I am merging now!

p2p/index.js
platfowner: ah, i mean singular form 'setPeerCandidate'. :)
thanks any way!

PR #821, opened 2021-12-24
Feature/liayoo/query limits

liayoo:
- Added GET_RESP_BYTES_LIMIT and GET_RESP_MAX_SIBLINGS node params for limits on db queries
- For eval/match queries, only the bytes limit is applied for now. Still trying to think of a good way to restrict the state usage for rules/functions/owners..


test/integration/node.test.js
```diff
@@ -37,6 +38,7 @@ const ENV_VARIABLES = [
     UNSAFE_PRIVATE_KEY: '921cc48e48c876fc6ed1eb02a76ad520e8d16a91487f9c7e03441da8e35a0947',
     BLOCKCHAIN_CONFIGS_DIR: 'blockchain-configs/3-nodes', PORT: 8082, P2P_PORT: 5002,
     ENABLE_GAS_FEE_WORKAROUND: true, ENABLE_EXPRESS_RATE_LIMIT: false,
+    GET_RESP_BYTES_LIMIT: 77819, GET_RESP_MAX_SIBLINGS: 999, // For get_value limit tests
```
platfowner, line 41: 👍

db/index.js
```diff
@@ -472,7 +473,29 @@ class DB {
     if (stateNode === null) {
       return null;
     }
-    return stateNode.toStateSnapshot(options);
+    if (options && options.fromApi) {
```
platfowner, line 476: 👍

common/result-code.js
```diff
@@ -150,6 +150,8 @@ const JsonRpcApiResultCode = {
   PROTO_VERSION_INCOMPATIBLE: 30103,
   // ain_get
   GET_INVALID_OPERATION: 30201,
+  GET_EXCEEDS_MAX_BYTES: 30202,
+  GET_EXCEEDS_MAX_SIBLINGS: 30203,
```
platfowner, line 154: Could you move these two to 'common' area with 300xx values (just after line 146)?

liayoo: Done! PTAL~

blockchain-configs/afan-shard/node_params.json
```diff
@@ -25,6 +25,8 @@
   "ENABLE_STATUS_REPORT_TO_TRACKER": true,
   "ENABLE_TX_SIG_VERIF_WORKAROUND": false,
   "EVENT_HANDLER_PORT": 6000,
+  "GET_RESP_BYTES_LIMIT": 256000000,
```
platfowner, line 28: qq) does this value have any other meaning than something like empirical insights?

blockchain-configs/afan-shard/node_params.json
```diff
@@ -25,6 +25,8 @@
...
+  "GET_RESP_MAX_SIBLINGS": 100000,
```
platfowner, line 29: what about starting from a little smaller value e.g. 50k for safety until more rigorous experiments?

db/index.js
```diff
@@ -472,7 +473,29 @@ class DB {
...
+      if (stateNode.numChildren() > NodeConfigs.GET_RESP_MAX_SIBLINGS) {
+        return {
+          code: JsonRpcApiResultCode.GET_EXCEEDS_MAX_SIBLINGS,
+          message: `The data exceeds the max sibling limit of the requested node: ` +
```
platfowner, line 480: we have been using 'error_message' for those from deeper modules (in this case res.result.result.message?) while just 'message' from shallow modules (e.g. dev client version check).
i've been thinking of using 'message' all the time like other nodejs applications.
what do you think?
if you agree, i think just just leaving some todo in this PR would suffice?

db/index.js
```diff
@@ -472,7 +473,29 @@ class DB {
...
+      if (stateNode.getTreeBytes() > NodeConfigs.GET_RESP_BYTES_LIMIT) {
+        return {
+          code: JsonRpcApiResultCode.GET_EXCEEDS_MAX_BYTES,
+          message: `The data exceeds the max byte limit of the requested node: ` +
+              `${stateNode.getTreeBytes()} > ${NodeConfigs.GET_RESP_BYTES_LIMIT}`
+        };
+      }
+    }
+    const result = stateNode.toStateSnapshot(options);
+    if (options && options.fromApi) {
```
platfowner, line 493: what about applying the bytes limit just based on the tree bytes and well-document it?

db/index.js
```diff
@@ -599,7 +632,12 @@ class DB {
       // No matched local path.
       return null;
     }
-    return this.convertOwnerMatch(this.matchOwnerForParsedPath(localPath), isGlobal);
+    const result = this.convertOwnerMatch(this.matchOwnerForParsedPath(localPath), isGlobal);
```
platfowner, line 635: what about applying the same bytes calculation as the getValue() although it looks like somewhat unnatural?
for example, if the tree bytes exceeds the limit, we just return error for matchOwner(), evalOwner(), etc?

blockchain-configs/afan-shard/node_params.json
liayoo: Not really, I just followed the convention used by firebase. Fyi, the request body size limit is currently set to 100MB, so 256MB for the responses seems reasonable?

platfowner: I see. the scale looks reasonable to me, thanks. btw, could we use less-digit numbers like 200MB or 300MB as we're doing so for other parameters?

db/index.js
liayoo: Just to clarify, do you mean we check the numChildren & treeBytes of the state node at the owner tree of the localPath?

db/index.js
liayoo: Sounds great! Will leave a todo for now~

blockchain-configs/afan-shard/node_params.json
liayoo: Got it! Let's start with 200MB and see if it needs an increase.

db/index.js
platfowner: exactly, less friendly for users but more helpful for system healthiness?

liayoo: I agree! Updated to use checkRespTreeLimits() static function for the get, match, and eval apis. PTAL~

db/index.js
```diff
@@ -680,7 +737,15 @@ class DB {
         resultList.push(this.evalOwner(
             op.ref, op.permission, auth, CommonUtil.toMatchOrEvalOptions(op)));
       }
-    });
+      const bytes = sizeof(resultList);
```
platfowner, line 740: Another approach we can think of seems to be pose restrictions on the size of op_list, e.g.
applying get_op_list_size_limit=50 and set_op_list_size_limit=50 for get and set operations, respectively, as we're applying batch_tx_list_size_limit=50 for batch txs.
what do you think..?

liayoo: That would simplify things and be more intuitive for the users. Thanks for the idea!

liayoo: Added GET_OP_LIST_SIZE_LIMIT as a node param & set_op_list_size_limit as a blockchain param. PTAL~

common/result-code.js
```diff
@@ -144,6 +144,10 @@ const FunctionResultCode = {
 const JsonRpcApiResultCode = {
   // Common code
   SUCCESS: 0,
+  GET_EXCEEDS_MAX_BYTES: 30001,
+  GET_EXCEEDS_MAX_SIBLINGS: 30002,
+  GET_EXCEEDS_OP_LIST_SIZE_LIMIT: 30003,
+  SET_EXCEEDS_OP_LIST_SIZE_LIMIT: 30004,
```
platfowner, line 150: minor) GET_INVALID_OPERATION can be moved to this 'common' area with 3000x value as well?

blockchain-configs/1-node/blockchain_params.json
```diff
@@ -70,7 +70,8 @@
       "3": "http://localhost:3000"
     },
     "tx_bytes_limit": 10000,
-    "batch_tx_list_size_limit": 50
+    "batch_tx_list_size_limit": 50,
+    "set_op_list_size_limit": 1500
```
platfowner, line 74: is this for testing? what about  using the same value (=50)?

blockchain-configs/afan-shard/node_params.json
```diff
@@ -43,6 +46,7 @@
   "PORT": 8080,
   "REST_FUNCTION_CALL_TIMEOUT_MS": 10000,
   "REQUEST_BODY_SIZE_LIMIT": "100mb",
+  "SET_OP_LIST_SIZE_LIMIT": 50,
```
platfowner, line 49: do we need this?

db/index.js
```diff
@@ -640,10 +661,52 @@ class DB {
...
+        }
+      }
+    }
+    return true;
+  }
+
   // TODO(platfowner): Add tests for op.fid.
+  // NOTE(liayoo): This function is only for external uses (APIs).
   get(opList) {
+    if (opList.length > NodeConfigs.GET_OP_LIST_SIZE_LIMIT) {
```
platfowner, line 701: what about adding some validity check here and to /get dev client api?

blockchain-configs/1-node/blockchain_params.json
liayoo: Right, this is increased for unit tests (e.g. 'cannot exceed service state budget' test case in the DB test)

db/index.js
```diff
@@ -1128,6 +1191,15 @@ class DB {
   }
 
   executeMultiSetOperation(opList, auth, timestamp, tx, blockNumber, blockTime) {
+    const setOpListSizeLimit = DB.getBlockchainParam(
+        'resource/set_op_list_size_limit', blockNumber, this.stateRoot);
+    if (blockNumber > 0 && opList.length > setOpListSizeLimit) {
```
platfowner, line 1196: could you add a simple test case for this (maybe in node.test.js)?

platfowner (approved):
LGTM with comments (please address). 
Approving earlier, thanks for addressing this issue!

blockchain-configs/1-node/blockchain_params.json
platfowner: got it thanks. this can be added as notes somewhere?

liayoo: done~

blockchain-configs/afan-shard/node_params.json
liayoo: oops, removed. thanks!

db/index.js
liayoo: added a validity check here only~ (/get API also calls this function) thanks!

db/index.js
liayoo: done!

common/result-code.js
liayoo: done~

db/index.js
liayoo: Ahha...!! I'll add a json rpc error code for the case~

platfowner: thanks!

Issue #817, opened 2021-12-24
[BUG] Need to optimize code for many branches (e.g. accounts)


platfowner:

platfowner:
Fixed, thanks!

Issue #478, opened 2021-07-15
Need state auto optimization features


platfowner:
We need auto state optimization features e.g.
- `.rule: { state_opt: { max_children: 1000 }}`

platfowner:
Fixed, thanks!

PR #824, opened 2021-12-25
Increase min_gc_num_siblings_deleted to 100

platfowner:
Change summary:
- Use Map() for peerCandidates
- Fix integration tests
- Increase min_gc_num_siblings_deleted to 100

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/817 
- https://github.com/ainblockchain/ain-blockchain/issues/478


liayoo (approved):
LGTM, thanks! Let me know if we should restart the test with these new configs!

platfowner:
> LGTM, thanks! Let me know if we should restart the test with these new configs!

Thanks for asking. I think we can apply this when we restart servers for other reasons since it's expected that this change won't be very critical.

PR #829, opened 2021-12-27
Feature/liayoo/block rewards

liayoo:
- Introduced reward blockchain params (type & annual_rate)
- Updated _distributeFee native function
  - Deprecated proposer rewards
  - Added block rewards on top of the tx fee
- Fixed variable names: unitWriteGasLimit -> unitWriteGasAmount


db/functions.js
```diff
@@ -799,6 +797,16 @@ class Functions {
     }, context);
   }
 
+  getBlockRewardMultiplier(context) {
+    const { rewardType, rewardAnnualRate, epochMs } = context;
+    switch (rewardType) {
```
minsulee2, line 802: Do we need the switch here for only one case?!

db/functions.js
```diff
@@ -812,31 +820,31 @@ class Functions {
...
+        if (index === validators.length - 1) {
+          txFee = gasCostTotal - txFeeSum;
+        } else {
+          txFee = gasCostTotal * (validatorStake / totalAtStake);
+          txFeeSum += txFee;
+        }
       }
-      this.incrementConsensusRewards(validator, validatorReward, context);
-    });
+      if (txFee + blockReward > 0) {
```
minsulee2, line 844: What circumstances the `if (txFee + blockReward > 0)` is not met?

db/functions.js
liayoo: Right, I've decided to use a switch as the reward type is subject to change in the future..

db/functions.js
liayoo: It would be 0 when there are no transactions in the block & the blockRewardMultiplier = 0. I thought the check would be good to have for safety reasons.

db/functions.js
minsulee2: I see. Thanks for the explain!

db/functions.js
minsulee2: Oh, I thought there is a small fee even if there is no transactions at all cuz it generates 'block' at least. Very unfortunate for nodes who generates a block without transactions.. :(

minsulee2 (approved):
LGTM!

db/functions.js
liayoo: Well, with the new changes, they'll get 8% interest on their stake even when there are no txs! :)

db/functions.js
```diff
@@ -812,31 +820,31 @@ class Functions {
...
-      if (index === validators.length - 1) {
-        validatorReward = validatorRewardTotal - rewardSum;
-      } else {
-        validatorReward = validatorRewardTotal * (validatorStake / totalAtStake);
-        rewardSum += validatorReward;
+    let txFeeSum = 0;
+    for (let index = 0; index < validators.length; index++) {
+      const validatorAddr = validators[index];
+      const validatorStake = votes[validatorAddr].stake;
+      const blockReward = blockRewardMultiplier * validatorStake;
```
platfowner, line 834: 👍

platfowner, line 834: 👍

db/index.js
```diff
@@ -871,24 +871,34 @@ class DB {
           'resource/account_registration_gas_amount', blockNumber, this.stateRoot);
       const restFunctionCallGasAmount = DB.getBlockchainParam(
           'resource/rest_function_call_gas_amount', blockNumber, this.stateRoot);
+      const rewardType = DB.getBlockchainParam('reward/type', blockNumber, this.stateRoot);
+      const rewardAnnualRate = DB.getBlockchainParam('reward/annual_rate', blockNumber, this.stateRoot);
+      const epochMs = DB.getBlockchainParam('genesis/epoch_ms', blockNumber, this.stateRoot);
+      const blockchainParams = {
```
platfowner, line 877: 👍

platfowner (approved):
LGTM
Thanks for the nice redesign & implementation!

liayoo:

blockchain-configs/afan-shard/blockchain_params.json
```diff
@@ -92,6 +96,6 @@
     "parent_chain_poc": "http://127.0.0.1:8081",
     "reporting_period": 5,
     "max_shard_report": 100,
-    "num_shard_report_deleted": 20
+    "num_shard_report_deleted": 100
```
platfowner, line 99: oh, thanks for noticing and fixing this!

PR #831, opened 2021-12-27
Feature/liayoo/check op list

liayoo:
Added op_list validity check & result code. 
Note that the equivalent check for the set API is already done in Transaction.isInStandardFormat().


minsulee2 (approved):
LGTM!

platfowner (approved):
LGTM
Thanks for handling this!

PR #832, opened 2021-12-27
Feature/minsulee2/keep more api nodes

minsulee2:
Keep a half of whitelisted(block generation) nodes when reorg.

- Disconnect any nodes if ENABLE_JSON_RPC_API is false
- Disconnect whitelisted if they are more than majority, do not disconnect otherwise.

### DIFF
--- a/p2p/index.js
+++ b/p2p/index.js
@@ -271,6 +271,9 @@ class P2pClient {
       return;
     }
     const randomPeerAddress = _.shuffle(addressArray)[0];
+    if (!this.outbound[randomPeerAddress]) {
+      return;
+    }
     this.closeSocketWithP2pStateUpdate(this.outbound[randomPeerAddress].socket);
   }
 
--- a/p2p/index.js
+++ b/p2p/index.js
@@ -292,12 +292,10 @@ class P2pClient {
       const whitelist = this.server.node.db.getValue('/consensus/proposer_whitelist');
       const [whitelisted, notWhitelisted] =
           _.partition(bidirectedConnections, ((address) => whitelist[address]));
-      const whitelistDisconnectThreshold = Math.round(NodeConfigs.MAX_NUM_INBOUND_CONNECTION / 2);
-      if (whitelisted.length > whitelistDisconnectThreshold) {
-        const numDisconnectionCandidates = whitelisted.length - whitelistDisconnectThreshold;
-        const randomWhiteListed = _.shuffle(whitelisted).slice(0, numDisconnectionCandidates);
-        const disconnectionCandidates = _.concat(randomWhiteListed, notWhitelisted);
-        this.pickRandomPeerAndDisconnect(disconnectionCandidates);
+      const whitelistDisconnectThreshold = Math.floor(NodeConfigs.MAX_NUM_INBOUND_CONNECTION / 2);
+      // NOTE(minsulee2): Keep less than majority whitelisted.
+      if (whitelisted.length >= whitelistDisconnectThreshold) {
+        this.pickRandomPeerAndDisconnect(whitelisted);
       } else {
         this.pickRandomPeerAndDisconnect(notWhitelisted);
       }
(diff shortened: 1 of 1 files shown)


p2p/index.js
```diff
@@ -266,6 +266,14 @@ class P2pClient {
     this.updateP2pState();
   }
 
+  pickRandomPeerAndDisconnect(addressArray) {
+    if (addressArray.length === 0) {
+      return;
+    }
+    const randomPeerAddress = _.shuffle(addressArray)[0];
+    this.closeSocketWithP2pStateUpdate(this.outbound[randomPeerAddress].socket);
```
liayoo, line 274: I don't think it's guaranteed that the randomPeerAddress is a key in this.outbound. Could you add a check?

minsulee2: Indeed! Updated! Thanks!

minsulee2:
PTAL! All addressed!

p2p/index.js
```diff
@@ -275,8 +286,22 @@ class P2pClient {
     const bidirectedConnections = Object.keys(this.outbound).filter(address => {
       return Object.keys(this.server.inbound).includes(address);
     });
-    const randomPeerAddress = _.shuffle(bidirectedConnections)[0];
-    this.closeSocketWithP2pStateUpdate(this.outbound[randomPeerAddress].socket);
+    if (NodeConfigs.ENABLE_JSON_RPC_API) {
+      const whitelist = this.server.node.db.getValue('/consensus/proposer_whitelist');
```
platfowner, line 290: 👍

p2p/index.js
```diff
@@ -275,8 +286,22 @@ class P2pClient {
     const bidirectedConnections = Object.keys(this.outbound).filter(address => {
       return Object.keys(this.server.inbound).includes(address);
     });
-    const randomPeerAddress = _.shuffle(bidirectedConnections)[0];
-    this.closeSocketWithP2pStateUpdate(this.outbound[randomPeerAddress].socket);
+    if (NodeConfigs.ENABLE_JSON_RPC_API) {
```
platfowner, line 289: could you add some note that this condition is for 'API server nodes'?

p2p/index.js
```diff
@@ -275,8 +286,22 @@ class P2pClient {
...
+      const whitelist = this.server.node.db.getValue('/consensus/proposer_whitelist');
+      const [whitelisted, notWhitelisted] =
+          _.partition(bidirectedConnections, ((address) => whitelist[address]));
+      const whitelistDisconnectThreshold = Math.round(NodeConfigs.MAX_NUM_INBOUND_CONNECTION / 2);
+      if (whitelisted.length > whitelistDisconnectThreshold) {
+        const numDisconnectionCandidates = whitelisted.length - whitelistDisconnectThreshold;
+        const randomWhiteListed = _.shuffle(whitelisted).slice(0, numDisconnectionCandidates);
+        const disconnectionCandidates = _.concat(randomWhiteListed, notWhitelisted);
```
platfowner, line 297: if we can define the role of the bridge nodes is to connect whitelisted nodes (proposers) and the others and interpret `whitelisted.length > whitelistDisconnectThreshold` as too less connections to non-whitelisted nodes, could we just let it try to disconnect one of the whitelisted peer connections?
this strategy can be summarized:
- too many whitelisted -> less whitelisted (i.e. disconnect) and more non-whitelisted
- too few whitelisted  -> less non-whitelisted (i.e. disconnect) and more whitelisted (in probability)

p2p/index.js
platfowner: Also could you add a todo for introducing a separate flag as we can encounter such situations that we want to enable the json rpc api but not to assign the 'bridge node' role?

platfowner (approved):
LGTM with some comments.
Thanks for the quick delivery of the nice ideas and implementations!

p2p/index.js
minsulee2: Done!

p2p/index.js
minsulee2: Agreed. I keep it simpler for now that whitelisted sustains less than majority of nodes that is keeping 2 nodes if MAX_INBOUND equals to 5 for example. Thanks for the nice intuition!!

minsulee2:
Thanks for the reviews! Merging now!

PR #836, opened 2021-12-28
Update error trap logic when blockPool.hashToBlockInfo[lastHash] is undefined

minho-comcom-ai:
…ndefined


consensus/index.js
```diff
@@ -567,7 +567,7 @@ class Consensus {
     for (const vote of lastVotes) {
       blockPool.addSeenVote(vote);
     }
-    if (!blockPool.hashToBlockInfo[lastHash].notarized) {
+    if (!blockPool.hashToBlockInfo[lastHash] || !blockPool.hashToBlockInfo[lastHash].notarized) {
```
liayoo, line 570: question) Were you able to see the blockPool.hashToBlockInfo[lastHash] case, or is this for safety? Asking because the last block should be in hashToBlockInfo at this point

liayoo (approved):
LGTM with a question. Thanks!

consensus/index.js
minho-comcom-ai: I found this when I tried to launch a new node with staging-api's peers.

liayoo: I see! Thanks for fixing it. Let me merge this for our next release.

PR #837, opened 2021-12-28
Fix isAlive always true bug

liayoo:
- Fixed isAlive always being true
- Updated endpointUrls & checkin/checkoutIds in tools configs
- Updated start_node_genesis_gcp.sh to set PEER_CANDIDATE_JSON_RPC_URL to node 0's address only for nodes 0-9


minhoryang:
Why don't we just swap the ordering of 'Object.assign()' argument?

liayoo:
> Why don't we just swap the ordering of 'Object.assign()' argument?

Thanks for asking! That would totally work, but since only the isAlive value needs changing, it just seemed simpler~

start_node_genesis_gcp.sh
```diff
@@ -168,7 +168,10 @@ elif [[ $SEASON = 'summer' ]]; then
     KEYSTORE_DIR=testnet_prod_keys
 elif [[ "$SEASON" = "sandbox" ]]; then
     export BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/testnet-sandbox
-    export PEER_CANDIDATE_JSON_RPC_URL="http://130.211.244.169:8080/json-rpc"
+    if [[ $NODE_INDEX -lt 10 ]]; then
```
platfowner, line 171: 👍

platfowner (approved):
LGTM
Thanks!

liayoo:

PR #838, opened 2021-12-28
Feature/liayoo/extend state rule

liayoo:
Added max_height, max_size, max_bytes state rule props


db/index.js
```diff
@@ -2317,26 +2318,66 @@ class DB {
...
+      const maxChildren = stateRuleObj[RuleProperties.MAX_CHILDREN];
+      const maxChildrenEvalResult = Object.keys(newValue).length <= maxChildren;
+      if (!maxChildrenEvalResult) {
+        return {
+          ruleString: JSON.stringify(stateRuleObj),
+          evalResult: maxChildrenEvalResult,
+        };
+      }
+    }
+    const { height, size } = CommonUtil.getObjectHeightAndSize(newValue);
```
platfowner, line 2359: could you move the parts for height and size below before `if (!CommonUtil.isDict(newValue)) {` so that the size rule is checked?

common/common-util.js
```diff
@@ -919,6 +919,24 @@ class CommonUtil {
     }
     return maxOccurrences;
   }
+
+  static getObjectHeightAndSize(value) {
```
platfowner, line 923: what about moving this to state-util.js?

db/index.js
liayoo: Do you mean right before L2343? I thought that if the new value is a non-object, it'd only need the byte check?

common/common-util.js
liayoo: done~

db/state-util.js
```diff
@@ -203,15 +203,27 @@ function sanitizeRuleConfig(rule) {
...
+            CommonUtil.numberOrZero(rule[RuleProperties.STATE][RuleProperties.MAX_SIZE]));
+      }
+      if (rule[RuleProperties.STATE].hasOwnProperty(RuleProperties.MAX_BYTES)) {
+        CommonUtil.setJsObject(sanitized, [RuleProperties.STATE, RuleProperties.MAX_BYTES],
+            CommonUtil.numberOrZero(rule[RuleProperties.STATE][RuleProperties.MAX_BYTES]));
       }
       if (rule[RuleProperties.STATE].hasOwnProperty(RuleProperties.GC_MAX_SIBLINGS)) {
         CommonUtil.setJsObject(sanitized, [RuleProperties.STATE, RuleProperties.GC_MAX_SIBLINGS],
-            rule[RuleProperties.STATE][RuleProperties.GC_MAX_SIBLINGS]);
+            CommonUtil.numberOrZero(rule[RuleProperties.STATE][RuleProperties.GC_MAX_SIBLINGS]));
```
platfowner, line 222: 👍

platfowner (approved):
LGTM

liayoo:
Thanks for the review!! Now merging~

PR #840, opened 2021-12-29
Deprecate dev flags

liayoo:
- Removed dev flags
  - enableReceiptsRecording => always enabled
  - enableNtpSync => always enabled
  - enableP2pMessageTags => always enabled
  - enableWsCompression => always enabled
    - R15 currently running w/ this option but no significant diff in CPU util
  - enableTxBroadcastLimit => always disabled
- Added tag checking before broadcasting msgs to peers
  - R15 currently running w/ this option

### DIFF
--- a/p2p/index.js
+++ b/p2p/index.js
@@ -374,13 +374,21 @@ class P2pClient {
       return;
     }
     const stringPayload = JSON.stringify(payload);
-    const tagSet = new Set(tags);
-    Object.entries(this.outbound).forEach(([address, node]) => {
-      if (!tagSet.has(address) &&
-          _.get(node, 'peerInfo.consensusStatus.state') === ConsensusStates.RUNNING) {
-        node.socket.send(stringPayload);
-      }
-    });
+    if (DevFlags.enableP2pMessageTagsChecking) {
+      const tagSet = new Set(tags);
+      Object.entries(this.outbound).forEach(([address, node]) => {
+        if (!tagSet.has(address) &&
+            _.get(node, 'peerInfo.consensusStatus.state') === ConsensusStates.RUNNING) {
+          node.socket.send(stringPayload);
+        }
+      });
+    } else {
+      Object.values(this.outbound).forEach((node) => {
+        if (_.get(node, 'peerInfo.consensusStatus.state') === ConsensusStates.RUNNING) {
+          node.socket.send(stringPayload);
+        }
+      });
+    }
     logger.debug(`SENDING: ${JSON.stringify(consensusMessage)}`);
   }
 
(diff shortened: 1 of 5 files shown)


liayoo:
@platfowner I wasn't sure if the enableStateTreeTransfer dev flag should be deprecated or not. Let me know what you think~

minsulee2 (approved):
LGTM!

p2p/index.js
```diff
@@ -418,24 +418,19 @@ class P2pClient {
...
     const stringPayload = JSON.stringify(payload);
-    if (DevFlags.enableTxBroadcastLimit) {
-      _.shuffle(Object.values(this.outbound)).slice(0, 2).forEach((node) => {
-        node.socket.send(stringPayload);
-      });
-    } else {
-      Object.values(this.outbound).forEach((node) => {
+    const tagSet = new Set(tags);
+    Object.entries(this.outbound).forEach(([address, node]) => {
+      if (!tagSet.has(address)) {
```
platfowner, line 430: 👍

p2p/index.js
```diff
@@ -367,17 +367,17 @@ class P2pClient {
...
     if (!payload) {
       logger.error('The consensus msg cannot be broadcasted because of msg encapsulation failure.');
       return;
     }
     const stringPayload = JSON.stringify(payload);
-    Object.values(this.outbound).forEach((node) => {
-      if (_.get(node, 'peerInfo.consensusStatus.state') === ConsensusStates.RUNNING) {
+    const tagSet = new Set(tags);
+    Object.entries(this.outbound).forEach(([address, node]) => {
+      if (!tagSet.has(address) &&
```
platfowner, line 379: 👍 

thanks for the good idea!
btw, if the tags can be modified by an intermediate blockchain node, could this expose new vulnerabilities?

liayoo: Hmm, that's true, I haven't thought of that. We could add signatures instead of addresses, or keep a record of received messages from each peer to determine if a peer has seen the message or not? 
If you think it's too much of a risk without these measures, we could keep this feature off with a flag & try turning it on with additional security measures?

platfowner: right, i agree that we can just leave some todos (or comments) for now and try to revisit with the solutions you mentioned. thanks!

platfowner (approved):
LGTM with some comments.
Thanks!

p2p/index.js
liayoo: Added a dev flag & a todo. This made me realize that we should keep the enableP2pMessageTags dev flag and turn it off for the mainnet since a node can deliberately send messages with huge tags. What do you think?

platfowner: We don't have any size limits on the p2p messages yet? If so, huge messages still can be sent even though the tags are disabled? cc: @minsulee2

liayoo: Right.. I think we need to set the [maxPayload](https://github.com/websockets/ws/blob/master/doc/ws.md#new-websocketserveroptions-callback) option. From the sandbox testing, a block seems to be around 100-170KB (tx_bytes_limit = 10000), and when requesting a chain we receive 20 blocks max, so how about setting it to ~5MB?

liayoo: Or in the worst case: 10KB/tx * (100tx +a consensus txs)/block * 20 blocks ≈ 20MB +a => 50MB?

platfowner: for now, that seems to be reasonable. please feel free to just add a todo in this pr. thanks!

liayoo: left a todo~ thanks!

liayoo:
Thanks for the review & finding a critical vulnerability. Now merging!

p2p/index.js
```diff
@@ -418,21 +426,22 @@ class P2pClient {
...
     const payload = P2pUtil.encapsulateMessage(MessageTypes.TRANSACTION, { transaction, tags });
     if (!payload) {
       logger.error('The transaction cannot be broadcasted because of msg encapsulation failure.');
       return;
     }
     const stringPayload = JSON.stringify(payload);
-    if (DevFlags.enableTxBroadcastLimit) {
-      _.shuffle(Object.values(this.outbound)).slice(0, 2).forEach((node) => {
-        node.socket.send(stringPayload);
+    if (DevFlags.enableP2pMessageTagsChecking) {
```
platfowner, line 436: 👍

PR #843, opened 2021-12-29
Add epoch transition interval renewal logic

liayoo:
- Added epoch transition interval renewal logic
- Decreased HEALTH_THRESHOLD_EPOCH: 600 -> 10
- Added chainSyncInProgress to p2p status


liayoo:
It would be nice if we can run R16 with this PR included!

consensus/index.js
```diff
@@ -1279,6 +1298,11 @@ class Consensus {
     return result;
   }
 
+  isHealthy(lastFinalizedBlock) {
```
minsulee2, line 1301: What about `isHealthyBlock`?

consensus/constants.js
```diff
@@ -3,7 +3,7 @@ const ConsensusConsts = {
   DAY_MS: 86400000,
   MAX_INVALID_BLOCKS_ON_MEM: 100,
   MAX_FINALIZED_BLOCK_INFO_ON_MEM: 10000,
-  HEALTH_THRESHOLD_EPOCH: 600,
+  HEALTH_THRESHOLD_EPOCH: 10,
```
platfowner, line 6: what do you think about moving some of these values to blockchain-params.json or node-params.json? e.g.
to blockchain-params.json:
- STAKE_LOCKUP_EXTENSION
- MAX_INVALID_BLOCKS_ON_MEM (does this affect block producing?)
to node-params.json:
- MAX_FINALIZED_BLOCK_INFO_ON_MEM (?)
to be removed:
- DAY_MS (used nowhere?)

consensus/index.js
```diff
@@ -137,36 +138,54 @@ class Consensus {
...
-          const iNTPData = await ntpsync.ntpLocalClockDeltaPromise();
-          logger.debug(`(Local Time - NTP Time) Delta = ${iNTPData.minimalNTPLatencyDelta} ms`);
-          this.timeAdjustment = iNTPData.minimalNTPLatencyDelta;
-          this.ntpData = { ...iNTPData, syncedAt: Date.now() };
-        } catch (err) {
-          logger.error(`ntpsync error: ${err} ${err.stack}`);
+      try {
+        const lastBlock = this.node.bc.lastBlock();
+        if (lastBlock && lastBlock.number > 0 && !this.isHealthy(lastBlock) &&
+            CommonUtil.timestampExceedsThreshold(
```
minsulee2, line 144: Can this be in `this.isHealthy`?

consensus/index.js
```diff
@@ -137,36 +138,54 @@ class Consensus {
...
+            const iNTPData = await ntpsync.ntpLocalClockDeltaPromise();
+            logger.debug(`(Local Time - NTP Time) Delta = ${iNTPData.minimalNTPLatencyDelta} ms`);
+            this.timeAdjustment = iNTPData.minimalNTPLatencyDelta;
+            this.ntpData = { ...iNTPData, syncedAt: Date.now() };
+          } catch (err) {
+            logger.error(`ntpsync error: ${err} ${err.stack}`);
+          }
+        }
+        currentTime -= this.timeAdjustment;
+        const absEpoch = Math.floor((currentTime - this.startingTime) / epochMs);
```
minsulee2, line 170: Need to add Math.abs()? or need to pos/neg checks?

consensus/index.js
liayoo: I think the two mean different things. timestampExceedsThreshold() is checked so that it doesn't trap itself in an infinite loop (right after the renewal, the consensus may still be unhealthy, for example, if no blocks are finalized within an epoch). And if it's been renewed a while ago, the consensus process could be healthy but this.lastEpochTransitionRenewed timestamp is old.

consensus/constants.js
liayoo: Thanks for the idea! The way you organized them seems perfect. 
Fyi, MAX_INVALID_BLOCKS_ON_MEM wouldn't affect the state but I think it would be better if all nodes share the same value, so blockchain params should work.

consensus/index.js
liayoo: Thanks for catching it! Updated!

consensus/index.js
liayoo: This is meant to be more like isConsensusHealthy(). I'll update to make it clearer!

minsulee2: Oh, Right Thanks!

consensus/constants.js
platfowner: great, thanks!
btw, i forgot to ask about HEALTH_THRESHOLD_EPOCH.
do you think there could be any chances that we need to 'change' this value in the future?
if so, this can go to blockchain params as well?

consensus/index.js
minsulee2: Thanks for the explain!

platfowner (approved):
LGTM
Thanks for the nice solution and refactoring!

consensus/constants.js
liayoo: Ah that's another good idea! Moved it to blockchain params!

liayoo:
Thanks for the reviews~ Now merging!

PR #847, opened 2021-12-29
Ignore minimalNTPLatencyDelta that's too big/small

liayoo:
Sandbox node 14 stopped because of minimalNTPLatencyDelta that's too big, causing currentTime to be less than startingTime. Updated to ignore minimalNTPLatencyDelta values that are too small/big.

Following is the last ntpsync result from node 14 before it stopped.
```
{
    "averageNTPDelta": 962443388956.375,
    "averageNTPLatency": 9.75,
    "minimalNTPLatencyDelta": 3849773555827,
    "minimalNTPLatency": 8,
    "totalSampleCount": 4,
    "syncedAt": 1640784755831
}
```


platfowner (approved):
LGTM
Thanks!

liayoo:
Thanks for the review! It seems the issue has been resolved after this fix.

Issue #853, opened 2021-12-30
[BUG] broken tests integration/consensus.test.js


platfowner:
```
yarn run test_integration_consensus -g 'can penalize malicious validators'
yarn run v1.22.10
$ BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/3-nodes ./node_modules/mocha/bin/mocha --timeout 640000 test/integration/consensus.test.js -g 'can penalize malicious validators'


  Consensus
    Penalties
      1) can penalize malicious validators 


  0 passing (2m)
  1 failing

  1) Consensus
       Penalties
         can penalize malicious validators :

      AssertionError: expected 1643440067262 to deeply equal NaN
      + expected - actual

      -1643440067262
      +NaN
      
      at Context.<anonymous> (test/integration/consensus.test.js:575:14)



error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
it seems the 3rd argument is missing for getLockupExtensionForNewOffenses(). 
could you take a look? @liayoo

platfowner:
Fixed, thanks!

PR #854, opened 2021-12-30
Add validator whitelist

liayoo:
- Added /consensus/validator_whitelist
- Added genesis_validator_whitelist in the blockchain params
- Fixed bugs in rule util
- Added checks for validators & total_at_stake in proposal tx value
- Fixed broken consensus integration test (#853)


platfowner (approved):
LGTM
Thanks!

PR #857, opened 2021-12-30
Fix api access tool scripts

platfowner:


liayoo (approved):
LGTM. Thanks for fixing these!

PR #860, opened 2021-12-30
Expose safe env variables only by /get_config

platfowner:


common/constants.js
```diff
@@ -652,6 +652,32 @@ const APP_DEPENDENT_SERVICE_TYPES = [
...
+  'HOSTING_ENV',
+  'DEBUG',
+  'PORT',
+  'STAKE',
+  'SYNC_MODE',
+];
+
+function getEnvVariables() {
+  const envs = {};
+  for (const envName of Object.keys(NodeParams)) {
```
liayoo, line 666: question) is NodeParams used instead of NodeConfigs in order to get more "raw" values?

liayoo (approved):
LGTM with a question, thanks!

common/constants.js
platfowner: currently, we're showing such env variables 1) overwriting node params or 2) included in EXPOSED_ENV_VARIABLE_LIST. so if there are env variables not in 1) case, it needs to be added to EXPOSED_ENV_VARIABLE_LIST manually. (if an env variable is in node configs but not in node params, the value is not used in overwriting node params so it should go to the list.)

platfowner:
Thanks for the fast review, let me merge it asap.

Issue #862, opened 2021-12-30
Need to handle dup variable labels in paths


platfowner:

platfowner:
Fixed, thanks!

PR #863, opened 2021-12-30
Add dup check for variable labels in state paths

platfowner:
Change summary:
- Add dup check for variable labels in state paths
- Fix typo in grafana.json

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/862

### DIFF
--- a/db/state-util.js
+++ b/db/state-util.js
@@ -234,11 +234,17 @@ function makeWriteRuleCodeSnippet(ruleString) {
   return WRITE_RULE_CODE_SNIPPET_PREFIX + ruleString;
 }
 
-function getVariableLabels(parsedRulePath) {
+function getVariableLabels(parsedRulePath, allowVariableLabels) {
   const variableLabels = {};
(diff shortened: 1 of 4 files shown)


db/state-util.js
```diff
@@ -582,6 +611,9 @@ function isValidConfigTreeRecursive(
       }
     }
     subtreePath.pop();
+    if (params.variableLabels && params.variableLabels[label] !== undefined) {
+      delete params.variableLabels[label];
```
liayoo, line 615: aren't we not allowing the same var labels? why do we need to delete here?

platfowner: oh, it corresponds to the 'pop' operation in recursion.

liayoo: ahhhhh yes. thanks for the explanation!

liayoo (approved):
LGTM, thanks!

platfowner:
Thanks for the swift review, now merging.. @liayoo

PR #870, opened 2021-12-31
Added min_gas_price resource blockchain param

liayoo:


blockchain-configs/1-node/blockchain_params.json
```diff
@@ -80,7 +80,8 @@
     "set_op_list_size_limit": 1500,
     "min_gc_num_siblings_deleted": 100,
     "block_chunk_size": 100000000,
-    "snapshot_chunk_size": 100000000
+    "snapshot_chunk_size": 100000000,
```
platfowner, line 83: ah, 100mb could be too big for snapshot sharing between blockchain peers.
if this value is small enough (e.g. 1mb), the saved snapshot file can be directly streamed between peers **without modification**.
what do you think?

liayoo: 1MB sounds good to me~

db/index.js
```diff
@@ -1700,6 +1700,14 @@ class DB {
           TxResultCode.TX_NO_TX_BODY,
           `[${LOG_HEADER}] Missing tx_body: ${JSON.stringify(tx)}`, 0);
     }
+    const gasPrice = tx.tx_body.gas_price;
+    const minGasPrice = DB.getBlockchainParam('resource/min_gas_price', blockNumber, this.stateRoot);
+    if (!CommonUtil.isInteger(gasPrice) || (gasPrice !== 0 && gasPrice < minGasPrice)) {
```
platfowner, line 1705: what about checking `gasPrice < 0` case as well?

blockchain-configs/1-node/blockchain_params.json
liayoo: I'll reduce the block chunk size to 1mb as well then~ (although it's not being used right now..)

db/index.js
liayoo: gasPrice < minGasPrice should cover it unless we set the min_gas_price to a negative number? should we still add the check to be extra careful?

blockchain-configs/1-node/blockchain_params.json
```diff
@@ -80,7 +80,8 @@
     "set_op_list_size_limit": 1500,
     "min_gc_num_siblings_deleted": 100,
     "block_chunk_size": 100000000,
```
platfowner, line 82: we don't need this any more? 
can this value just deleted for now?

platfowner (approved):
LGTM with some comments.
Thanks!!

blockchain-configs/1-node/blockchain_params.json
liayoo: done~

PR #872, opened 2021-12-31
Fix missing chain ids.

liayoo:
Extra thanks to @cshcomcom who found & located the issue!!


platfowner (approved):
LGTM
Thanks for fixing it!!

liayoo:
Thanks for the fast review! Just confirmed that the APIs now work with a non-zero chain id. Now merging~

PR #888, opened 2022-01-06
Add block_timestamp at the highest level in the snapshot for convenience

platfowner:


liayoo:
Question) Where do you plan to use this timestamp? Also, I assume this doesn't affect building states from the old snapshots?

platfowner:
> Question) Where do you plan to use this timestamp? Also, I assume this doesn't affect building states from the old snapshots?

Ah, it's for future use cases. When I tried to set block timestamp snapshot in db, I thought it would be nice if it's available on the top level of the snapshot. Currently, we don't have clear use case yet.

liayoo (approved):
LGTM, thanks!

Issue #906, opened 2022-01-11
Failed to add new block, TypeError: Cannot read properties of undefined (reading 'timestamp')


liayoo:
While test deploying to dev with a command `bash deploy_blockchain_incremental_gcp.sh dev lia 0 0 0 --keystore --no-keep-data --no-keep-code`, the following error was observed:

[NODE] [mergeChainSegment] Failed to add new block (1 / 0xda250c82aaa8266f6230c8b27249f3b5554dc31d34996b5c5d371b9699f1410e) to chain: TypeError: Cannot read properties of undefined (reading 'timestamp')
    at BlockchainNode.buildBlockchainSnapshot (/home/ain-blockchain-2022-01-11T05-42/node/index.js:373:34)
    at Function.validateStateProofHash (/home/ain-blockchain-2022-01-11T05-42/consensus/index.js:757:31)
    at Function.validateAndExecuteBlockOnDb (/home/ain-blockchain-2022-01-11T05-42/consensus/index.js:848:17)
    at BlockchainNode.mergeChainSegment (/home/ain-blockchain-2022-01-11T05-42/node/index.js:755:19)
    at P2pClient.handleChainSegment (/home/ain-blockchain-2022-01-11T05-42/p2p/index.js:690:42)
    at WebSocket.<anonymous> (/home/ain-blockchain-2022-01-11T05-42/p2p/index.js:626:16)
    at WebSocket.emit (node:events:390:28)
    at Receiver.receiverOnMessage (/home/ain-blockchain-2022-01-11T05-42/node_modules/ws/lib/websocket.js:1008:20)
    at Receiver.emit (node:events:390:28)
    at Receiver.dataMessage (/home/ain-blockchain-2022-01-11T05-42/node_modules/ws/lib/receiver.js:517:14)

platfowner:
Fixed, thanks!

PR #907, opened 2022-01-11
Handle mismatched genesis block hash case and null value cases properly

platfowner:
Change summary:
- Add chain segment precheck:
  - Handle genesis block hash mismatch case properly
  - Handle block last_hash mismatch case properly
- Handle null value cases of lastBlock and lastNotarizedBlock properly
- Fix undefined block bug with --no-keep-code --no-keep-data deploy

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/906


node/index.js
```diff
@@ -355,11 +355,11 @@ class BlockchainNode {
   async writeSnapshot(blockNumber) {
     const LOG_HEADER = 'writeSnapshot';
 
-    const snapshot = this.buildBlockchainSnapshot(blockNumber, this.stateManager.getFinalRoot());
+    const block = this.bc.getBlockByNumber(blockNumber);
```
liayoo, line 358: how about checking if block !== null just in case?

p2p/index.js
```diff
@@ -675,13 +675,25 @@ class P2pClient {
...
       return;
     }
+    const firstBlock = chainSegment.length > 0 ? chainSegment[0] : null;
+    const genesisBlockHash = this.server.node.bc.genesisBlockHash;
+    if (firstBlock && firstBlock.number === 0 && firstBlock.hash !== genesisBlockHash) {
+      const p2pUrl = P2pUtil.getP2pUrlFromAddress(this.outbound, address);
+      logger.error(
+          `[${LOG_HEADER}] Genesis block from ${address} (${p2pUrl}) has a mismatched hash: ${firstBlock.hash} / ${genesisBlockHash}`
+          + `\n${new Error().stack}.`);
+      await CommonUtil.sleep(NodeConfigs.P2P_GENESIS_BLOCK_HASH_MISMATCH_SLEEP_MS);
```
liayoo, line 692: just curious, was there an issue without the sleep()?

consensus/index.js
```diff
@@ -1047,20 +1054,26 @@ class Consensus {
...
+      logger.info(`[${LOG_HEADER}] Current lastNotarizedBlock: ${lastNotarizedBlock}`);
+    }
+    for (const blockInfo of catchUpList) {
       logger.debug(`[${LOG_HEADER}] Adding notarized chain's block: ` +
           `${JSON.stringify(blockInfo, null, 2)}`);
       if (!blockInfo.block || !blockInfo.proposal) {
         break;
       }
-      if (blockInfo.block.number < lastNotarizedBlock.number) {
+      if (lastNotarizedBlock && blockInfo.block.number < lastNotarizedBlock.number) {
```
liayoo, line 1076: thanks for adding the check! 👍

consensus/index.js
```diff
@@ -1124,7 +1137,8 @@ class Consensus {
 
   getSnapDb(latestBlock) {
     const LOG_HEADER = 'getSnapDb';
-    const lastFinalizedHash = this.node.bc.lastBlock().hash;
+    const lastBlock = this.node.bc.lastBlock();
+    const lastFinalizedHash = lastBlock ? lastBlock.hash : null;
```
liayoo, line 1141: ditto 👍

p2p/index.js
platfowner: ah, it seems there are two cases we need to handle:
- malicious peer with wrong chain segments (case 1) 
- new node starts with a wrong genesis block (case 2)

To cover both cases, we cannot just stop the new node when there is a block hash mismatch but to try again to get chain segments from other peers. But if it's in case 2, it will repeatedly make block hash mismatch errors and this is why some sleep time was added. Please let me know if you have other ideas.

p2p/index.js
```diff
@@ -675,13 +675,52 @@ class P2pClient {
     return true;
   }
 
-  handleChainSegment(number, chainSegment, catchUpInfo, socket) {
+  precheckChainSegment(chainSegment, peerAddress) {
```
liayoo, line 678: 👍

p2p/index.js
```diff
@@ -675,13 +675,52 @@ class P2pClient {
...
+    const LOG_HEADER = 'precheckChainSegment';
+
+    const p2pUrl = P2pUtil.getP2pUrlFromAddress(this.outbound, peerAddress);
+    for (let i = 0; i < chainSegment.length; i++) {
+      const block = chainSegment[i];
```
liayoo, line 683: Since this is a precheck, we can't be sure the block is in a proper format. What do you think of adding a check like this: `if (!block) { return false; }` ?

p2p/index.js
liayoo: Thanks for the explanation! I think that's reasonable. Maybe in a later PR, we could introduce a peer blacklist so it can deal with both cases?

node/index.js
platfowner: Right. Just tried to make the blockchain snapshot build possible even without the block (as this snapshot is for debugging purpose). PTAL

p2p/index.js
platfowner: I think that's a good idea, which was in my mind too. Just added a todo.

p2p/index.js
platfowner: Just tried to add a basic format checking with hasRequiredFields() and a todo. PTAL

platfowner:
All comments addressed. PTAL~ @liayoo

node/index.js
liayoo: Ah, that's a good idea! Thanks!

liayoo (approved):
LGTM, thanks for handling the issue!

platfowner:

Issue #916, opened 2022-01-13
[BUG] Need to set max_state_tree_size_per_byte with 0.00625 (= 1/160)


platfowner:
We need to set max_state_tree_size_per_byte with 0.00625 (= 1/160) as 160 bytes is actual lower bound of bytes per state:
https://github.com/ainblockchain/ain-blockchain/blob/develop/db/state-node.js#L73

platfowner:
Done, thanks!

PR #917, opened 2022-01-13
Compute available size by available bytes and replace tree size checks with tree bytes checks

platfowner:
Change summary:
- Compute availableTreeSize by availableTreeBytes
- Update max_state_tree_size_per_byte with 0.00625 (= 1/160)
- Replace tree size checks with tree bytes checks

Related code:
```
  computeNodeBytes() {
    return sizeof(this.value) + 160;
  }
```

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/916


liayoo:
Thanks for handling this! I have a question regarding the tree size checking logic in checkStateGasBudgets(); are we keeping it?

platfowner:
> Thanks for handling this! I have a question regarding the tree size checking logic in checkStateGasBudgets(); are we keeping it?

Right, just tried to replace the existing tree size checks with tree bytes checks. PTAL

platfowner:

db/index.js
```diff
@@ -1431,98 +1432,61 @@ class DB {
...
+          return {
+            code: TxResultCode.GAS_EXCEED_STATE_BUDGET_LIMIT_FOR_FREE_TIER,
+            message: `Exceeded state budget limit for free tier ` +
+                `(${stateFreeTierUsage[StateLabelProperties.TREE_BYTES]} > ${budgets.freeStateBudget})`
+          };
         }
         // else, we allow apps without stakes
       } else {
         const singleAppStateBudget = budgets.appsStateBudget * appStake / appStakesTotal;
-        const singleAppTreeSizeBudget = budgets.appsStateBudget * appStake / appStakesTotal;
```
platfowner, line 1508: was this a bug? (appsStateBudget -> appsTreeSizeBudget?)

db/index.js
```diff
@@ -1431,98 +1432,61 @@ class DB {
...
+            code: TxResultCode.GAS_EXCEED_STATE_BUDGET_LIMIT_FOR_APP,
+            message: `Exceeded state budget limit for app ${appName} ` +
+                `(${appStateUsage[StateLabelProperties.TREE_BYTES]} > ${singleAppStateBudget})`
+          };
         }
       }
     }
-    return true;
+    return {
+      code: 0
```
liayoo, line 1488: minor) TxResultCode.SUCCESS?

db/index.js
liayoo: Oh, wow. Yes. Thanks for fixing/removing it!

node/index.js
```diff
@@ -477,40 +477,38 @@ class BlockchainNode {
   getStateUsageWithStakingInfo(appName) {
     if (!appName) return null;
     const stateTreeHeightLimit = this.getBlockchainParam('resource/state_tree_height_limit');
+    const maxStateTreeSizePerByte = this.getBlockchainParam('resource/max_state_tree_size_per_byte');
```
liayoo, line 480: Do we need max_state_tree_size_per_byte and availableTreeSize now that state budget limits are only calculated with bytes?

platfowner: Thanks for asking. 
I also had the same question and my conclusion was that availableTreeSize can be helpful for users to plan their state use as it's more accessible than expecting them to compute it by themself?

db/index.js
platfowner: Right, just fixed. Thanks!

platfowner:
Additional comments address, PTAL~ @liayoo

node/index.js
platfowner: For users, I think we can provide a guide that availableTreeSize (= availableTreeBytes * max_state_tree_size_per_byte) indicates a maximum value of the tree size they are allowed to add.

liayoo: Great, I agree that it could be helpful for users to have the info. I'll try to add the guide when I update the app section in our docs.

liayoo (approved):
LGTM, thanks for handling this issue!

platfowner:

Issue #927, opened 2022-01-28
[BUG] Need to fix apps_state_budget_ratio = 0.495 < 1.0 


platfowner:
We need to fix apps_state_budget_ratio = 0.495 < 1.0

platfowner:
Fixed!

PR #930, opened 2022-01-30
Fix bandwidth budget updates not applying

liayoo:
In getBandwidthBudgets(), the db.blockNumberSnapshot was always set to -2, meaning it always uses the original blockchain params (See https://github.com/ainblockchain/ain-blockchain/blob/master/db/index.js#L588). 
Updated to use the default block number (node.bc.lastBlockNumber()) instead, by setting it to null.

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/927

### DIFF
--- a/tx-pool/index.js
+++ b/tx-pool/index.js
@@ -181,15 +181,15 @@ class TransactionPool {
     return appStakesTotal > 0 ? appsBandwidthBudgetPerBlock * appStake / appStakesTotal : 0;
   }
 
-  getBandwidthBudgets(blockNumber, stateVersion) {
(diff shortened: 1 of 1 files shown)


platfowner (dismissed):
LGTM
Thanks for fixing this!

platfowner (approved):
Let me re-approve it.
Once we see no issues in the testnet deploy, I think we can just move on to the mainnet.

liayoo:
> Let me re-approve it. Once we see no issues in the testnet deploy, I think we can just move on to the mainnet.

Ok, thanks for the quick review! Let me merge & retry the deploy process from the beginning.

PR #937, opened 2022-02-03
Add a script for running tests on cloud (deploy_test_gcp.sh)

platfowner:
Change summary:
- Add a script for running tests on cloud (deploy_test_gcp.sh)

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/762

Run time:
- macbook pro (1.4 GHz Quad-Core Intel Core i5) 
  `yarn run test_integration_node`
  => 9 m 06 s

- c2-standard-4: 
  `bash deploy_test_gcp.sh seo 0 --keep-code test_integration_node`
  => 8 m 59 s

- c2-standard-16: 
  `bash deploy_test_gcp.sh seo 1 --keep-code test_integration_node`
  => 8 m 51 s


liayoo (approved):
LGTM, thanks!

deploy_test_gcp.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/bash
+
+if [[ $# -lt 2 ]]; then
+    printf "Usage: bash deploy_test_gcp.sh <GCP Username> <Instatnce Index> [--setup] [--stop-only] [--keep-code|--no-keep-code]\n"
```
liayoo, line 4: could you add the testing option in usage?

deploy_test_gcp.sh
```diff
@@ -0,0 +1,122 @@
...
+    printf '\n'
+    printf "\n* >> Installing node modules on instance ${TEST_TARGET_ADDR} *********************************************************\n\n"
+    gcloud compute ssh ${TEST_TARGET_ADDR} --command "cd ./ain-blockchain; yarn install --ignore-engines" --project $PROJECT_ID --zone ${TEST_ZONE}
+fi
+
+# stop test servers first
+stop_servers
+
+printf "\n* >> Running tests on instance ${TEST_TARGET_ADDR} *********************************************************\n\n"
+TEST_CMD="cd ./ain-blockchain; yarn run ${TESTING_OPTION}"
```
liayoo, line 120: would it be possible to run the tests with nohup and &?

platfowner: what would be the benefit of running background?

deploy_test_gcp.sh
platfowner: Done, thanks.

deploy_test_gcp.sh
liayoo: It'd mean that we can run the tests and not worry about having the laptop on/having wifi/etc.? It used to take me a couple of hours to run the integration tests, so I thought it'd be good to have :)

platfowner: I see. I think that makes sense. Just added --bg option and --cat-log option for running background tests and cat-logging the test logs. PTAL

liayoo: Thanks! Looks good~

platfowner:

PR #938, opened 2022-02-04
Install node modules before killing jobs in genesis blockchain deploy

platfowner:
Change summary:
- Install node modules before killing jobs in genesis blockchain deploy to minimize the down time


liayoo (approved):
LGTM, thanks!

PR #941, opened 2022-02-07
Support TIMER_FLAG_EARLY_APPLIED_BLOCK_NUMBER env variable

platfowner:
Change summary:
- Support TIMER_FLAG_EARLY_APPLIED_BLOCK_NUMBER env variable
- Fix deploy_test_gcp.sh for integration/dapp test

Background:
- When genesis deploy is necessary, TIMER_FLAG_EARLY_APPLIED_BLOCK_NUMBER env variable allows to apply timer flags of which enabled block numbers are less than the env variable value early (i.e. at block number '2' instead of at 'enabled_block' block number).

### DIFF
--- a/common/common-util.js
+++ b/common/common-util.js
@@ -963,6 +963,10 @@ class CommonUtil {
     if (!CommonUtil.isNumber(enabledBlockNumber)) {
       return null;
     }
+    const earlyAppliedBlockNumber = process.env.TIMER_FLAG_EARLY_APPLIED_BLOCK_NUMBER ? Number(process.env.TIMER_FLAG_EARLY_APPLIED_BLOCK_NUMBER) : null;
+    if (CommonUtil.isNumber(earlyAppliedBlockNumber) && enabledBlockNumber <= earlyAppliedBlockNumber) {
+      return 2;
+    }
     return enabledBlockNumber;
   }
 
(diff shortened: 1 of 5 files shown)


common/common-util.js
```diff
@@ -941,35 +941,61 @@ class CommonUtil {
...
+    }
+    return enabledBlockNumber;
+  }
+
+  static getDisabledBlockNumberFromTimerFlag(timerFlag) {
+    const disabledBlockNumber = timerFlag['disabled_block'];
+    if (!CommonUtil.isNumber(disabledBlockNumber)) {
+      return null;
+    }
+    const earlyAppliedBlockNumber = process.env.TIMER_FLAG_EARLY_APPLIED_BLOCK_NUMBER ? Number(process.env.TIMER_FLAG_EARLY_APPLIED_BLOCK_NUMBER) : null;
```
liayoo, line 978: So if a flag is disabled, the same TIMER_FLAG_EARLY_APPLIED_BLOCK_NUMBER is applied to the disabled block number, which means the flag never gets enabled, right? This makes me realize that applyBandagesForBlockNumber() should check the disabled block number as well.. (let me update in the next pr!)

liayoo (approved):
LGTM, thanks!

common/common-util.js
platfowner: Right. For bandage files, however I think applying one-way change on 'enabled_block' number (i.e., what's currently done) seems to be okay since it's not clear how to revert them?

One thing we can think about is the order of the bandage files. If we want to be more precise, the order of the bandage files need to be sorted by the original 'enabled_block' numbers but it seems to be too much as this is just a backup feature for such a situation that we need to deploy with --no-keep-data option on e.g. testnet spring-summer switches?

liayoo: I see, I was thinking maybe we could not apply the bandage if enabled block number === disabled block number (e.g. TIMER_FLAG_EARLY_APPLIED_BLOCK_NUMBER is set at a later block number), but it does seem tricky to revert in a case where enabled block number < disabled block number. Do you think we could not use disabled block numbers and instead explicitly add reverting changes if the need arises?

platfowner: Yes, with 'disabled_block', we can disable only the feature (with isEnabledTimerFlag()) and the corresponding bandage files can be reverted explicitly with reverting flags (e.g. 'revert_some_feature')?

liayoo: Ah, ok, I see the advantage of keeping disabled_block. Thanks for the explanation!

platfowner: just added a note. thanks!

Issue #879, opened 2022-01-04
Add 'reverted' transaction state


liayoo:
We need to add REVERTED to TransactionStates enum.

liayoo:
Added in #968

PR #968, opened 2022-03-07
Add REVERTED state to TransactionStates.

liayoo:
Related: https://github.com/ainblockchain/ain-blockchain/issues/879

### DIFF
--- a/common/constants.js
+++ b/common/constants.js
@@ -556,9 +556,9 @@ const WriteDbOperations = {
  */
 const TransactionStates = {
   FINALIZED: 'FINALIZED',
-  REVERTED: 'REVERTED',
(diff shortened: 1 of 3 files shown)


common/constants.js
```diff
@@ -556,12 +556,17 @@ const WriteDbOperations = {
  */
 const TransactionStates = {
   FINALIZED: 'FINALIZED',
+  REVERTED: 'REVERTED',
```
platfowner, line 559: question: Is 'reverted' different from 'failed' that it's in the block while 'failed' is not?

could you add some comments for each definition?

liayoo: That's correct! Added some notes. Thanks!

platfowner (approved):
LGTM
Thanks!

liayoo:
Thanks for the review, now merging~

PR #979, opened 2022-03-21
Use 2 for both chain and network ID

minsulee2:
Updates:
- Use 2 for chain ID!
- Rebuild genesis_block.json.gz for testnet-dev


liayoo:
Could you revert the change for the testnet-prod genesis block?

platfowner:
> Could you revert the change for the testnet-prod genesis block?

Right, the file shouldn't be changed, could you check why? @minsulee2

platfowner:
> > Could you revert the change for the testnet-prod genesis block?
> 
> Right, the file shouldn't be changed, could you check why? @minsulee2

Ah, just realized that the genesis block generation requires owner private keys. So for testnet-prod and mainnet-prod, I think you can ignore the errors and simply revert related changes from this PR. @minsulee2  cc: @liayoo

minsulee2:
@platfowner @liayoo I found out that the gz file in testnet-prod was removed when running the script. Thanks for the explain! Mystery resolved!! PTAL!

platfowner (approved):
LGTM
Thanks!

Could you also update the developer's guide?
https://docs.ainetwork.ai/developer-guide/ai-network-design/network-id-and-chain-id

minsulee2:
Thanks for the fast review! Merging now and move on to developer's guide!

Issue #996, opened 2022-03-28
[BUG] Integration test breaks


platfowner:
How to reproduce:
Turn off the disable_tx_receipt_recording timer flag in blockchain-configs/base:
```
  "disable_tx_receipt_recording": {
    "enabled_block": null,
    "has_bandage": false,
    "_note": "Remove this flag or set disabled_block to 2 if you need to run with the tx receipt recording feature"
  }
```
and run
```
yarn run test_integration_node
```

Error messages:
```
yarn run v1.22.10
$ BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/3-nodes ./node_modules/mocha/bin/mocha --timeout 640000 test/integration/node.test.js
[createTimerFlagEnabledBandageMap] Registering bandage files:
[createTimerFlagEnabledBandageMap] [0] Registering /Users/seo/git/comcomai/blockchain/ain-blockchain/db/bandage-files/allow_lower_case_app_names_only.js
[createTimerFlagEnabledBandageMap] [1] Registering /Users/seo/git/comcomai/blockchain/ain-blockchain/db/bandage-files/create_app_config_sanitization.js
[createTimerFlagEnabledBandageMap] [2] Registering /Users/seo/git/comcomai/blockchain/ain-blockchain/db/bandage-files/native_service_path_length_check.js
[createTimerFlagEnabledBandageMap] [3] Registering /Users/seo/git/comcomai/blockchain/ain-blockchain/db/bandage-files/update_max_state_tree_size_per_byte.js
[createTimerFlagEnabledBandageMap] [4] Registering /Users/seo/git/comcomai/blockchain/ain-blockchain/db/bandage-files/staking_balance_total_sum.js
[createTimerFlagEnabledBandageMap] [5] Registering /Users/seo/git/comcomai/blockchain/ain-blockchain/db/bandage-files/update_max_state_tree_size_per_byte2.js
[createTimerFlagEnabledBandageMap] [6] Registering /Users/seo/git/comcomai/blockchain/ain-blockchain/db/bandage-files/update_bandwidth_budget_params.js
[createTimerFlagEnabledBandageMap] [7] Skipping for timer flag: disable_tx_receipt_recording


  Blockchain Node
[createTimerFlagEnabledBandageMap] Registering bandage files:
[createTimerFlagEnabledBandageMap] [0] Registering /Users/seo/git/comcomai/blockchain/ain-blockchain/db/bandage-files/allow_lower_case_app_names_only.js
[...]

PR #999, opened 2022-03-29
Set free limit ratio as 1.0 at integration_node_test

ehgmsdk20:
Changes:
- Set env vars for testing
   * FREE_TX_POOL_SIZE_LIMIT_RATIO: 1.0
   * FREE_TX_POOL_SIZE_LIMIT_RATIO_PER_ACCOUNT: 1.0
- enable float env vars

Related issues:
#996


common/constants.js
```diff
@@ -103,7 +103,7 @@ function setNodeConfigs() {
     if (valFromEnvVar !== undefined) {
       if (CommonUtil.isBool(valFromNodeParams)) {
         NodeConfigs[param] = CommonUtil.convertEnvVarInputToBool(valFromEnvVar);
-      } else if (CommonUtil.isIntegerString(valFromEnvVar)) {
+      } else if (!isNaN(valFromEnvVar)) {
```
platfowner, line 106: It looks like this change could be a little bit risky.
What about introducing a new util function like CommonUtil.isNumberString() that covers floating point numbers using regular expression?
regexp ref: https://github.com/ainblockchain/ain-blockchain/blob/master/db/state-util.js#L128

ehgmsdk20: regex를 이용하여 isFloatString 함수 추가했습니다!

db/rule-util.js
```diff
@@ -16,6 +16,10 @@ class RuleUtil {
     return this.isString(value) && /^-?\d+$/.test(value);
   }
 
+  isFloatString(value) {
```
platfowner, line 19: Could you add test cases?
As this is a primitive function, it'd be safe with thorough test cases.
(please see other cases in rule-util.test.js)

ehgmsdk20: Thanks for your comments! I added test code for isFloatString.

platfowner (approved):
LGTM
Thanks!

PR #1006, opened 2022-04-12
Update some script

minsulee2:
As devnet is now using chain Id as 2, some updates need to be done.
- Update signAndSendTx at tools to dynamically get chain Id
- Update he-health-care tool


minsulee2:
Thanks for the fast review! Merging now

Issue #1016, opened 2022-04-27
Need to extend account registration gas amount for nonce and timestamp 


platfowner:
Currently account registration gas amount is applied only for account balance, and we need to extend it for nonce and timestamp.

platfowner:
Resolved.

PR #1017, opened 2022-04-27
Extended account registration gas amount

platfowner:
Change summary:
- Apply account registration gas amount for nonce and timestamp in addition to the balance 
- Guard extended account registration gas amount with extend_account_registration_gas_amount timer flag
- Make nonce available in transaction execution (not necessary for this PR)
- Invalidate multi-set transactions with empty op_list
- Add more test cases
- Fix some typos in package.json 

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1016


package.json
```diff
@@ -34,8 +34,8 @@
...
     "test_unit_event_handler": "BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/1-node ENABLE_REST_FUNCTION_CALL=true UNSAFE_PRIVATE_KEY=b22c95ffc4a5c096f7d7d0487ba963ce6ac945bdc91c79b64ce209de289bec96 ./node_modules/mocha/bin/mocha --timeout 160000 test/unit/event-handler.test.js",
-    "test_unit_functions": "BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/1-nodeE NABLE_REST_FUNCTION_CALL=true UNSAFE_PRIVATE_KEY=b22c95ffc4a5c096f7d7d0487ba963ce6ac945bdc91c79b64ce209de289bec96 ./node_modules/mocha/bin/mocha --timeout 160000 test/unit/functions.test.js",
-    "test_unit_network_util": "BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/1-nodeE NABLE_REST_FUNCTION_CALL=true UNSAFE_PRIVATE_KEY=b22c95ffc4a5c096f7d7d0487ba963ce6ac945bdc91c79b64ce209de289bec96 ./node_modules/mocha/bin/mocha --timeout 160000 test/unit/network-util.test.js",
+    "test_unit_functions": "BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/1-node ENABLE_REST_FUNCTION_CALL=true UNSAFE_PRIVATE_KEY=b22c95ffc4a5c096f7d7d0487ba963ce6ac945bdc91c79b64ce209de289bec96 ./node_modules/mocha/bin/mocha --timeout 160000 test/unit/functions.test.js",
```
liayoo, line 37: 👀 thanks for fixing them!

common/path-util.js
```diff
@@ -7,16 +7,20 @@ const RuleUtil = require('../db/rule-util');
...
+  }
+
   static getAccountBalancePath(address) {
-    return CommonUtil.formatPath([PredefinedDbPaths.ACCOUNTS, address, PredefinedDbPaths.BALANCE]);
+    return `${PathUtil.getAccountPath(address)}/${PredefinedDbPaths.BALANCE}`;
   }
 
   static getAccountNoncePath(address) {
-    return CommonUtil.formatPath([PredefinedDbPaths.ACCOUNTS, address, PredefinedDbPaths.ACCOUNTS_NONCE]);
+    return `${PathUtil.getAccountPath(address)}/${PredefinedDbPaths.ACCOUNTS_NONCE}`;
```
liayoo, line 19: ditto

common/path-util.js
```diff
@@ -7,16 +7,20 @@ const RuleUtil = require('../db/rule-util');
...
   }
 
   static getAccountTimestampPath(address) {
-    return CommonUtil.formatPath([PredefinedDbPaths.ACCOUNTS, address, PredefinedDbPaths.ACCOUNTS_TIMESTAMP]);
+    return `${PathUtil.getAccountPath(address)}/${PredefinedDbPaths.ACCOUNTS_TIMESTAMP}`;
```
liayoo, line 23: ditto

db/index.js
```diff
@@ -1316,13 +1328,29 @@ class DB {
...
+          this.executeMultiSetOperation(op.op_list, auth, nonce, timestamp, tx, blockNumber, blockTime));
     } else {
       Object.assign(
-          result, this.executeSingleSetOperation(op, auth, timestamp, tx, blockNumber, blockTime));
+          result, this.executeSingleSetOperation(op, auth, nonce, timestamp, tx, blockNumber, blockTime));
+    }
+    if (isEnabledTimerFlag('extend_account_registration_gas_amount', blockNumber)) {
+      // Apply account registration gas amount for nonce and timestamp.
+      const isNonExistingAccount = this.checkIfNonExistingAccount(tx, auth);
+      if (wasNonExistingAccount && isNonExistingAccount && nonce !== -1) {
```
liayoo, line 1346: 🙌

db/index.js
```diff
@@ -1265,7 +1266,7 @@ class DB {
     return result;
   }
 
-  executeMultiSetOperation(opList, auth, timestamp, tx, blockNumber, blockTime) {
+  executeMultiSetOperation(opList, auth, nonce, timestamp, tx, blockNumber, blockTime) {
```
liayoo, line 1269: question) are the nonces added to executeMultiSetOperation() and executeSingleSetOperation() in order to match the function signature of executeOperation()?

common/path-util.js
```diff
@@ -7,16 +7,20 @@ const RuleUtil = require('../db/rule-util');
 const ruleUtil = new RuleUtil();
 
 class PathUtil {
+  static getAccountPath(address) {
+    return CommonUtil.formatPath([PredefinedDbPaths.ACCOUNTS, address]);
+  }
+
   static getAccountBalancePath(address) {
-    return CommonUtil.formatPath([PredefinedDbPaths.ACCOUNTS, address, PredefinedDbPaths.BALANCE]);
+    return `${PathUtil.getAccountPath(address)}/${PredefinedDbPaths.BALANCE}`;
```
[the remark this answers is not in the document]
platfowner: At first I implemented so, but realized it's not efficient as the input paths are converted to arrays and then converted to a string. So, in the case the input format is clear I think just using the string concatenation is faster and simpler.

db/index.js
platfowner: Ah, at first I added it to use it inside the function call but realized that the account registration gas amount from the nonce or timestamp needs to be computed in executeOperation() with the following reasons:
- Nonce / timestamp update is done in executeOperation() per transaction
- One transaction can have multiple operations (e.g. SET operation)

As a result the change is not necessary in this PR, but worth to keep the change for further use cases in the future?

platfowner:
Let me leave some quick answers to the questions and suggestions. PTAL~ @liayoo

common/path-util.js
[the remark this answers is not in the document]
liayoo: I see! Sounds good to me. Thanks for the explanation.

db/index.js
liayoo: Right. Makes sense & I agree that we can keep the nonce args for future use.

test/integration/sharding.test.js
```diff
@@ -491,7 +491,7 @@ describe('Sharding', () => {
           [server2Addr]: true,
           [server3Addr]: true,
         }
-      });
+      }, 200);
```
liayoo, line 494: 👍

tx-pool/transaction.js
```diff
@@ -242,6 +242,10 @@ class Transaction {
       logger.info(`Transaction body is in a non-standard format:\n${diffLines}\n`);
       return false;
     }
+    if (sanitized.operation.type === WriteDbOperations.SET &&
```
liayoo, line 245: Huh, I thought we were already checking this somewhere, but I guess not.. Thanks for adding it!

liayoo (approved):
LGTM, thanks!

platfowner:
Thanks for the review, now merging.. @liayoo

PR #1018, opened 2022-04-28
Turn on --bg option by default in deploy_test_gcp.sh

platfowner:
Change summary:
- Turn on --bg option by default in deploy_test_gcp.sh
- Add --fg option instead

### DIFF
--- a/deploy_test_gcp.sh
+++ b/deploy_test_gcp.sh
@@ -152,7 +152,7 @@ function deploy_test() {
 }
 
 if [[ $INSTANCE_INDEX = "all" ]]; then
-    if [[ $FOREGROUND_OPTION = "--fg" ]] || [[ $STOP_ONLY_OPTION = "--stop-only" ]]; then
+    if [[ $FOREGROUND_OPTION = "--fg" ]]; then
         # serialized function calls
         deploy_test 0 test_unit
         deploy_test 1 test_integration_function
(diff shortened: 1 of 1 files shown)


deploy_test_gcp.sh
```diff
@@ -141,30 +141,18 @@ function deploy_test() {
...
-        deploy_test 2 test_integration_node &
-        deploy_test 3 test_integration_sharding &
-        deploy_test 4 test_integration_blockchain &
-        deploy_test 5 test_integration_consensus &
-        deploy_test 6 test_integration_dapp &
-        deploy_test 7 test_integration_he_protocol &
-        deploy_test 8 test_integration_he_sharding &
-        deploy_test 9 test_integration_event_handler
-    else
+    if [[ $FOREGROUND_OPTION = "--fg" ]] || [[ $STOP_ONLY_OPTION = "--stop-only" ]]; then
```
kriii, line 155: This logic changed parallelized into serialized when the stop-only option turned on, is it intended? (It seems that all two choices are workable, but I asked if there could be some details in this.)

kriii (approved):
LGTM with a comment!
Thanks!

deploy_test_gcp.sh
platfowner: Ah, --stop-only needs to be removed, and just done. Thanks!

platfowner:
Let me merge this PR now. Feel free to add any further comments.

Issue #1020, opened 2022-04-29
Need to add app_creation_gas_amount for app creation txs


platfowner:
We need to add app_creation_gas_amount for app creation txs.

platfowner:
Resolved.

PR #1021, opened 2022-04-29
Add app_creation_gas_amount for app creation txs

platfowner:
Change summary:
- Add app_creation_gas_amount for app creation txs
- Set missing gas_price for chatbot tool script

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1020

### DIFF
--- a/db/functions.js
+++ b/db/functions.js
@@ -702,7 +702,7 @@ class Functions {
     const accountPath = CommonUtil.isServAcntName(addrOrServAcnt) ?
         PathUtil.getServiceAccountPathFromAccountName(addrOrServAcnt) :
         PathUtil.getAccountPath(addrOrServAcnt);
-    const curAccountValue = this.db.getValue(accountPath);
+    const curAccountValue = this.db.getValue(accountPath, { isShallow: true });
     return curAccountValue === null;
   }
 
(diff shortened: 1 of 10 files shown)


db/functions.js
```diff
@@ -789,6 +789,10 @@ class Functions {
     if (errorCode) {
       return this.returnFuncResult(context, errorCode);
     }
+    let extraGasAmount = 0;
+    if (isEnabledTimerFlag('add_app_creation_gas_amount', context.blockNumber)) {
```
liayoo, line 799: Is it intended to be fined for failed txs as well? If not, how about moving this to L823?

db/functions.js
```diff
@@ -775,6 +775,12 @@ class Functions {
     return { sanitizedVal, errorCode: null };
   }
 
+  isNonExistingApp(appName) {
+    const appPath = PathUtil.getAccountPath(appName);
```
liayoo, line 779: How about using getManageAppConfigPath()?

db/functions.js
platfowner: As you can see, extra gas amount is added in the success case only (line 831 `this.returnFuncResult(context, FunctionResultCode.SUCCESS, extraGasAmount)`), but the extra gas amount is computed here before core business logic including rule, owner, and config settings.

db/functions.js
platfowner: Oops, just fixed. thanks!

platfowner:
Just addressed the comments, PTAL~ @liayoo

db/functions.js
liayoo: Ah, I see. Thanks for the explanation!

liayoo (approved):
LGTM, thanks!

platfowner:

PR #1024, opened 2022-05-04
Reorg local blockchain run scripts

platfowner:
Change summary:
- Reorg local blockchain run scripts
- Add `ENABLE_EVENT_HANDLER=true \` to local blockchain node 2 for ain-js event manager testing


start_local_blockchain.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+printf "\n[[[[[ start_local_blockchain.sh ]]]]]\n\n"
```
liayoo, line 3: Could you leave a note at the top about the event handler node (how it's different from start_servers_local.sh)?

liayoo (approved):
LGTM with a comment. Thanks!

start_local_blockchain.sh
platfowner: Just done, thanks!

platfowner:
Thanks for the fast review, now merging.. @liayoo

PR #1025, opened 2022-05-04
Add a note for testing ainblockchain/ain-js module

platfowner:


start_local_blockchain.sh
```diff
@@ -2,6 +2,9 @@
 #
 # A script to start a local blockchain.
 #
+# Note that ENABLE_EVENT_HANDLER=true is set for node 2 for testing
+# the event manager modules of ainblockchain/ain-js.
```
liayoo, line 6: minor) [@ainblockchain/ain-js](https://www.npmjs.com/package/@ainblockchain/ain-js) is the official npm package name~

liayoo (approved):
LGTM with a comment, thanks!

start_local_blockchain.sh
platfowner: Just fixed, thanks!

PR #1027, opened 2022-05-06
Add privateKeyToAccount.js tool script

platfowner:
Change summary:
- Add privateKeyToAccount.js tool script
- Let privateKeyToKeystore.js tool output keystore string 
- Add package script commands for starting local blockchain


liayoo (approved):
LGTM, thanks!

PR #1031, opened 2022-05-11
Feature/csh/value changed from

cshcomcom:
## Description
- Now we can select source of VALUE_CHANGED event (`BLOCK`, `USER`)

### DIFF
--- a/block-pool/index.js
+++ b/block-pool/index.js
@@ -566,7 +566,7 @@ class BlockPool {
     let tallied = 0;
     for (const vote of lastVotes) {
       if (CommonUtil.isFailedTx(tempDb.executeTransaction(
-          Transaction.toExecutable(vote, chainId), true, true, blockNumber, blockTime))) {
+          Transaction.toExecutable(vote, chainId), true, true, blockNumber, blockTime, null))) {
         logger.debug(`[${LOG_HEADER}] failed to execute last vote: ${JSON.stringify(vote, null, 2)}`);
       } else {
         tallied += _get(lastBlock.validators, `${vote.address}.stake`, 0);
@@ -591,7 +591,7 @@ class BlockPool {
    * @param {DB} baseDb The DB instance should be the base of where evidence votes should be executed on.
    * @returns { offenses, evidence }
    */
-  getOffensesAndEvidence(validators, recordedInvalidBlockHashSet, blockNumber, blockTime, baseDb) {
+  getOffensesAndEvidence(validators, recordedInvalidBlockHashSet, blockNumber, blockTime, baseDb, eventSource) {
     const LOG_HEADER = 'getOffensesAndEvidence';
     const totalAtStake = ConsensusUtil.getTotalAtStake(validators);
     const baseBlockNumber = baseDb.blockNumberSnapshot;
@@ -620,7 +620,8 @@ class BlockPool {
         const stake = _get(validators, `${vote.address}.stake`, 0);
         if (stake > 0) {
           const res = baseDb.executeTransaction(
-              Transaction.toExecutable(vote, chainId), true, true, blockNumber, blockTime);
+              Transaction.toExecutable(vote, chainId), true, true,
+              blockNumber, blockTime, eventSource);
           if (CommonUtil.isFailedTx(res)) {
             logger.debug(`[${LOG_HEADER}] Failed to execute evidence vote:\n${JSON.stringify(vote, null, 2)}\n${JSON.stringify(res, null, 2)})`);
           } else {
--- a/block-pool/index.js
+++ b/block-pool/index.js
@@ -620,8 +620,7 @@ class BlockPool {
         const stake = _get(validators, `${vote.address}.stake`, 0);
         if (stake > 0) {
           const res = baseDb.executeTransaction(
-              Transaction.toExecutable(vote, chainId), true, true,
-              blockNumber, blockTime, eventSource);
+              Transaction.toExecutable(vote, chainId), true, true, blockNumber, blockTime, eventSource);
           if (CommonUtil.isFailedTx(res)) {
             logger.debug(`[${LOG_HEADER}] Failed to execute evidence vote:\n${JSON.stringify(vote, null, 2)}\n${JSON.stringify(res, null, 2)})`);
           } else {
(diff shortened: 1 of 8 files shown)


block-pool/index.js
```diff
@@ -620,7 +620,8 @@ class BlockPool {
         const stake = _get(validators, `${vote.address}.stake`, 0);
         if (stake > 0) {
           const res = baseDb.executeTransaction(
-              Transaction.toExecutable(vote, chainId), true, true, blockNumber, blockTime);
+              Transaction.toExecutable(vote, chainId), true, true,
+              blockNumber, blockTime, eventSource);
```
minsulee2, line 624: Quadruple spaces?

minsulee2: Oh, never mind. I got confused.

minsulee2 (commented):
Before moving on, I wonder why the source matters?

block-pool/index.js
```diff
@@ -620,7 +620,8 @@ class BlockPool {
         const stake = _get(validators, `${vote.address}.stake`, 0);
         if (stake > 0) {
           const res = baseDb.executeTransaction(
-              Transaction.toExecutable(vote, chainId), true, true, blockNumber, blockTime);
+              Transaction.toExecutable(vote, chainId), true, true,
```
platfowner, line 623: minor) this line seems to be too short after the line wrapping?

node/index.js
```diff
@@ -916,7 +917,7 @@ class BlockchainNode {
    * after removing the invalid transactions from the tx pool.
    */
   executeAndGetValidTransactions(
-      longestNotarizedChain, blockNumber, blockTime, baseDb, isExecutionOnly = false) {
+      longestNotarizedChain, blockNumber, blockTime, baseDb, isExecutionOnly = false, eventSource) {
```
platfowner, line 920: if an argument with a default value is followed by another argument without default values, the default value of the former argument becomes useless.
could you remove the default value?

event-handler/index.js
```diff
@@ -89,6 +92,11 @@ class EventHandler {
         continue;
       }
 
+      const eventSourceConfig = _.get(eventFilter, 'config.event_source', null);
```
platfowner, line 95: minor) eventSourceConfig sound like an object. what about something like expectedEventSource?

block-pool/index.js
```diff
@@ -566,7 +566,7 @@ class BlockPool {
     let tallied = 0;
     for (const vote of lastVotes) {
       if (CommonUtil.isFailedTx(tempDb.executeTransaction(
-          Transaction.toExecutable(vote, chainId), true, true, blockNumber, blockTime))) {
+          Transaction.toExecutable(vote, chainId), true, true, blockNumber, blockTime, null))) {
```
platfowner, line 569: null value skippable?

consensus/index.js
```diff
@@ -771,7 +773,8 @@ class Consensus {
       });
     }
     // Try executing the proposal tx on the proposal block's db state
-    const proposalTxExecRes = tempDb.executeTransaction(executableTx, true, false, number, blockTime);
+    const proposalTxExecRes = tempDb.executeTransaction(executableTx, true, false,
+        number, blockTime, null);
```
platfowner, line 777: null value skippable?

consensus/index.js
```diff
@@ -928,7 +931,7 @@ class Consensus {
           `[${LOG_HEADER}] Failed to create a temp state snapshot for vote ${JSON.stringify(executableTx)}`);
       return false;
     }
-    const voteTxRes = tempDb.executeTransaction(executableTx, true, false, prevBlockNumber + 1, prevBlockTime);
+    const voteTxRes = tempDb.executeTransaction(executableTx, true, false, prevBlockNumber + 1, prevBlockTime, null);
```
platfowner, line 934: null value skippable?

node/index.js
platfowner: Ah, i think it'd be better to set a default value (null) of eventSource.

consensus/index.js
```diff
@@ -414,10 +415,10 @@ class Consensus {
     }
     const totalAtStake = ConsensusUtil.getTotalAtStake(validators);
     const { offenses, evidence } = this.node.bp.getOffensesAndEvidence(
-        validators, recordedInvalidBlockHashSet, blockNumber, blockTime, tempDb);
+        validators, recordedInvalidBlockHashSet, blockNumber, blockTime, tempDb, null);
     const { transactions, receipts, gasAmountTotal, gasCostTotal } =
         this.node.executeAndGetValidTransactions(
-            longestNotarizedChain, blockNumber, blockTime, tempDb);
+            longestNotarizedChain, blockNumber, blockTime, tempDb, false, null);
```
platfowner, line 421: if executeAndGetValidTransactions() has a default value for eventSource, the last two params can be skipped.

cshcomcom:
> Before moving on, I wonder why the source matters?
@minsulee2 
Without this feature, users would not know whether VALUE_CHANGED event was due to the execution of transactions in block or a transaction received from other users. To solve this inconvenience, this feature has been added.

block-pool/index.js
cshcomcom: I agree. I change it to 1 line.

node/index.js
cshcomcom: Looks good. Now we can omit parameters when calling from other places.

event-handler/index.js
cshcomcom: Looks cool! Done

block-pool/index.js
cshcomcom: Done!

consensus/index.js
cshcomcom: Done!

consensus/index.js
cshcomcom: Done!

consensus/index.js
cshcomcom: I make `executeAndGetValidTransactions`'s last 2 parameters to skippable and skip them! Done.

cshcomcom:
All comments are addressed! PTAL @platfowner

cshcomcom:
Thanks for the review! Test completed and now merging!

Issue #1045, opened 2022-05-24
[BUG] Need to improve write rule validation logic


platfowner:

platfowner:
Fixed.

PR #1046, opened 2022-05-24
Add write rule keyword token blacklist

platfowner:
Change summary:
- Add write rule keyword token blacklist
- Enrich write rule punc token blacklist

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1045


platfowner:
Thanks for the review! Now merging.. @liayoo

Issue #600, opened 2021-10-05
Need get APIs for 'more items' (step 1: child node pagination)


platfowner:
We need get APIs for 'more items' (child pagination).

platfowner:
Step 1 resolved.

PR #1048, opened 2022-05-25
Add child node pagination with is_partial option

platfowner:
Change summary:
- Add child node pagination with is_partial = true option (By this, we support chained retrieval of children)
- Set fromApi=true for GET operations as well
- Let resp sibling number limit be skipped with is_partial option
- Let resp bytes limit be skipped with is_partial or is_shallow options

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/600

### DIFF
--- a/db/index.js
+++ b/db/index.js
@@ -2270,6 +2270,7 @@ class DB {
     return funcs;
   }
 
+  // TODO(platfowner): Consider optimizing subtree function retrieval logic.
   getSubtreeFunctions(funcNode) {
     return this.getSubtreeFunctionsRecursive(0, funcNode);
   }
@@ -2417,6 +2418,7 @@ class DB {
     return rules;
   }
 
+  // TODO(platfowner): Consider optimizing subtree rule retrieval logic.
   getSubtreeRules(ruleNode, ruleProp) {
     return this.getSubtreeRulesRecursive(0, ruleNode, ruleProp);
   }
@@ -2682,6 +2684,7 @@ class DB {
     return owners;
   }
 
+  // TODO(platfowner): Consider optimizing subtree owner retrieval logic.
   getSubtreeOwners(ownerNode) {
     return this.getSubtreeOwnersRecursive(0, ownerNode);
   }
--- a/db/index.js
+++ b/db/index.js
@@ -720,7 +720,7 @@ class DB {
 
   checkRespTreeLimitsForEvalOrMatch(rootLabel, localPath, options) {
     if (options && options.fromApi) {
-    const targetStateRoot = options.isFinal ? this.stateManager.getFinalRoot() : this.stateRoot;
+      const targetStateRoot = options.isFinal ? this.stateManager.getFinalRoot() : this.stateRoot;
       const fullPath = DB.getFullPath(localPath, rootLabel);
       const stateNode = DB.getRefForReadingFromStateRoot(targetStateRoot, fullPath);
       if (stateNode !== null) {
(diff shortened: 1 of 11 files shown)


db/radix-node.js
```diff
@@ -336,18 +336,65 @@ class RadixNode {
...
 
-  getChildStateNodeList() {
+  static compareRadixLabelWithPrefix(label, prefix) {
+    if (_.startsWith(prefix, label)) {
+      return 0;
+    }
+    return label.localeCompare(prefix);
+  }
+
+  getChildStateNodeListWithEndLabel(maxListSize = null, lastEndLabel = null) {
```
minsulee2, line 346: Q) I get the point that labels are localeCompared to pagenate tho, what is the label btw? I mean how does it originally set up?

test/unit/db.test.js
```diff
@@ -339,6 +339,74 @@ describe("DB operations", () => {
         })
       });
 
+      it('getValue to retrieve value near top of database with isPartial = true and lastEndLabel', () => {
+        assert.deepEqual(node.db.getValue('/apps/test', { isPartial: true }), {
```
minsulee2, line 343: Can we get a case that isPartial is false? or do we already have one ?

db/radix-tree.js
```diff
@@ -377,17 +378,26 @@ class RadixTree {
...
+    const labelList = nodesWithEndLabel.list.map(entry => entry.getLabel());
+    return {
+      list: labelList,
+      endLabel: nodesWithEndLabel.endLabel,
+    };
   }
 
-  getChildStateNodes() {
-    return this.root.getChildStateNodeList().sort((a, b) => a.serial - b.serial)
+  getChildStateNodesWithEndLabel(isPartial = false, lastEndLabel = null) {
```
minsulee2, line 390: Do we now get all if we do not set isPartial as true?

db/radix-node.js
```diff
@@ -336,18 +336,65 @@ class RadixNode {
...
+      };
+    }
+    if (this.hasChildStateNode() && !CommonUtil.isString(lastEndLabel)) {
       stateNodeList.push({
         serial: this.getSerial(),
         stateNode: this.getChildStateNode()
       });
+      endLabel = this.getLabel();
+    }
+    if (CommonUtil.isNumber(maxListSize) && stateNodeList.length === maxListSize) {
```
liayoo, line 362: how about `stateNodeList.length >= maxListSize`?

db/radix-node.js
```diff
@@ -336,18 +336,65 @@ class RadixNode {
...
+          maxListSize - stateNodeList.length : null;
+      const stateNodeListFromChild = child.getChildStateNodeListWithEndLabel(
+          maxListSizeForChild,
+          labelComparison > 0 ? null : lastEndLabelForChild);
+      if (stateNodeListFromChild.list.length > 0) {
+        stateNodeList.push(...stateNodeListFromChild.list);
+        endLabel = stateNodeListFromChild.endLabel !== null ?
+            this.getLabel() + stateNodeListFromChild.endLabel : null;
+      }
+      if (CommonUtil.isNumber(maxListSize) && stateNodeList.length === maxListSize) {
```
liayoo, line 387: ditto

db/index.js
```diff
@@ -753,19 +753,19 @@ class DB {
     const resultList = [];
     for (const op of opList) {
       if (op.type === undefined || op.type === ReadDbOperations.GET_VALUE) {
-        resultList.push(this.getValue(op.ref, CommonUtil.toGetOptions(op)));
+        resultList.push(this.getValue(op.ref, CommonUtil.toGetOptions(op, true)));
```
liayoo, line 756: Thanks for adding these~

db/radix-node.js
platfowner: done. thanks!

db/radix-node.js
platfowner: done. thanks!

db/radix-node.js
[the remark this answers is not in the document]
platfowner: Just added more test cases (see 'getChildStateNodeListWithLabel with non-existing lastEndLabel' in radix-node.test.js). PTAL

platfowner: As explained in the AIM doc (AIM-029), the labels are the radix label in the radix tree (for efficiency all child state nodes of a state node are leaf nodes of a radix tree). So, by giving the last end label of the previous query, we can specify the starting point in the radix tree for the next query.

test/unit/db.test.js
platfowner: The default value of isPartial is 'false' and the test cases without any the option correspond to the isPartial = false case.

db/radix-tree.js
platfowner: yes, isPartial = false means 'full' fetch in terms of the child branching factor.

platfowner:
All comments addressed, PTAL~ @liayoo @minsulee2

db/radix-node.js
minsulee2: Oh, I see. I asked this before announcing AIM-029 :) Now, I totally get it. Thanks for the long explain !!

db/radix-tree.js
minsulee2: Got this. I should take it into account when it comes to Insight.

minsulee2 (approved):
LGTM! waiting for @liayoo 's call :)

db/radix-node.js
```diff
@@ -336,18 +336,65 @@ class RadixNode {
...
+      if (labelComparison < 0) {
+        continue;
+      }
+      const maxListSizeForChild = CommonUtil.isNumber(maxListSize) ?
+          maxListSize - stateNodeList.length : null;
+      const stateNodeListFromChild = child.getChildStateNodeListWithEndLabel(
+          maxListSizeForChild,
+          labelComparison > 0 ? null : lastEndLabelForChild);
+      if (stateNodeListFromChild.list.length > 0) {
+        stateNodeList.push(...stateNodeListFromChild.list);
```
liayoo, line 383: question) If this list is concatenated, how does the user know how to reconstruct the full tree from this list? Could you elaborate on the merging process described in AIM-029?

platfowner: That's a good questions. :)
Users don't need to reconstruct the radix tree as what they want is the full list of child state nodes. The order of the child state nodes, however is unknown for the users, which is saved as 'serial' values. For supporting partial results merging in the correct order, I have a plan to return the serial values as well. Just added a todo for that.

liayoo: Ah, I get it. Thanks for the explanation & the todo!!

liayoo (approved):
LGTM, thanks!

db/radix-node.js
platfowner: Ah, notably is_partial = true option also returns shallow results like is_shallow = true option, fyi. @liayoo

platfowner:
Thanks for the review! Now merging.. @minsulee2 @liayoo

PR #1056, opened 2022-06-03
Load json file for topology

minsulee2:
Some updates:
- Enable loading json file to draw topology graph
- Add example json file

### DIFF
--- a/tracker-server/example.json
+++ b/tracker-server/example.json
@@ -1,16 +1,16 @@
 {
   "nodes": [
     {
-      "address": "t"
+      "address": "0x000"
     },
     {
-      "address": "e"
+      "address": "0x001"
     },
     {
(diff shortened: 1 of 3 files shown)


tracker-server/example.json
```diff
@@ -0,0 +1,49 @@
+{
+  "nodes": [
+    {
+      "address": "t"
```
platfowner, line 4: minor) what about more friendly address samples e.g. "0x001", "0xADDR_001", ...?

platfowner (approved):
Thanks!

tracker-server/example.json
minsulee2: Updated!

minsulee2:
Thanks for the fast review! Merging now !

PR #1060, opened 2022-06-09
Remove disconnectable peer candidates

minsulee2:
Enable to remove disconnectable peer candidates. Updated the way how to connect peers one another.
1. Pick some peers on the peerCandidate list
2. Try to gradually connect the peers one by one
3. Remove peer info if connection fails, keep otherwise.
4. Run from 1 to 3 til the connection is full.


p2p/index.js
```diff
@@ -448,6 +449,30 @@ class P2pClient {
     }, NodeConfigs.PEER_CANDIDATES_CONNECTION_INTERVAL_MS);
   }
 
+  async removeOldPeerCandidates() {
+    const LOG_HEADER = 'removeOldPeerCandidates';
+    for (let entry of this.peerCandidates.entries()) {
```
platfowner, line 454: what about using .keys() instead?
`for (const jsonRpcEndpoint of this.peerCandidates.keys()) {`

platfowner:
what's the benefit of removing

p2p/index.js
```diff
@@ -448,6 +449,30 @@ class P2pClient {
...
+        }
+      } catch (error) {
+        logger.error(`[${LOG_HEADER}] ${error}`);
+        continue;
+      }
+    }
+  }
+
+  setIntervalForOldPeerCandidatesRemoval() {
+    this.intervalOldPeerCandidatesRemoval = setInterval(async () => {
```
platfowner, line 471: this generates traffic to all peer candidates. i think we need to think about the following issues:
- the traffic grows in square of the network size. e.g. 100 -> 10,1000, 1000 -> 1,000,000 per hour in worst case

could you consider improving discoverPeerWithGuardingFlag() instead of generating extra traffic?

platfowner:
oh, my mistake.

p2p/index.js
platfowner: Discussed offline:
- assignRandomPeerCandidate() ([code](https://github.com/ainblockchain/ain-blockchain/blob/master/p2p/index.js#L338))
    - Return a merged array of shuffled candidates
- discoverPeerWithGuardingFlag() ([code](https://github.com/ainblockchain/ain-blockchain/blob/master/p2p/index.js#L338)) 
  - Try to make a peer connection with the peer candidates one by one
  - If fails, remove the peer candidates from the peer candidate pool 
  - Rename: -> discoverNewPeer()

minsulee2:
PTAL! All addressed!

blockchain-configs/afan-shard/node_params.json
```diff
@@ -49,6 +49,7 @@
   "MAX_NUM_SNAPSHOTS": 10,
   "OLD_CHAIN_SEGMENT_LENGTH": 20,
   "OLD_CHAIN_SEGMENT_SLEEP_MS": 10000,
+  "OLD_PEER_CANDIDATES_REMOVAL_INTERVAL_MS": 3600000,
```
platfowner, line 52: not used any more?

p2p/index.js
```diff
@@ -330,44 +311,101 @@ class P2pClient {
...
       const notQueriedCandidateEntries = [...this.peerCandidates.entries()].filter(([, value]) => {
         // NOTE(minsulee2): this gets stuck if the never queried node gets offline. To avoid this,
         // the node which queried more than 5 minutes ago can also be considered as notQueried.
         return value.queriedAt === null ? true :
             Date.now() - value.queriedAt > NodeConfigs.PEER_CANDIDATE_RETRY_THRESHOLD_MS;
       });
-      if (notQueriedCandidateEntries.length > 0) {
-        return _.shuffle(notQueriedCandidateEntries)[0][0];
+      const notQueriedCandidatesObject = Object.fromEntries(notQueriedCandidateEntries);
+      const notQueriedCandidates = Object.keys(notQueriedCandidatesObject);
```
platfowner, line 330: can be simplified by using .map() like `const notQueriedCandidateEntries = [...this.peerCandidates.entries()].filter(...).map(...);`?

p2p/index.js
```diff
@@ -330,44 +311,101 @@ class P2pClient {
...
+        // 5. Semaphore off
         this.isConnectingToPeerCandidates = false;
       }
     }
   }
 
+  async discoverNewPeers() {
+    const maxNumberOfNewPeers = this.getMaxNumberOfNewPeers();
+    const nextPeerCandidateJsonRpcUrlList = this.assignRandomPeerCandidates();
+    const availableJsonRpcUrlList = nextPeerCandidateJsonRpcUrlList.slice(0, maxNumberOfNewPeers);
```
platfowner, line 400: need to apply `maxNumberOfNewPeers` to the for-loop below instead after making discoverNewPeer() return success or failure?

p2p/index.js
```diff
@@ -330,44 +311,101 @@ class P2pClient {
...
+      if (url !== myJsonRpcUrl && !this.peerCandidates.has(url) && this.isValidJsonRpcUrl(url) &&
+          P2pUtil.checkPeerWhitelist(address)) {
+        this.setPeerCandidate(url, address, null);
+      }
+    });
+  }
+
+  async discoverNewPeer(jsonRpcUrl) {
+    const LOG_HEADER = 'discoverNewPeer';
     if (!this.isConnectingToPeerCandidates) {
```
platfowner, line 376: now need to move this guarding to one level upper to discoverNewPeers()?

p2p/index.js
```diff
@@ -330,44 +311,101 @@ class P2pClient {
...
+    const peerCandidateJsonRpcUrlList = _.get(peerCandidateInfo, 'peerCandidateJsonRpcUrlList', []);
+    Object.entries(peerCandidateJsonRpcUrlList).forEach(([address, url]) => {
+      if (url !== myJsonRpcUrl && !this.peerCandidates.has(url) && this.isValidJsonRpcUrl(url) &&
+          P2pUtil.checkPeerWhitelist(address)) {
+        this.setPeerCandidate(url, address, null);
+      }
+    });
+  }
+
+  async discoverNewPeer(jsonRpcUrl) {
```
platfowner, line 374: as we're already using a similar name discoverNewPeers() could we use different name like tryNewPeerConnection() or something?

p2p/index.js
```diff
@@ -330,44 +311,101 @@ class P2pClient {
...
+      throw new Error(`Invalid peerCandidateInfo from peer candidate url (${peerCandidateInfo}).`);
+    }
+    return peerCandidateInfo;
+  }
+
+  setPeerCandidateFromPeerCandidateInfo(peerCandidateInfo) {
+    const jsonRpcUrl = _.get(peerCandidateInfo, 'networkStatus.urls.jsonRpc.url');
+    const address = _.get(peerCandidateInfo, 'address');
+    if (!this.isValidJsonRpcUrl(jsonRpcUrl) || !CommonUtil.isValAddr(address)) {
+      throw new Error('peerCandidateInfo is not correctly set.' +
```
platfowner, line 360: minor) what about just '... is invalid.'?

p2p/index.js
```diff
@@ -330,44 +311,101 @@ class P2pClient {
...
+          `(${JSON.stringify(peerCandidateInfo)})`);
+    }
+    this.setPeerCandidate(jsonRpcUrl, address, Date.now());
```
platfowner, line 363: minor) it looks like this function just updates 'queriedAt' time. what about just 'updatePeerCandidateInfo()' or more specifically 'updatePeerCandidateQueryTime()' or something like that?

p2p/index.js
```diff
@@ -1168,56 +1204,26 @@ class P2pClient {
    * @param {string} peerCandidateJsonRpcUrl should be something like
    * http(s)://xxx.xxx.xxx.xxx/json-rpc
    */
-  async connectWithPeerCandidateUrl(peerCandidateJsonRpcUrl) {
+  async connectWithPeerCandidateUrl(peerCandidateInfo) {
```
platfowner, line 1207: now need to be updated to 'connectWithPeerCandidateInfo'?

p2p/index.js
```diff
@@ -1168,56 +1204,26 @@ class P2pClient {
...
-    const jsonRpcUrlFromResp = _.get(peerCandidateInfo, 'networkStatus.urls.jsonRpc.url');
     const address = _.get(peerCandidateInfo, 'address');
-    if (!jsonRpcUrlFromResp) {
-      logger.error(`Invalid peer candidate json rpc url from peer candidate url ` +
-          `(${peerCandidateJsonRpcUrl}).`);
-      return;
-    }
-    if (jsonRpcUrlFromResp !== myJsonRpcUrl) {
-      this.setPeerCandidate(jsonRpcUrlFromResp, address, Date.now());
+    if (!address) {
```
platfowner, line 1211: what about applying stronger conditions e.g. CommonUtil.isCksumAddr()?

p2p/index.js
```diff
@@ -1168,56 +1204,26 @@ class P2pClient {
...
+      throw new Error(`Wrong peerCandidateInfo(${JSON.stringify(peerCandidateInfo)})`);
```
platfowner, line 1212: what about using a more specific message?

p2p/index.js
```diff
@@ -330,44 +311,101 @@ class P2pClient {
...
+    const resp = await sendGetRequest(
+        peerCandidateJsonRpcUrl, JSON_RPC_METHODS.P2P_GET_PEER_CANDIDATE_INFO, { });
+    const peerCandidateInfo = _.get(resp, 'data.result.result');
+    if (!CommonUtil.isDict(peerCandidateInfo)) {
+      throw new Error(`Invalid peerCandidateInfo from peer candidate url (${peerCandidateInfo}).`);
+    }
+    return peerCandidateInfo;
+  }
+
+  setPeerCandidateFromPeerCandidateInfo(peerCandidateInfo) {
```
platfowner, line 356: minor) what about 'populatePeerCandidatesFromPeerCandidateInfo()' or something?

blockchain-configs/afan-shard/node_params.json
minsulee2: Removed!

p2p/index.js
minsulee2: Done!

p2p/index.js
minsulee2: Right it is something more like update rather than just set. I took updatePeerCandidateInfo! Thanks!

p2p/index.js
minsulee2: Explained above!

p2p/index.js
minsulee2: I took populatePeerCandidatesFromPeerCandidateInfo!

p2p/index.js
```diff
@@ -1138,24 +1175,22 @@ class P2pClient {
 
   /**
    * Checks validity of JSON-RPC endpoint url based on HOSTING_ENV.
-   * @param {string} url is an IPv4 ip address.
+   * @param {string} url is json rpc endpoint url.
    */
   isValidJsonRpcUrl(url) {
```
platfowner, line 1180: could you add some test cases for this function (maybe after switching to a static function)?

platfowner: maybe it can be moved to p2p-util.js?

p2p/index.js
```diff
@@ -1168,56 +1203,23 @@ class P2pClient {
...
-    }
-    const resp = await sendGetRequest(
-        peerCandidateJsonRpcUrl, JSON_RPC_METHODS.P2P_GET_PEER_CANDIDATE_INFO, { });
-    const peerCandidateInfo = _.get(resp, 'data.result.result');
-    if (!peerCandidateInfo) {
-      logger.error(`Invalid peer candidate info from peer candidate url ` +
-          `(${peerCandidateJsonRpcUrl}).`);
-      return;
-    }
-    // NOTE(platfowner): As peerCandidateUrl can be a domain name url with multiple nodes,
```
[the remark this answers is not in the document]
minsulee2: My apology! Revived!

p2p/index.js
minsulee2: Good to move p2p-util! Done!

minsulee2:
At last! PTAL again!

platfowner (approved):
LGTM

minsulee2:
Thanks! Merging now!

Issue #1053, opened 2022-05-31
[BUG] Inconsistent structure of json-rpc results


platfowner:
We're using somewhat inconsistent structures for json-rpc results:

Examples:
https://github.com/ainblockchain/ain-blockchain/blob/develop/json_rpc/version.js#L44 
```
      } else if (semver.lt(coercedVer, minProtocolVersion) ||
          (maxProtocolVersion && semver.gt(coercedVer, maxProtocolVersion))) {
        const latency = Date.now() - beginTime;
        trafficStatsManager.addEvent(TrafficEventTypes.JSON_RPC_GET, latency);
        done(null, JsonRpcUtil.addProtocolVersion({
          code: JsonRpcApiResultCode.PROTO_VERSION_INCOMPATIBLE,
          message: 'Incompatible protocol version.'
        }));
      } else {
        const latency = Date.now() - beginTime;
        trafficStatsManager.addEvent(TrafficEventTypes.JSON_RPC_GET, latency);
        done(null, JsonRpcUtil.addProtocolVersion({
          code: JsonRpcApiResultCode.SUCCESS,
          result: 'Success'
        }));
      }
```

https://github.com/ainblockchain/ain-blockchain/blob/develop/json_rpc/database.js#L53
```
        case ReadDbOperations.GET:
          result = node.db.get(args.op_list);
          latency = Date.now() - beginTime;
          trafficStatsManager.addEvent(TrafficEventTypes.JSON_RPC_GET, latency);
          done(null, JsonRpcUtil.addProtocolVersion({ result }));
          return;
        default:
          latency = Date.now() - beginTime;
          trafficStatsManager.addEvent(TrafficEventTypes.JSON_RPC_GET, latency);
          done(null, JsonRpcUtil.addProtocolVersion({
            result: {
              code: JsonRpcApiResultCode.GET_INVALID_OPERATION,
              message: 'Invalid get operation'
            }
          }));
      }
    },
```

platfowner:
ain-js provider.ts:
```
  /**
   * Creates the JSON-RPC payload and sends it to the node.
   * @param {string} rpcMethod
   * @param {any} params
   * @return {Promise<any>}
   */
  async send(rpcMethod: string, params?: any): Promise<any> {
    const data = {
      jsonrpc: "2.0",
      method: rpcMethod,
      params: Object.assign(params || {}, { protoVer: this.ain.net.protoVer }),
      id: 0
    };
[...]

PR #1063, opened 2022-06-15
Internally separate error from result in get API results

platfowner:
Change summary:
- Internally separate error from result in get API results
- Add enableErrorResultSeparationForGetApis dev flag

Note:
- With enableErrorResultSeparationForGetApis = false, the APIs are backward-compatible.

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1053

### DIFF
--- a/db/index.js
+++ b/db/index.js
@@ -497,25 +497,38 @@ class DB {
 
   static readFromStateRoot(stateRoot, rootLabel, refPath, options, shardingPath) {
     const isGlobal = options && options.isGlobal;
-    if (!stateRoot) return null;
+    if (!stateRoot) {
+      return {
+        result: null
+      };
+    }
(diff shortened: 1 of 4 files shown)


db/index.js
```diff
@@ -525,18 +538,34 @@ class DB {
   }
 
   getValue(valuePath, options) {
+    return this.getValueWithError(valuePath, options).result;
```
minsulee2, line 541: Is it guaranteed it always have result?

minsulee2 (approved):
LGTM with minor question!

db/index.js
platfowner: Yes, we can see that readFromStateRoot() always returns 'result'.

platfowner:
Thanks for the fast review! Now merging.. @minsulee2

PR #1070, opened 2022-06-23
Fix result formats of ain_validateAppName() and ain_checkProtocolVersion()

platfowner:
Change summary:
- Fix result format of ain_validateAppName() json rpc API
- Let ain_checkProtocolVersion() json rpc API return boolean result
- Check empty tx_list in ain_sendSignedTransactionBatch() json rpc API
- Make json rpc APIs always return result

Related PRs:
- https://github.com/ainblockchain/ain-js/pull/82

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1053

### DIFF
--- a/common/version-util.js
+++ b/common/version-util.js
@@ -50,6 +50,7 @@ class VersionUtil {
       res.status(200)
         .set('Content-Type', 'application/json')
         .send({
+          result: null,
           code: JsonRpcApiResultCode.PROTO_VERSION_NOT_SPECIFIED,
           message: 'Protocol version not specified.',
           protoVer: BlockchainConsts.CURRENT_PROTOCOL_VERSION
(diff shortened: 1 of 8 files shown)


common/version-util.js
```diff
@@ -50,6 +50,7 @@ class VersionUtil {
       res.status(200)
         .set('Content-Type', 'application/json')
         .send({
+          result: null,
```
liayoo, line 53: question) is there a reason why we use null here and false for other APIs?

platfowner: This is a proto version validator applied for all json-rpc requests, so it doesn't know what's the expected type of 'result' (i.e., whether it's boolean or not).

liayoo (approved):
LGTM, thanks!

platfowner:
Thanks for the review. Now merging.. @minsulee2 @liayoo

PR #1074, opened 2022-06-27
Docker multi-arch build support

kriii:
Now, support both `linux/amd64` and `linux/arm64` architecture builds.


platfowner (approved):
LGTM
Thanks!

kriii:
Thanks for the review! I checked the deploying works well and now merge.

Issue #1072, opened 2022-06-27
[BUG] Broken integration test


platfowner:
Broken integration:

test log:
```
yarn run test_integration_sharding

yarn run v1.22.18
$ BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/3-nodes ./node_modules/mocha/bin/mocha --timeout 640000 test/integration/sharding.test.js
[createTimerFlagEnabledBandageMap] Registering bandage files:
[createTimerFlagEnabledBandageMap] [0] Registering /home/se/ain-blockchain/db/bandage-files/allow_lower_case_app_names_only.js
[createTimerFlagEnabledBandageMap] [1] Registering /home/se/ain-blockchain/db/bandage-files/create_app_config_sanitization.js
[createTimerFlagEnabledBandageMap] [2] Registering /home/se/ain-blockchain/db/bandage-files/native_service_path_length_check.js
[createTimerFlagEnabledBandageMap] [3] Registering /home/se/ain-blockchain/db/bandage-files/update_max_state_tree_size_per_byte.js
[createTimerFlagEnabledBandageMap] [4] Registering /home/se/ain-blockchain/db/bandage-files/staking_balance_total_sum.js
[createTimerFlagEnabledBandageMap] [5] Registering /home/se/ain-blockchain/db/bandage-files/update_max_state_tree_size_per_byte2.js
[createTimerFlagEnabledBandageMap] [6] Registering /home/se/ain-blockchain/db/bandage-files/update_bandwidth_budget_params.js
[createTimerFlagEnabledBandageMap] [7] Skipping for timer flag: disable_tx_receipt_recording
[createTimerFlagEnabledBandageMap] [8] Skipping for timer flag: extend_account_registration_gas_amount
[createTimerFlagEnabledBandageMap] [9] Registering /home/se/ain-blockchain/db/bandage-files/add_app_creation_gas_amount.js


  Sharding
[createTimerFlagEnabledBandageMap] Registering bandage files:
[createTimerFlagEnabledBandageMap] [0] Registering /home/se/ain-blockchain/db/bandage-files/allow_lower_case_app_names_only.js
[createTimerFlagEnabledBandageMap] [1] Registering /home/se/ain-blockchain/db/bandage-files/create_app_config_sanitization.js
[createTimerFlagEnabledBandageMap] [2] Registering /home/se/ain-blockchain/db/bandage-files/native_service_path_length_check.js
[...]

PR #1076, opened 2022-06-29
Hotfix integration test

minsulee2:
My apology! This is the same issue as it's just fixed.
Ref: #1072

### DIFF
--- a/test/integration/he_sharding.test.js
+++ b/test/integration/he_sharding.test.js
@@ -370,6 +370,10 @@ describe('HE Sharding', () => {
         }
         expect(latestAfter).to.be.greaterThan(latestBefore);
       });
+
+      after(() => {
+        delete ENV_VARIABLES[2].PEER_CANDIDATE_JSON_RPC_URL;
+      });
     });
   });
 })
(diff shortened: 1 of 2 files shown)


test/integration/he_sharding.test.js
```diff
@@ -352,6 +352,7 @@ describe('HE Sharding', () => {
         server1_proc.kill();
         await waitForNewBlocks(server2, sharding.reporting_period * 3);
         console.log(`        --> Restarting server[0]...`);
+        ENV_VARIABLES[2].PEER_CANDIDATE_JSON_RPC_URL = "http://localhost:9092/json-rpc";
```
platfowner, line 355: minor) what about deleting this env variable once the test is completed to minimize side effects (sharding.test.js case as well)?

minsulee2: Done!

minsulee2:
PTAL!

test/integration/he_sharding.test.js
platfowner: minor) oh, could this line be moved to 'before()' like deletion is done in 'after()'?

platfowner (approved):
LGTM with some suggestions.
Thanks for handling this issue!!

test/integration/he_sharding.test.js
minsulee2: Done!

minsulee2:
Thanks for the fast review! Merging now!

PR #1078, opened 2022-07-01
Release/v1.0.8

minsulee2:
Updated features:
- read and write limits
- Update some scripts
- Update tests
- Add child node pagination with is_partial option
- Tweak API data size related node params
- Remove disconnectable peer candidates
- Internally separate error from result in get API results
- Simplify result format by not using 'error' property
- Remove invalid txs when finalize block
- Turn on UPDATE_FRONT_DB_OPTION for all nodes
- Docker multi-arch build support
- Use getTreeMaxSiblings() instead of numChildren() in max siblings limit check


minsulee2:
Thanks! merging now!

PR #1088, opened 2022-07-20
Feature/ehgmsdk20/tx state changed event

ehgmsdk20:
Change summary:

Add TX_STATE_CHANGED event
Add test code for TX_STATE_CHANGED event


common/constants.js
```diff
@@ -664,6 +673,11 @@ const ValueChangedEventSources = {
   USER: 'USER',
 };
 
+const CauseForFilterDeletion = {
```
liayoo, line 676: suggestion) FilterDeletionReasons

event-handler/index.js
```diff
@@ -120,6 +122,72 @@ class EventHandler {
...
+        transaction: transaction,
+        tx_state: {
+          before: beforeState,
+          after: afterState,
+        },
+      });
+      this.eventChannelManager.transmitEventByEventFilterId(eventFilterId, blockchainEvent);
+      const [channelId, clientFilterId] = eventFilterId.split(':');
+
+      // NOTE(ehgmsdk20): When the state no longer changes, the event filter is removed.
```
liayoo, line 162: 🙌

common/constants.js
ehgmsdk20: It looks better, thanks!
Is it better to change 
`cause: CauseForFilterDeletion` 
to
`reason: FilterDeletionReasons`? (event-handler/index.js line 186)

common/constants.js
```diff
@@ -572,6 +572,13 @@ function isTxInBlock(state) {
   return state === TransactionStates.FINALIZED || state === TransactionStates.REVERTED;
 }
 
+function isPermanentState(state) {
```
platfowner, line 575: suggestion) what about isEndState()?

common/constants.js
```diff
@@ -664,6 +673,11 @@ const ValueChangedEventSources = {
   USER: 'USER',
 };
 
+const FilterDeletionReasons = {
+  TIMED_OUT: 'TIMED_OUT',
```
platfowner, line 677: suggestion) what about FILTER_TIMEOUT?

common/constants.js
```diff
@@ -664,6 +673,11 @@ const ValueChangedEventSources = {
...
+  PERMANENT_STATE: 'PERMANENT_STATE',
```
platfowner, line 678: suggestion) what about END_OF_STATE?

event-handler/index.js
```diff
@@ -120,6 +122,72 @@ class EventHandler {
...
+        logger.error(`[${LOG_HEADER}] Missing config in eventFilter(${eventFilterId})`);
+        continue;
+      }
+
+      if (txHash !== transaction.hash) {
+        continue;
+      }
+
+      const deleteCallback = this.eventFilterIdToDeleteCallback.get(eventFilterId);
+
```
platfowner, line 147: could you remove this empty line?

event-handler/index.js
```diff
@@ -18,6 +19,7 @@ class EventHandler {
     this.stateEventTreeManager = new StateEventTreeManager();
     this.eventFilters = {};
     this.eventTypeToEventFilterIds = {};
+    this.eventFilterIdToDeleteCallback = new Map();
```
platfowner, line 22: it seems this maps is currently for timeout callbacks. then what about just 'eventFilterIdToTimeoutCallback'?

event-handler/index.js
```diff
@@ -120,6 +122,72 @@ class EventHandler {
...
+      logger.error(`[${LOG_HEADER}] Invalid Tx(${JSON.stringify(transaction)})`);
+      return;
+    }
+    for (const eventFilterId of this.eventTypeToEventFilterIds[
+          BlockchainEventTypes.TX_STATE_CHANGED]) {
+      const eventFilter = this.eventFilters[eventFilterId];
+      const txHash = _.get(eventFilter, 'config.tx_hash', null);
+      const timeout = _.get(eventFilter, 'config.timeout', null);
+
+      if (!txHash || !timeout) {
```
platfowner, line 137: could you check if these sanity checks are done when event filters are registered?

event-handler/index.js
```diff
@@ -120,6 +122,72 @@ class EventHandler {
...
+        logger.error(`[${LOG_HEADER}] Missing config in eventFilter(${eventFilterId})`);
+        continue;
+      }
+
+      if (txHash !== transaction.hash) {
```
platfowner, line 142: from performance perspective, we need to use a map instead of this loop check?

event-handler/index.js
```diff
@@ -120,6 +122,72 @@ class EventHandler {
...
+    if (!_.get(transaction, 'hash', null)) {
+      logger.error(`[${LOG_HEADER}] Invalid Tx(${JSON.stringify(transaction)})`);
+      return;
+    }
+    for (const eventFilterId of this.eventTypeToEventFilterIds[
+          BlockchainEventTypes.TX_STATE_CHANGED]) {
+      const eventFilter = this.eventFilters[eventFilterId];
+      const txHash = _.get(eventFilter, 'config.tx_hash', null);
+      const timeout = _.get(eventFilter, 'config.timeout', null);
+
```
platfowner, line 136: minor) could you remove empty lines below unless necessary?

event-handler/index.js
```diff
@@ -120,6 +122,72 @@ class EventHandler {
...
+      const [channelId, clientFilterId] = eventFilterId.split(':');
+
+      // NOTE(ehgmsdk20): When the state no longer changes, the event filter is removed.
+      if (isPermanentState(afterState)) {
+        this.emitFilterDeleted(eventFilterId, FilterDeletionReasons.PERMANENT_STATE);
+        this.deregisterEventFilter(clientFilterId, channelId);
+        return;
+      }
+
+      this.eventFilterIdToDeleteCallback.set(eventFilterId, setTimeout(() => {
```
platfowner, line 169: why is this callback is set when an event is emitted?

event-handler/index.js
```diff
@@ -120,6 +122,72 @@ class EventHandler {
...
+        this.emitFilterDeleted(eventFilterId, FilterDeletionReasons.TIMED_OUT);
+        this.deregisterEventFilter(clientFilterId, channelId);
+      }, timeout));
+    }
+  }
+
+  emitFilterDeleted(eventFilterId, FilterDeletionReason) {
```
platfowner, line 176: s/FilterDeletionReason/filterDeletionReason/?

event-handler/index.js
```diff
@@ -120,6 +122,72 @@ class EventHandler {
...
+        },
+      });
+      this.eventChannelManager.transmitEventByEventFilterId(eventFilterId, blockchainEvent);
+      const [channelId, clientFilterId] = eventFilterId.split(':');
+
+      // NOTE(ehgmsdk20): When the state no longer changes, the event filter is removed.
+      if (isPermanentState(afterState)) {
+        this.emitFilterDeleted(eventFilterId, CauseForFilterDeletion.PERMANENT_STATE);
+        this.deregisterEventFilter(clientFilterId, channelId);
+        return;
```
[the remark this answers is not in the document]
ehgmsdk20: Thanks! I fixed this bug and added test code to catch this bug. [commit](https://github.com/ainblockchain/ain-blockchain/pull/1088/commits/29a1f5ee96af19ece928cf46be0460e1ce862e7a)

common/constants.js
ehgmsdk20: It looks better! i changed to isEndedState(). Thanks!

common/constants.js
ehgmsdk20: Fixed, thanks!

common/constants.js
ehgmsdk20: Fixed, thanks!

event-handler/index.js
ehgmsdk20: Fixed, thanks!

event-handler/index.js
ehgmsdk20: That looks more accurate, thanks!

event-handler/index.js
ehgmsdk20: Is it correct to say something like txHashToFilterId?

event-handler/index.js
ehgmsdk20: Fixed, thanks!

event-handler/index.js
ehgmsdk20: Fixed, thanks!

common/constants.js
liayoo: Yes, please! :)

event-handler/index.js
ehgmsdk20: I put it in case it stays without going to the end of state after it is emitted once, but there seems to be very few cases like that. I will delete it thanks!

ehgmsdk20: [commit](https://github.com/ainblockchain/ain-blockchain/pull/1088/commits/e8e863d17e1e8ce9ac4c0e0fb71646020724f510)

event-handler/index.js
ehgmsdk20: These sanity checks are done in validateEventFilterConfig function(line 197) when event filters are registered.
`case BlockchainEventTypes.TX_STATE_CHANGED:
    const txHash = _.get(config, 'tx_hash', null);
    const timeout = _.get(config, 'timeout', null);

    // NOTE(ehgmsdk20): If the epoch_ms is changed, it must be changed to get it
    // through the getBlockchainParams function.
    const epochMs = _.get(BlockchainParams, 'genesis.epoch_ms', 30000);

    if (!txHash|| !timeout) {
      throw new EventHandlerError(EventHandlerErrorCode.MISSING_PARAMS_IN_CONFIG,
          `config.tx_hash or config.timeout is missing (${JSON.stringify(config)})`);
    }
    if (!CommonUtil.isValidHash(txHash)) {
      throw new EventHandlerError(EventHandlerErrorCode.INVALID_TX_HASH,
          `Invalid tx hash (${txHash})`);
    }
    if (
      !CommonUtil.isNumber(timeout) ||
      timeout > NodeConfigs.TX_POOL_TIMEOUT_MS ||
      timeout < epochMs
    ) {
      throw new EventHandlerError(EventHandlerErrorCode.INVALID_TIMEOUT,
        `Invalid timeout (${timeout})\nTimeout must be a number between ` +
        `${epochMs} and ${NodeConfigs.TX_POOL_TIMEOUT_MS}`);
    }
    break;`

event-handler/index.js
ehgmsdk20: Done, thanks for good idea! [commit](https://github.com/ainblockchain/ain-blockchain/pull/1088/commits/9e37c2141478c57f8d8d35f125c1f813875d528f)

event-handler/index.js
```diff
@@ -120,6 +123,55 @@ class EventHandler {
...
+    if (!_.get(transaction, 'hash', null)) {
+      logger.error(`[${LOG_HEADER}] Invalid Tx(${JSON.stringify(transaction)})`);
+      return;
+    }
+    const eventFilterIds = this.txHashToEventFilterIds.get(transaction.hash);
+    if (!eventFilterIds) {
+      return;
+    }
+    for (const eventFilterId of eventFilterIds) {
+      const deleteCallback = this.eventFilterIdToTimeoutCallback.get(eventFilterId);
```
platfowner, line 137: minor) timeoutCallback here and below?

event-handler/index.js
```diff
@@ -208,6 +297,25 @@ class EventHandler {
       }
       if (eventFilter.type === BlockchainEventTypes.VALUE_CHANGED) {
         this.stateEventTreeManager.deregisterEventFilterId(eventFilterId);
+      } else if (eventFilter.type === BlockchainEventTypes.TX_STATE_CHANGED) {
+        const deleteCallback = this.eventFilterIdToTimeoutCallback.get(eventFilterId);
```
platfowner, line 301: minor) ditto?

test/unit/event-handler.test.js
```diff
@@ -88,6 +117,168 @@ describe('EventHandler Test', () => {
...
+      it('emit tx_state_changed event which is not permenant state', async () => {
+        const numberOfFiltersBefore = Object.keys(eventHandler.eventFilters).length;
+        const now = Date.now();
+        const timeout = 2 * epochMs;
+        eventHandler.createAndRegisterEventFilter(now, now,
+        BlockchainEventTypes.TX_STATE_CHANGED, {
+          tx_hash: validTxHash,
+          timeout,
+        });
+        // NOTE(ehgmsdk20): To check whether the timer of deleteCallback is reset
```
platfowner, line 155: minor) timeoutCallback?

common/constants.js
```diff
@@ -572,6 +572,13 @@ function isTxInBlock(state) {
   return state === TransactionStates.FINALIZED || state === TransactionStates.REVERTED;
 }
 
+function isEndedState(state) {
```
platfowner, line 575: minor) what about just isEndState?

common/constants.js
```diff
@@ -664,6 +673,11 @@ const ValueChangedEventSources = {
   USER: 'USER',
 };
 
+const FilterDeletionReasons = {
+  FILTER_TIMEOUT: 'FILTER_TIMEOUT',
+  END_OF_STATE: 'END_OF_STATE',
```
platfowner, line 678: minor) sorry, can this be changed again to 'END_STATE_REACHED' or something?

event-handler/index.js
platfowner: great, thanks!

event-handler/index.js
```diff
@@ -158,6 +210,32 @@ class EventHandler {
...
+
+        if (!txHash|| !timeout) {
+          throw new EventHandlerError(EventHandlerErrorCode.MISSING_PARAMS_IN_CONFIG,
+              `config.tx_hash or config.timeout is missing (${JSON.stringify(config)})`);
+        }
+        if (!CommonUtil.isValidHash(txHash)) {
+          throw new EventHandlerError(EventHandlerErrorCode.INVALID_TX_HASH,
+              `Invalid tx hash (${txHash})`);
+        }
+        if (
```
platfowner, line 229: minor) could you use
```
        if (!CommonUtil.isNumber(timeout) ||
            timeout > NodeConfigs.TX_POOL_TIMEOUT_MS ||
            timeout < epochMs) {
```
style?

event-handler/index.js
```diff
@@ -158,6 +210,32 @@ class EventHandler {
...
+          !CommonUtil.isNumber(timeout) ||
+          timeout > NodeConfigs.TX_POOL_TIMEOUT_MS ||
+          timeout < epochMs
+        ) {
+          throw new EventHandlerError(EventHandlerErrorCode.INVALID_TIMEOUT,
+            `Invalid timeout (${timeout})\nTimeout must be a number between ` +
```
platfowner, line 235: 👍

event-handler/index.js
```diff
@@ -178,6 +256,17 @@ class EventHandler {
...
+      } else if (eventType === BlockchainEventTypes.TX_STATE_CHANGED) {
+        const eventFilterIds = this.txHashToEventFilterIds.get(config.tx_hash)
+        if (eventFilterIds) {
+          eventFilterIds.push(eventFilterId);
+        } else {
+          this.txHashToEventFilterIds.set(config.tx_hash, [eventFilterId]);
+        }
+        this.eventFilterIdToTimeoutCallback.set(eventFilterId, setTimeout(() => {
+          this.emitFilterDeleted(eventFilterId, FilterDeletionReasons.FILTER_TIMEOUT);
+          this.deregisterEventFilter(clientFilterId, channelId);
```
platfowner, line 268: minor) what about swapping the above two lines?

event-handler/index.js
```diff
@@ -120,6 +123,55 @@ class EventHandler {
...
+      return;
+    }
+    const eventFilterIds = this.txHashToEventFilterIds.get(transaction.hash);
+    if (!eventFilterIds) {
+      return;
+    }
+    for (const eventFilterId of eventFilterIds) {
+      const deleteCallback = this.eventFilterIdToTimeoutCallback.get(eventFilterId);
+      if (deleteCallback) {
+        clearTimeout(deleteCallback);
```
platfowner, line 139: is it safe to clearTimeout here (i.e. couldn't afterState be not an end state)?
could this be just skipped here as it will be called by deregisterEventFilter() anyway.

event-handler/index.js
```diff
@@ -208,6 +297,25 @@ class EventHandler {
...
         this.stateEventTreeManager.deregisterEventFilterId(eventFilterId);
+      } else if (eventFilter.type === BlockchainEventTypes.TX_STATE_CHANGED) {
+        const deleteCallback = this.eventFilterIdToTimeoutCallback.get(eventFilterId);
+        if (deleteCallback) {
+          clearTimeout(deleteCallback);
+        }
+        this.eventFilterIdToTimeoutCallback.delete(eventFilterId);
+
+        const txHash = _.get(eventFilter.config, 'tx_hash', null);
+        if (!txHash) {
```
platfowner, line 308: minor) can this be just skipped as it's always done by validateEventFilterConfig()?

test/unit/event-handler.test.js
```diff
@@ -88,6 +117,168 @@ describe('EventHandler Test', () => {
...
+        expect(eventHandler.txHashToEventFilterIds.get(validTxHash)).to.equal(undefined);
+      });
+    });
+
+    describe('emitTxStateChanged', () => {
+      beforeEach( async () => { // NOTE(cshcomcom): To avoid id collisions.
+        await new Promise((resolve) => setTimeout(resolve, 1000));
+      });
+
+      it('emit tx_state_changed event which is not permenant state', async () => {
```
platfowner, line 146: s/permenant state/an end state/?

test/unit/event-handler.test.js
```diff
@@ -88,6 +117,168 @@ describe('EventHandler Test', () => {
...
+        expect(numberOfFiltersBefore + 1).to.equal(numberOfFiltersAfter); // Filter is not deleted
+        expect(eventHandler.txHashToEventFilterIds.get(validTxHash)).to.deep.equal(
+          [eventHandler.getGlobalFilterId(now, now)]);
+        eventHandler.deregisterEventFilter(now, now);
+        numberOfFiltersAfter = Object.keys(eventHandler.eventFilters).length;
+        expect(numberOfFiltersBefore).to.equal(numberOfFiltersAfter);
+        expect(eventHandler.txHashToEventFilterIds.get(validTxHash)).to.equal(undefined);
+      });
+
+      it('emit tx_state_changed event which is permenant state', () => {
```
platfowner, line 175: s/permenant state/an end state/?

tx-pool/index.js
```diff
@@ -94,6 +96,9 @@ class TransactionPool {
       this.freeTxCountTotal++;
       this.updateFreeTxCountPerAccount(tx.address, 1);
     }
+    if (this.eh) {
+      this.eh.emitTxStateChanged(tx, null, txState);
```
platfowner, line 100: 👍

tx-pool/index.js
```diff
@@ -417,7 +422,20 @@ class TransactionPool {
...
     const sizeBefore = this.txCountTotal;
     for (const [address, txList] of this.transactions.entries()) {
-      const filterFunc = (tx) => !this.isTimedOutFromPool(tx.extra.created_at, blockTimestamp);
+      const filterFunc = (tx) => {
+        const isTimedOut = this.isTimedOutFromPool(tx.extra.created_at, blockTimestamp);
+        if (isTimedOut) {
+          const tracked = this.transactionTracker.get(tx.hash);
+          const beforeState = _.get(tracked, 'state', null);
+          if (tracked && !isTxInBlock(beforeState)) {
+            tracked.state = TransactionStates.TIMED_OUT;
```
platfowner, line 431: 👍

platfowner (commented):
Thanks for the nice implementation.
Let me leave some suggestions and comments.

event-handler/index.js
ehgmsdk20: Done, thanks!

event-handler/index.js
ehgmsdk20: Done, thanks!

test/unit/event-handler.test.js
ehgmsdk20: Done, thanks!

common/constants.js
ehgmsdk20: Done, thanks!

common/constants.js
ehgmsdk20: Done, thanks!

event-handler/index.js
ehgmsdk20: Done, thanks!

event-handler/index.js
ehgmsdk20: The logical flow is what you said, but after an event filter is deregistered, no events can be passed to that filter.

event-handler/index.js
ehgmsdk20: There seems to be a difference between the timeout I thought and the timeout you thought. 
I thought that it exists to delete a tx that does not have a timeout when it is registered in the event filter. So, when emitTxStateChanged is executed at least once, clearTimeout is called. 
You seems to think of timeout as the time until tx goes to endState. 
Which one is better?

test/unit/event-handler.test.js
ehgmsdk20: Done, thanks!

test/integration/event_handler.test.js
```diff
@@ -280,5 +285,137 @@ describe('Event Handler Test', function() {
...
+          if (messageType === BlockchainEventMessageTypes.EMIT_EVENT &&
+              eventType === BlockchainEventTypes.FILTER_DELETED) {
+            expect(payload.reason).to.equal(FilterDeletionReasons.FILTER_TIMEOUT);
+            expect(payload.filter_id).to.equal(filterId.toString());
+            done();
+          }
+        });
+      });
+
+      it('deleted because of permanent state', function(done) {
```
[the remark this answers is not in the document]
ehgmsdk20: Done, thanks!

test/unit/event-handler.test.js
ehgmsdk20: Done, thanks!

event-handler/index.js
```diff
@@ -208,6 +297,25 @@ class EventHandler {
...
+          throw new EventHandlerError(EventHandlerErrorCode.MISSING_PARAMS_IN_CONFIG,
+              `Missing tx_hash in config (${eventFilterId})`, eventFilterId);
+        }
+        let eventFilterIds = this.txHashToEventFilterIds.get(txHash);
+        eventFilterIds = eventFilterIds.filter((filterId) => filterId !== eventFilterId)
+        if (eventFilterIds.length === 0) {
+          this.txHashToEventFilterIds.delete(txHash)
+        } else {
+          this.txHashToEventFilterIds.set(txHash, eventFilterIds);
```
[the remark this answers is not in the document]
ehgmsdk20: Thanks for good idea! Edit done! [commit](https://github.com/ainblockchain/ain-blockchain/pull/1088/commits/6f56ce0d89c3cb9a89d78429c09c62d59beed9dc)

event-handler/index.js
ehgmsdk20: I moved this logic inside the emitFilterDeleted function so that filter is deleted immediately after emitting the event

ehgmsdk20:
> Thanks for the nice implementation. Let me leave some suggestions and comments.

Thanks for review. 
Most of the reviews were reflected, but some reviews seem to need more discussion.
PTAL @platfowner

event-handler/index.js
```diff
@@ -18,6 +19,8 @@ class EventHandler {
     this.stateEventTreeManager = new StateEventTreeManager();
     this.eventFilters = {};
     this.eventTypeToEventFilterIds = {};
+    this.eventFilterIdToTimeoutCallback = new Map();
+    this.txHashToEventFilterIds = new Map();
```
platfowner, line 23: now txHashToEventFilterIdSet?

event-handler/index.js
```diff
@@ -120,6 +123,60 @@ class EventHandler {
     }
   }
 
+  emitTxStateChanged(transaction, beforeState, afterState) {
+    const LOG_HEADER = 'emitTxStateChanged';
+    if (!_.get(transaction, 'hash', null)) {
+      logger.error(`[${LOG_HEADER}] Invalid Tx(${JSON.stringify(transaction)})`);
+      return;
+    }
+    const eventFilterIds = this.txHashToEventFilterIds.get(transaction.hash);
```
platfowner, line 132: now eventFilterIdSet?

event-handler/index.js
```diff
@@ -178,6 +259,16 @@ class EventHandler {
       this.eventTypeToEventFilterIds[eventType].add(eventFilterId);
       if (eventType === BlockchainEventTypes.VALUE_CHANGED) {
         this.stateEventTreeManager.registerEventFilterId(config.path, eventFilterId);
+      } else if (eventType === BlockchainEventTypes.TX_STATE_CHANGED) {
+        const eventFilterIds = this.txHashToEventFilterIds.get(config.tx_hash);
```
platfowner, line 263: ditto?

event-handler/index.js
```diff
@@ -208,6 +299,19 @@ class EventHandler {
...
         this.stateEventTreeManager.deregisterEventFilterId(eventFilterId);
+      } else if (eventFilter.type === BlockchainEventTypes.TX_STATE_CHANGED) {
+        const timeoutCallback = this.eventFilterIdToTimeoutCallback.get(eventFilterId);
+        if (timeoutCallback) {
+          clearTimeout(timeoutCallback);
+        }
+        this.eventFilterIdToTimeoutCallback.delete(eventFilterId);
+
+        const txHash = _.get(eventFilter.config, 'tx_hash', null);
+        const eventFilterIds = this.txHashToEventFilterIds.get(txHash);
```
platfowner, line 310: ditto?

event-handler/index.js
```diff
@@ -208,6 +299,19 @@ class EventHandler {
...
+        eventFilterIds.delete(eventFilterId);
+        if (eventFilterIds.size === 0) {
+          this.txHashToEventFilterIds.delete(txHash);
```
platfowner, line 313: 👍

event-handler/index.js
platfowner: oh, i see.
suggestion) what about 'deleteFilterAndEmitFilterDeleted'?
or it seems okay to have deregisterFilter at the original place.

platfowner: to me, having them separate looks clearer.

event-handler/index.js
platfowner: what do you mean by "a tx that does not have a timeout"?

assume that there are three states A, B, C and only C is an end state. what happens when there is a state change A -> B but no more?
won't this filter be removed?
i thought that timeout callback is a kind of safety gear to avoid filter map bloating.
what do you think?

ehgmsdk20: I thought the filter was just there to prevent registering a fictitious tx_hash (a valid tx hash but not actually sending it?).

ehgmsdk20: And there seems to be no case where the above-mentioned state B does not change any more. B will eventually go to the TIMED_OUT state after some time.

ehgmsdk20: The difference between the two methods seems to be noticeable when tx times out. (Will the filter be deleted because of timeout or end_state_reached?)

platfowner: oh, now it got the underlying idea of the timeout callback. thanks for the explanation.
then, from the user's perspective, if the timeout callback can be cleared by some reasons different from the timeout itself, can it be confusing somehow? 
@ehgmsdk20

event-handler/index.js
ehgmsdk20: Fixed, thanks!

event-handler/index.js
ehgmsdk20: Fixed, thanks!

event-handler/index.js
ehgmsdk20: Fixed, thanks!

event-handler/index.js
ehgmsdk20: Fixed, thanks!

event-handler/index.js
ehgmsdk20: I separated filter deletion logic, thanks!

event-handler/index.js
```diff
@@ -120,6 +123,56 @@ class EventHandler {
...
+        tx_state: {
+          before: beforeState,
+          after: afterState,
+        },
+      });
+      this.eventChannelManager.transmitEventByEventFilterId(eventFilterId, blockchainEvent);
+
+      // NOTE(ehgmsdk20): When the state no longer changes, the event filter is removed.
+      if (isEndState(afterState)) {
+        // NOTE(ehgmsdk20): Send message that the filter has been deleted before deleting,
```
platfowner, line 153: 👍

event-handler/index.js
platfowner: for the purpose you mentioned, can we apply a constant timeout value instead of letting the user set it? e.g. epoch_ms * 10 = 200000?
@ehgmsdk20

ehgmsdk20: Oh, it would be nice to set timeout to a constant! When I tested with ain-js, most of the transactions were completed within 2 * epoch time. It seems that it would be okay to set it to only 5 * epoch time. (If we ignore the case where tx state is timed_out)

platfowner: summary of separate discussion:
- Let's go with constant timeout value
- When a tx state changed event filter is registered, the tx tracker is looked up for the input tx hash and if it's already there, an state changed event is initially emitted with the state there
- In the case of initially emitting state above, timeout callback is not added
@ehgmsdk20

ehgmsdk20: Done! PTAL @platfowner

event-handler/index.js
```diff
@@ -120,6 +123,75 @@ class EventHandler {
...
+    const timeoutId = this.eventFilterIdToTimeoutCallback.get(eventFilterId);
+    if (timeoutId) {
+      clearTimeout(timeoutId);
+    }
+
+    const epochMs = this.node.getBlockchainParam('genesis/epoch_ms');
+    this.eventFilterIdToTimeoutCallback.set(eventFilterId, setTimeout(() => {
+      this.emitFilterDeleted(eventFilterId, FilterDeletionReasons.FILTER_TIMEOUT);
+      this.eventChannelManager.deregisterFilterByEventFilterId(eventFilterId);
+    }, 5 * epochMs));
```
platfowner, line 192: Can the timeout value be added as a node param e.g. EVENT_HANDLER_FILTER_DELETION_TIMEOUT_MS = 100000?

blockchain-configs/afan-shard/node_params.json
```diff
@@ -28,8 +28,9 @@
   "ENABLE_REST_FUNCTION_CALL": false,
   "ENABLE_STATUS_REPORT_TO_TRACKER": true,
   "ENABLE_TX_SIG_VERIF_WORKAROUND": false,
-  "EVENT_HANDLER_PORT": 6000,
+  "EVENT_HANDLER_FILTER_DELETION_TIMEOUT_MS": 15000,
```
platfowner, line 31: qq) just curious when it has 15000 and when it has 100000?

platfowner (approved):
LGTM with some comments.
Thanks for the nice work!

blockchain-configs/afan-shard/node_params.json
ehgmsdk20: We ended up with a filter timeout of 5 * epoch_ms. In the case of afan-shard or he-shard, filter timeout is set to 15000 because epoch_ms=3000. and the rest is set to epoch_ms=20000, so filter timeout is set to 100000.

platfowner: got it. thanks for the explanation.

ehgmsdk20:
Thanks for your reviews!

PR #1136, opened 2023-01-30
Delete dependabot.yml

ehgmsdk20:
According to the [Github official documentation](https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts#detection-of-insecure-dependencies), dependabot's security scan is executed only for the default branch.
With dependabot.yml, I couldn't make the security scan run in the develop branch, so I removed the dependabot.yml file.


platfowner (approved):
LGTM
Thanks!

Issue #1138, opened 2023-02-08
[BUG] REST functions are called 10 times


platfowner:
REST functions are called 10 times, which should be 3.

PR #1160, opened 2023-04-26
Skip REST function call when eventSource is null 

platfowner:
Change summary:
- Trigger REST function call only when eventSource is NOT null ('USER' or 'BLOCK')
- Move value changed event emission logic to db side

cc: @jakepyo 

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1138

### DIFF
--- a/test/unit/functions.test.js
+++ b/test/unit/functions.test.js
@@ -300,6 +300,7 @@ describe("Functions", () => {
               timestamp: 1234567890000,
               blockNumber: 1000,
               blockTime: 1234567890999,
+              eventSource: 'USER',
             });
         assert.deepEqual(triggerRes.func_results, {
           "0x00001": {
@@ -342,7 +343,9 @@ describe("Functions", () => {
                 "timestamp": 1234567890000,
                 "blockNumber": 1000,
                 "blockTime": 1234567890999,
+                "eventSource": "USER",
               },
+              "eventSource": "USER",
               "params": {},
               "prevValue": "prev value",
               "timestamp": 1234567890000,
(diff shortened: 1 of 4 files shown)


test/unit/functions.test.js
```diff
@@ -2605,14 +2659,84 @@ describe("Functions", () => {
...
+            "timestamp": 1566736760322,
+            "gas_price": 1,
+          },
+          "extra": {
+            "created_at": 1566736760323,
+            "executed_at": 1566736760324,
+          }
+        }
+        const triggerRes = functions.matchAndTriggerFunctions(
+            CommonUtil.parsePath(refPathRest), value, "prev value", { addr: 'abcd' },
```
jiyoung-an, line 2681: question) Is `addr` wallet address? (if so, does the function check if the wallet address is valid?)

test/unit/functions.test.js
```diff
@@ -2605,14 +2659,84 @@ describe("Functions", () => {
...
+              timestamp: 1234567890000,
+              blockNumber: 1000,
+              blockTime: 1234567890999,
+              eventSource: null,
+            });
+        return triggerRes.func_promises.then((resp) => {
+          assert.deepEqual(resp, {
+            func_count: 1,
+            trigger_count: 0,
+            fail_count: 0,
```
jiyoung-an, line 2727: question) Is the `fail_count` value 0 because the event source is skip, not fail?

jiyoung-an (approved):
LGTM with some questions, thank!

test/unit/functions.test.js
platfowner: The validity check is not done within the function. Instead it's done at the very front of the APIs for user transactions. See:
- https://github.com/ainblockchain/ain-blockchain/blob/master/json_rpc/transaction.js#L99
- https://github.com/ainblockchain/ain-blockchain/blob/master/tx-pool/transaction.js#L41

test/unit/functions.test.js
platfowner: Yes, it counts only 'triggered but failed' function calls.

platfowner:

PR #1170, opened 2023-05-12
Release/v1.0.14

platfowner:
New updates:
- Fix dockerfile node version 
- Add chain id option to tool scripts
- Fix sample chatbot server and tools
- Improve REST function call logic regarding eventSource=null 
- Add developers tool scripts (user whitelisting, url whitelisting)
- Support keystore / mnemonic / private_key key options to tool scripts
- Support HOSTING_ENV === 'aws' option
- Fix ainUtil.privateToAccount() usecases 
- Tweak json rpc url setting in start_node_docker.sh
- Add volume mount option to start_node_docker.sh


kriii (approved):
LGTM, Thanks!

Issue #1173, opened 2023-05-23
Need to reduce the number of blockchain nodes to 5


platfowner:
We need to reduce the number of blockchain nodes to 5 from 10.

PR #1174, opened 2023-05-23
Reduce the number of blockchain nodes to 5 from 10

platfowner:
Change summary:
- Reduce the number of blockchain nodes to 5 from 10 (ENABLE_JSON_RPC_API, ENABLE_REST_FUNCTION_CALL, ENABLE_EVENT_HANDLER, PEER_WHITELIST)
- Add a sample keystore file for transfer tool

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1173


deploy_blockchain_incremental_gcp.sh
```diff
@@ -337,16 +355,16 @@ else
     START_NODE_CMD_BASE="$GO_TO_PROJECT_ROOT_CMD && . start_node_incremental_gcp.sh"
 fi
 
-# Tracker server is deployed with BEGIN_PARENT_NODE_INDEX = -1
-if [[ $BEGIN_PARENT_NODE_INDEX = -1 ]]; then
+# Tracker server is deployed with PARENT_NODE_INDEX_BEGIN = -1
```
ehgmsdk20, line 358: Suggestion) how about to change
PARENT_NODE_INDEX_BEGIN -> PARENT_NODE_INDEX_GE
PARENT_NODE_INDEX_END -> PARENT_NODE_INDEX_LE

ehgmsdk20 (approved):

deploy_blockchain_incremental_gcp.sh
platfowner: Thanks for the suggestion.
As those variables are directly set as commandline options (<Parent Node Index Begin>, <Parent Node Index End>) by the user and displayed in log messages, I think we can keep them have aligned names for readability?

platfowner:

PR #1177, opened 2023-05-26
Release/v1.0.15

platfowner:
New updates:
- Reduce the number of blockchain nodes to 5 from 10
- Disable peer whitelisting
- Use a default value of TRACKER_UPDATE_JSON_RPC_URL for testnet-prod 
- Add <Parent Node Index Begin> and <Parent Node Index End> options to deploy_blockchain_genesis_gcp.sh


blockchain-configs/testnet-prod/node_params.json
```diff
@@ -68,7 +68,7 @@
   "SNAPSHOTS_INTERVAL_BLOCK_NUMBER": 1000,
   "SYNC_MODE": "fast",
   "TARGET_NUM_OUTBOUND_CONNECTION": 3,
-  "TRACKER_UPDATE_JSON_RPC_URL": "NEED TO BE SET DIFFERENTLY FOR SPRING & SUMMER",
+  "TRACKER_UPDATE_JSON_RPC_URL": "http://35.221.137.80:8080/json-rpc",
```
ehgmsdk20, line 71: question) is it okay to write the ip directly?

start_node_docker.sh
```diff
@@ -3,9 +3,9 @@ if [[ $SEASON = 'mainnet' ]]; then
 elif [[ $SEASON = 'summer' ]]; then
     export BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/testnet-prod
     export TRACKER_UPDATE_JSON_RPC_URL="http://35.194.172.106:8080/json-rpc"
+    export PEER_CANDIDATE_JSON_RPC_URL="http://35.194.169.78:8080/json-rpc"
 elif [[ $SEASON = 'spring' ]]; then
     export BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/testnet-prod
-    export TRACKER_UPDATE_JSON_RPC_URL="http://35.221.137.80:8080/json-rpc"
```
ehgmsdk20, line 8: Why the TRACKER_UPDATE_JSON_RPC_URL is deleted from spring?

start_node_incremental_gcp.sh
```diff
@@ -109,108 +120,90 @@ printf "\n#### [Step 1] Configure env vars ####\n\n"
...
-    else
-        export PEER_CANDIDATE_JSON_RPC_URL="http://104.199.237.250:8080/json-rpc"
     fi
 elif [[ $SEASON = 'summer' ]]; then
     export BLOCKCHAIN_CONFIGS_DIR=blockchain-configs/testnet-prod
-    export TRACKER_UPDATE_JSON_RPC_URL=http://35.194.172.106:8080/json-rpc
-    if [[ $NODE_INDEX -lt 5 ]]; then
-        export PEER_CANDIDATE_JSON_RPC_URL="https://summer-api.ainetwork.ai/json-rpc"
+    export TRACKER_UPDATE_JSON_RPC_URL="http://35.194.172.106:8080/json-rpc"
+    export PEER_CANDIDATE_JSON_RPC_URL="http://35.194.169.78:8080/json-rpc"
```
ehgmsdk20, line 129: In mainnet, it seems that PEER_CANDIDATE_JSON_RPC_URL is not used anymore.
Is PEER_CANDIDATE_JSON_RPC_URL necessary for summer?

ehgmsdk20 (approved):
LGTM with some questions

blockchain-configs/testnet-prod/node_params.json
platfowner: We have two clusters (spring and summer) for the testnet-prod config, while we have only one cluster for other configs (e.g. mainnet, testnet-staging, etc). So we were not setting an explicit value for it. But for now we use only spring deprecating summer, so we decided to explicitly set the spring tracker's endpoint here.

start_node_docker.sh
platfowner: Now we can use the default value from blockchain-configs/testnet-prod/node_params.json as we have one now.

start_node_incremental_gcp.sh
platfowner: If env variables are not set explicitly, default values form node_params.json are used instead.

platfowner:

blockchain-configs/testnet-prod/node_params.json
ehgmsdk20: Oh, i got it. Thanks for kindly explaining this!

start_node_docker.sh
ehgmsdk20: I missed the blockchain-configs/testnet-prod/node_params.json, thanks!

Issue #1181, opened 2023-07-07
Need to tweak gc_max_siblings for transfer 


platfowner:
Currently it's 200. We can reduce it to e.g. 30.

![transfer_gc](https://github.com/ainblockchain/ain-blockchain/assets/46211617/07e7beb5-ed4a-4010-be7e-d08c96c3d19e)

platfowner:
Resolved.

PR #1188, opened 2023-07-12
Tweak transfer gc rule

platfowner:
Change sumary:
- Reduce min_gc_num_siblings_deleted to 10
- Tweak transfer gc rule

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1181

### DIFF
--- a/blockchain-configs/mainnet-prod/timer_flags.json
+++ b/blockchain-configs/mainnet-prod/timer_flags.json
@@ -46,5 +46,9 @@
   "update_min_gc_num_siblings_deleted": {
     "enabled_block": 1414800,
     "has_bandage": true
+  },
+  "update_min_gc_num_siblings_deleted2": {
+    "enabled_block": null,
+    "has_bandage": true
   }
 }
--- a/blockchain-configs/mainnet-prod/timer_flags.json
+++ b/blockchain-configs/mainnet-prod/timer_flags.json
@@ -50,5 +50,9 @@
   "update_min_gc_num_siblings_deleted2": {
     "enabled_block": null,
     "has_bandage": true
+  },
+  "tweak_transfer_gc_rule": {
+    "enabled_block": null,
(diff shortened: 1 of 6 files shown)


blockchain-configs/mainnet-prod/timer_flags.json
```diff
@@ -46,5 +46,13 @@
   "update_min_gc_num_siblings_deleted": {
     "enabled_block": 1414800,
     "has_bandage": true
+  },
+  "update_min_gc_num_siblings_deleted2": {
+    "enabled_block": null,
```
woomurf, line 51: If enabled_block is null, how is it applied?

db/bandage-files/update_min_gc_num_siblings_deleted2.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
woomurf, line 1: minor) How about update file name like `decrease_to_10_min_gc_num_siblings_deleted`?

blockchain-configs/mainnet-prod/timer_flags.json
platfowner: > If enabled_block is null, how is it applied?

These null values will be replaced with some numbers when they're actually deployed in the release process.

woomurf: Thank you for explaining!

woomurf (approved):
LGTM, Thanks!

db/bandage-files/update_min_gc_num_siblings_deleted2.js
platfowner: Thanks for the suggestion. For this, let's keep the other parts just simple highlighting the key pram updated.

platfowner:
Thanks for the review. Now merging.. @woomurf

Issue #1182, opened 2023-07-07
Need an API for expected transaction fee (dryrun) 


platfowner:

platfowner:
Resolved.

PR #1189, opened 2023-07-14
Add dryrun transaction API

platfowner:
Change summary:
- Add ain_dryrunSignedTransaction json rpc API
- Update transfer tool for dryrun
- Add test for dryrun transaction API 
- Simplify db restoring logic

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1182


node/index.js
```diff
@@ -792,8 +792,10 @@ class BlockchainNode {
...
           `(${perAccountFreePoolSize}) for account: ${executableTx.address}`);
     }
+    const eventSource = isDryrun ? null : ValueChangedEventSources.USER;
     const result = this.db.executeTransaction(
-        executableTx, false, true, this.bc.lastBlockNumber() + 1, this.bc.lastBlockTimestamp(), ValueChangedEventSources.USER);
+        executableTx, false, true, this.bc.lastBlockNumber() + 1, this.bc.lastBlockTimestamp(),
+        eventSource, isDryrun);
     if (CommonUtil.isFailedTx(result)) {
       if (DevFlags.enableRichTransactionLogging) {
         logger.error(
```
shyun-comcom, line 801: minor) Is it correct to log this even for the case of `!isDryrun`?

p2p/server.js
```diff
@@ -818,6 +818,7 @@ class P2pServer {
 
           continue;
         }
+        // NOTE(platfowner): There is no dryrun mode for batch transactions.
```
shyun-comcom, line 821: Returns error when `isDryrun` is true for batch transaction?

node/index.js
platfowner: enableRichTransactionLogging flag is used for debugging. so it's just fine with the error logging.

p2p/server.js
platfowner: Just changed to allow dryrun for batch txs in executeAndBroadcastTransaction().
Note however that actually it's impossible to send batch txs with isDryrun = true as we don't support dryrunSignedTransactionBatch().

platfowner:
PTAL~ @shyun-comcom

shyun-comcom:
Function `createAndExecuteTransaction` calls `executeAndBroadcastTransaction`.
Is there no need to set the `isDryrun` flag [here](https://github.com/ainblockchain/ain-blockchain/blob/feature/platfowner/feature/client/index.js#L849)?

platfowner:
> Function `createAndExecuteTransaction` calls `executeAndBroadcastTransaction`. Is there no need to set the `isDryrun` flag [here](https://github.com/ainblockchain/ain-blockchain/blob/feature/platfowner/feature/client/index.js#L849)?

createAndExecuteTransaction is for creating transactions for Dev Client SET APIs. Here are some context:
- Dev Client SET APIs are used for testing and development purposes and being disabled in testnet and mainnet deployment
- We have some plan to deprecate Dev Client SET APIs once all the test cases are migrated to json rpc APIs (there so many tests cases and it will take some time) 
- Currently isDryrun = true option is available for json rpc API ain_sendSignedTransactionDryrun only

shyun-comcom (approved):
LGTM!

PR #1196, opened 2023-07-19
Returns gas cost for dryrun in the case of FEE_BALANCE_TOO_LOW

shyun-comcom:


platfowner:
@shyun-comcom Thanks for sending this PR. Let me add some test cases before merging.

platfowner:
> @shyun-comcom Thanks for sending this PR. Let me add some test cases before merging.

Tests added, thanks!

Issue #1277, opened 2024-06-25
Consider extending tx_bytes_limit for bigger tx bodies


platfowner:

platfowner:
Resolved.

PR #1278, opened 2024-06-28
Increase tx_bytes_limit to 100k bytes from 10k bytes

platfowner:
Change summary:
- Increase tx_bytes_limit to 100k bytes from 10k bytes
- Change tx_bytes_limit checking to use sizeof(args.tx_body) instead of sizeof(args)
- Add test cases for ordered nonces (-2)

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1277


jiyoung-an (approved):
LGTM, Thanks!

platfowner:

PR #1279, opened 2024-07-01
Release/v1.3.1

platfowner:
Updates:
- Let verifyTransaction() and verifySignedMessage() handle exceptions 
- Check sizeof(args.tx_body) instead of sizeof(args) for tx_bytes_limit 
- Change tx_bytes_limit to 100000 from 10000 
- Add test cases for ordered nonces (-2)


jiyoung-an:
LGTM, Thanks!

jiyoung-an (approved):

platfowner:

Issue #1282, opened 2024-07-02
Need to change derivation path to "m/44'/60'/0'/0" from "m/44'/412'/0'/0/"


platfowner:
Recently we decided to use the derivation path of ethereum network ("m/44'/60'/0'/0") instead of the ai network's own ("m/44'/412'/0'/0/"). So ain-util and tools/account/seedPhraseToAccounts.js need to be updated accordingly.

cc: @jiyoung-an @shyun-comcom @minjaeli @YunsuBae @yoojinko @akastercomcom

platfowner:
https://bitbucket.org/comcomai/ain-connect-extension/src/main/src/Store/walletStore.js

const pathPrefix = "m/44'/60'/0'/0";

platfowner:
Resolved.

platfowner:
Resolved.

PR #1283, opened 2024-07-05
Upgrade ain-util package version to v1.2.1

platfowner:
Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1282

### DIFF
--- a/tools/transfer/sendTransferTx.js
+++ b/tools/transfer/sendTransferTx.js
@@ -17,9 +17,10 @@ function buildTransferTxBody(fromAddr, toAddr, key, amount, timestamp) {
     },
     gas_price: 500,
     timestamp,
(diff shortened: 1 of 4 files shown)


tools/transfer/sendTransferTx.js
```diff
@@ -18,6 +18,8 @@ function buildTransferTxBody(fromAddr, toAddr, key, amount, timestamp) {
     gas_price: 500,
     timestamp,
     nonce: -1
+//    nonce: -2
+//    nonce: 0
```
jiyoung-an, line 22: Is it intentional code?

jiyoung-an (approved):
LGTM, Thanks!

tools/transfer/sendTransferTx.js
platfowner: yes, just added a note.

platfowner:

Issue #1290, opened 2024-08-07
Need timeout and monitoring for event handler websocket connections


platfowner:
Error log:
```
{"name":"testnet-prod","hostname":"spring-node-0-taiwan","pid":1053759,"level":50,"msg":"[EVENT_CHANNEL_MANAGER] [handleConnection] The number of event channels exceeds its limit (10)","time":"2024-08-07T01:56:25.929Z","v":0}
```

We need to add for event handler websocket connections:
- auto-timeout logic 
- healthiness monitoring 

cc: @jiyoung-an @minjaeli

platfowner:
Issue resolved.

platfowner:
google load balancer's session affinity:
https://cloud.google.com/load-balancing/docs/backend-service#generated_cookie_affinity

PR #1292, opened 2024-08-08
Include more in the event handler status

platfowner:
Change summary:
- Include more in the event handler status
- Reuse existing ip addresses in event handler

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1290

### DIFF
--- a/event-handler/event-channel-manager.js
+++ b/event-handler/event-channel-manager.js
@@ -23,12 +23,11 @@ class EventChannelManager {
     this.heartbeatInterval = null;
   }
 
-  async getNetworkInfo() {
-    const ipAddr = await getIpAddress(NodeConfigs.HOSTING_ENV === HostingEnvs.COMCOM || NodeConfigs.HOSTING_ENV === HostingEnvs.LOCAL);
+  getNetworkInfo() {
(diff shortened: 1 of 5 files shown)


event-handler/event-channel-manager.js
```diff
@@ -23,12 +23,11 @@ class EventChannelManager {
...
   }
 
-  async getNetworkInfo() {
-    const ipAddr = await getIpAddress(NodeConfigs.HOSTING_ENV === HostingEnvs.COMCOM || NodeConfigs.HOSTING_ENV === HostingEnvs.LOCAL);
+  getNetworkInfo() {
+    const ipAddr = (NodeConfigs.HOSTING_ENV === HostingEnvs.COMCOM || NodeConfigs.HOSTING_ENV === HostingEnvs.LOCAL) ? this.node.ipAddrInternal : this.node.ipAddrExternal;
     const eventHandlerUrl = new URL(`ws://${ipAddr}:${NodeConfigs.EVENT_HANDLER_PORT}`);
     return {
       url: eventHandlerUrl.toString(),
-      port: NodeConfigs.EVENT_HANDLER_PORT,
```
jiyoung-an, line 31: Q. Could it be omitted because the port is fixed to 443?

jiyoung-an (approved):
LGTM with a question! Thanks!

platfowner:

event-handler/event-channel-manager.js
platfowner: ah, just removed as it's included in the url.

Issue #1285, opened 2024-07-26
Need to support deploy to on-premise machines


platfowner:
We need to support deploy to on-premise machines.

platfowner:
GCP ref docs:
- https://cloud.google.com/load-balancing/docs/negs
- https://cloud.google.com/load-balancing/docs/negs/internet-neg-concepts

PR #1311, opened 2024-08-23
Split code and data directories with season prefixes

platfowner:
Change summary:
- Split code and data directories with season prefixes

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1285


platfowner:

PR #1314, opened 2024-08-29
Use $SEASON as directory instead of prefix

platfowner:
Change summary:
- Use $SEASON as directory instead of prefix

Related issues:
- https://github.com/ainblockchain/ain-blockchain/issues/1285


platfowner:

